PR #4, opened 2022-04-06
add release automation when branch is tagged

devigned:
- build a release when tag is pushed matching 'v[0-9]+.[0-9]+.*'

Note: only for Linux amd64.


cpuguy83 (approved):
LGTM

PR #9, opened 2022-04-15
added get_rootfs to oci

Mossaka:


crates/containerd-shim-wasmtime-v1/src/sandbox/oci.rs
```diff
@@ -28,6 +30,17 @@ pub fn load(path: &str) -> Result<Spec, Error> {
     Ok(spec)
 }
 
+pub fn get_root(spec: &Spec) -> Result<&PathBuf, Error> {
+    let root = spec.root().as_ref().unwrap();
```
cpuguy83, line 34: Doesn't seem like we are using the error return in this function.

crates/containerd-shim-wasmtime-v1/src/sandbox/oci.rs
```diff
@@ -4,10 +4,12 @@ use oci_spec::runtime::Spec;
 use oci_spec::OciSpecError;
 use serde_json as json;
 use std::fs::{File, OpenOptions};
-use std::path::Path;
+use std::path::{Path};
+use cap_std::path::PathBuf;
 use thiserror::Error;
 use wasmtime_wasi::sync::file::File as WasiFile;
 use wasmtime_wasi::{Dir as WasiDir, WasiCtxBuilder};
+use log::info;
```
cpuguy83, line 12: Looks unused

crates/containerd-shim-wasmtime-v1/src/sandbox/oci.rs
Mossaka: fixed

cpuguy83 (approved):
LGTM

Issue #11, opened 2022-04-26
Add troubleshooting guide 


Mossaka:
Currently there is no troubleshotting guide. People might find it hard to follow readme to produce a hello world example. 

Known issues are, but not limited to:
1. containerd currently only support Linux. So in order to build `runwasi`, either you need to have a linux machien or run it in WSL on Windows
2. `docker buildx` is a dependency
3. `make load` is broken

cpuguy83:
How is `make load` broken?

cpuguy83:
containerd does run on Windows, however runwasi only supports Linux at the moment because it has to wire up networking and rootfs mounts.

danbugs:
> How is `make load` broken?

I think @Mossaka is referring to the fact that, when trying to run it on my machine, it didn't replace the export variable `CONTAINERD_NAMESPACE` with `default` even though this line exists:
```
CONTAINERD_NAMESPACE ?= default
```

Considering `?=` only sets the value if it isn't already set, I'm guessing `export CONTAINERD_NAMESPACE` sets it to empty making `?=` not work.

cpuguy83:
Hmmm... perhaps? I know I use this all the time.
But yeah, `export` should be moved... or even removed.

danbugs:
If we want to keep the `export`, we could just do:
```
export CONTAINERD_NAMESPACE ?= default
```

I've tested it and it works, @cpuguy83.

PR #12, opened 2022-04-26
Adding Windows Getting Started doc

danbugs:
This addresses issue #11

### DIFF
--- a/docs/windows-getting-started.md
+++ b/docs/windows-getting-started.md
@@ -21,8 +21,5 @@ After following these steps and navigating to the runwasi directory in your term
 
 After this, you can execute an example, like: `ctr run --rm --runtime=io.containerd.wasmtime.v1 docker.io/library/wasmtest:latest testwasm`.
 
-> To kill the process from the example:
->> - open a secondary terminal and run `ps aux | grep ctr`,
->> - find the process you want to kill, and
->> - run `kill -9 <the pid from the process you want to kill>`
+> To kill the process from the example, you can run: `ctr kill -s SIGKILL testwasm`.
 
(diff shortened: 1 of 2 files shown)


docs/windows-getting-started.md
```diff
@@ -0,0 +1,28 @@
+# Windows: Getting Started
+
+Currently, **runwasi** depends on POSIX environment (i.e., because it has to wire up networking and rootfs mounts). Therefore, to run it on Windows, we recommend utilizing the Windows Subsystem for Linux (WSL).
```
cpuguy83, line 3: Not just POSIX, this is using Linux specific features.

danbugs: I just changed this!

docs/windows-getting-started.md
```diff
@@ -0,0 +1,28 @@
...
+- run `make build`,
+- run `make install`,
+- run `make test/out/img.tar`,
+- open a secondary terminal and run `containerd`, and
+- run `make load`.
+
+After this, you can execute an example, like: `ctr run --rm --runtime=io.containerd.wasmtime.v1 docker.io/library/wasmtest:latest testwasm`.
+
+> To kill the process from the example:
+>> - open a secondary terminal and run `ps aux | grep ctr`,
```
cpuguy83, line 25: This should be `ctr kill -s SIGKILL testwasm`

Mossaka:
LGTM 

nit: I would recommend setting `make build` to be `cargo build --release` by default since the released binary is much much faster than debug mode.

Mossaka (approved):

cpuguy83:
@Mossaka To do this `make build TARGET=release` should do the trick.
Likewise `sudo make install TARGET=release`

docs/windows-getting-started.md
danbugs: Ooo~ I didn't know this was the right way to do this! Thanks (:

cpuguy83 (approved):
LGTM

PR #15, opened 2022-05-06
feat: generic engine in the shim library

Mossaka:
This PR implements a generic engine to the shim library. This removes the dependency on `wasmtime::Engine` in the library entirely, and it is up to the implementation to decide what instance of the engine should be.


crates/containerd-shim-wasmtime-v1/src/sandbox/manager.rs
```diff
@@ -24,33 +24,35 @@ use std::sync::Arc;
...
             phantom: std::marker::PhantomData,
         }
     }
 }
 
-impl<T> Manager for Service<T>
+impl<E: Send + Sync + Clone, T> Manager for Service<E, T>
 where
-    T: Sandbox<Instance = WasiInstance> + 'static,
+    T: Sandbox<E, Instance = WasiInstance> + 'static,
```
cpuguy83, line 55: Can `E`'s type constraints be moved into the `where` clause here?

cpuguy83 (changes requested):
Minor cosmetic change, everything else looks good.

crates/containerd-shim-wasmtime-v1/src/sandbox/manager.rs
Mossaka: Resolved

cpuguy83 (approved):
LGTM

Issue #16, opened 2022-05-06
Move wasi impl to separate crate


cpuguy83:
The repo has a few binaries and a wasi implementation that is fairly tied to wasmtime.
#15 makes the core library runtime agnostic, meaning it does not depend on wasmtime.

In order to completely remove wasmtime as a dependency from the core library it may be useful to move the binaries along with the `Wasi` instance implementation into a separate crate (of course both crates can be in this repo).

Mossaka:
I will take this one

PR #17, opened 2022-05-11
Refactor: move wasmtime shim binary to it's own bin

Mossaka:
This resolves issue #16


Cargo.toml
```diff
@@ -1,7 +1,39 @@
...
+[dev-dependencies]
+tempfile = "3.0"
+
 [workspace]
 members = [
-    "crates/containerd-shim-wasmtime-v1",
+    "crates/containerd-shim-wasm",
 ]
 
+[[bin]]
```
danbugs, line 25: @Mossaka ~ Why not use workspaces for this, so we can separate dependencies?

Mossaka: Thanks for your idea! I would like this shim implementation to be in the same workspace because they share the same `WasiInstance` config, and any change to the dependency should be reflect to all three binaries. When we migrate the Spin Shim and the ASP.NET shim to this repo, we will make each individual shim a workspace, with separate dependencies. Does this make sense?

cpuguy83 (approved):

PR #21, opened 2022-10-19
Support for cgroups

cpuguy83:
This is a series of patches to add support for running wasm code in a cgroup.
While some cgroup controllers can support running different threads in different cgroups, the main one we are interested in, the cgroup memory controller, this doesn't make sense since threads all share memory.
Because of this we'll fork off a new process (which should have CoW version of the main shim's memory) to run the wasm code.

Of course all the issues of fork apply here since this is fairly unsafe from a multithreaded program (which the shim is always multi-threaded), so care must be taken to not try do things like take a lock in the new process because this can cause a deadlock.
There may be some interesting things to test out with the wasmtime engine just to make sure we aren't going to deadlock if there's multiple things happening when the fork occurs.


cpuguy83:
Recommend reviewing this with whitespace changes disabled.

crates/containerd-shim-wasm/src/sandbox/exec.rs
```diff
@@ -0,0 +1,252 @@
...
+
+// This is is a wrapper around clone3 which allows us to pass a pidfd
+// This works otherwise just like normal forking semantics:
+// If this is the parent, the result will be Ok(Context::Parent(pid, pidfd)), where the pid is the pid of the new process.
+// If this is the child, the result will be Ok(Context::Child).
+//
+// Code that runs in the child must not do things like access locks or other shared state.
+//
+// Optionally you can pass in a reference to a file descriptor which will be populated with the pidfd (see pidfd_open(2)).
+pub unsafe fn fork(cgroup: Option<&Box<dyn Cgroup>>) -> Result<Context, Error> {
```
cpuguy83, line 126: Maybe this does not need to be marked as `unsafe`?

crates/containerd-shim-wasm/src/sandbox/exec.rs
```diff
@@ -0,0 +1,252 @@
...
+            if cgroup.version() == CgroupVersion::V2 {
+                cgfd = cgroup.open()?;
+                builder.flag_into_cgroup(&cgfd).flag_newcgroup();
+            }
+        }
+    } else {
+        debug!("no CAP_SYS_ADMIN, not creating new namespaces");
+    }
+
+    let res = { builder.call() };
```
cpuguy83, line 148: This is the unsafe part, perhaps since we are not really giving a way to mis-use it or do bad things with memory we don't need to be unsafe except for just this call?

Cargo.toml
```diff
@@ -17,9 +17,12 @@ cap-std = "0.24.1"
 oci-spec = "0.5"
 thiserror = "1.0"
 serde_json = "1.0"
+nix = "0.25"
 
 [dev-dependencies]
 tempfile = "3.0"
+libc = "0.2.135"
```
danbugs, line 24: It looks like most of our dependencies are using `major.minor`. How about we use `0.2` for `libc`, and also get rid of the patch versioning for `cap-std`, and `chorno`?

crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -17,16 +17,22 @@ serde = "1.0"
...
+caps = "0.1"
+proc-mounts = "0.3.0"
 
 [build-dependencies]
 ttrpc-codegen = { version = "0.3.0", optional = true }
 
 [dev-dependencies]
 tempfile = "3.0"
+pretty_assertions = "1"
+signal-hook = "0.3.14"
```
danbugs, line 35:
```suggestion
cap-std = "0.24"
chrono = "0.4"
log = "0.4"
clone3 = "0.2"
libc = "0.2"
caps = "0.1"
proc-mounts = "0.3"

[build-dependencies]
ttrpc-codegen = { version = "0.3", optional = true }

[dev-dependencies]
tempfile = "3"
pretty_assertions = "1"
signal-hook = "0.3"
```

~same as above...

.github/workflows/ci.yml
```diff
@@ -23,6 +23,8 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - name: Build
-        run: cargo build --verbose
+        run: cargo build --all --verbose
       - name: Run tests
-        run: cargo test --verbose
+        run: cargo test --all --verbose
+      - name: Run tests as root
+        run: cargo test --all --no-run 2>&1 | grep "Executable" | tee /dev/stderr | awk '{ print $5 }' | tr -d '()' | xargs -I {} sudo sh -c "{}"
```
danbugs, line 30:
```suggestion
        run: cargo test --all --no-run 2>&1 | grep "Executable" | tee /dev/stderr | awk '{ print $4 }' | tr -d '()' | xargs -I {} sudo sh -c "{}"
```

I've tested this locally, `$5` was empty for me and `$4` had the executables...

danbugs (commented):
~haven't finished reviewing the `src/instance.rs`, `crates/containerd-shim-wasm/src/sandbox/exec.rs`, and `crates/containerd-shim-wasm/src/sandbox/oci.rs` files just, but it's getting late, so figured I'd just post what I have so far. That is:
- two suggestions (regarding dependencies versioning), and
- a possible mistake in a CI action!

I'll probably review the rest during the weekend!

.github/workflows/ci.yml
cpuguy83: Yep, same.
I'm not sure why CI is like this, but can definitely confirm it's running the way I want it to in gh actions.

crates/containerd-shim-wasm/src/sandbox/exec.rs
danbugs: You're right~ It doesn't. We can simply wrap the `builder.call()` with `unsafe`.

danbugs, line 126:
```suggestion
pub fn fork(cgroup: Option<&dyn Cgroup>) -> Result<Context, Error> {
```

If we are passing a reference, we don't need the `Box` around this -- will suggest modifications to the two function calls of this too

crates/containerd-shim-wasm/src/sandbox/exec.rs
danbugs: Yep~ We can just wrap this individual call with `unsafe`.

danbugs, line 148:
```suggestion
    let res = unsafe { builder.call() };
```

crates/containerd-shim-wasm/src/sandbox/exec.rs
```diff
@@ -0,0 +1,252 @@
...
+
+    #[test]
+    fn test_fork() -> Result<(), Error> {
+        let test_exit_code = 42;
+        let cg = cgroups::new("test_fork".to_string())?;
+
+        // Use pipes to signal from the child to the parent
+        let (r, w) = pipe2(OFlag::O_CLOEXEC).unwrap();
+
+        let ret = unsafe { fork(Some(&cg)) };
```
danbugs, line 200:
```suggestion
        let ret = fork(Some(&*cg));
```

src/instance.rs
```diff
@@ -137,137 +136,95 @@ pub fn prepare_module(
...
-                            error!("exit code: {}", 137);
-                            let mut ec = lock.lock().unwrap();
-                            *ec = Some((137, Utc::now()));
-                        }
-                    };
+        oci::setup_cgroup(&cg, &spec)?;
 
-                    cvar.notify_all();
-                })?;
+        let res = unsafe { exec::fork(Some(&cg)) }?;
```
danbugs, line 188:
```suggestion
        let res = exec::fork(Some(&*cg))?;
```

crates/containerd-shim-wasm/src/sandbox/oci.rs
```diff
@@ -44,8 +29,51 @@ pub fn get_args(spec: &Spec) -> &[String] {
...
+
+    let p = linux.as_ref().unwrap().cgroups_path();
+    if p.is_none() {
+        return Ok(Box::new(NopCgroup {}));
+    }
+
+    cgroups::new(p.as_ref().unwrap().to_str().unwrap().to_string())
+}
+
+pub fn setup_cgroup(cg: &Box<dyn cgroups::Cgroup>, spec: &Spec) -> Result<()> {
```
danbugs, line 72:
```suggestion
pub fn setup_cgroup(cg: &dyn cgroups::Cgroup, spec: &Spec) -> Result<()> {
```

~don't need `Box` if we are passing a reference, will suggest fixes on the function calls of this too.

src/instance.rs
```diff
@@ -137,137 +136,95 @@ pub fn prepare_module(
...
-                            let mut ec = lock.lock().unwrap();
-                            *ec = Some((0, Utc::now()));
-                        }
-                        Err(_) => {
-                            error!("exit code: {}", 137);
-                            let mut ec = lock.lock().unwrap();
-                            *ec = Some((137, Utc::now()));
-                        }
-                    };
+        oci::setup_cgroup(&cg, &spec)?;
```
danbugs, line 186:
```suggestion
        oci::setup_cgroup(&*cg, &spec)?;
```

danbugs (changes requested):
some improvements here and there~

crates/containerd-shim-wasm/src/sandbox/error.rs
```diff
@@ -52,9 +55,6 @@ impl From<Error> for ttrpc::Error {
                 ttrpc::Error::RpcStatus(ttrpc::get_status(ttrpc::Code::FAILED_PRECONDITION, s))
             }
             Error::Oci(ref s) => match s {
```
danbugs, line 57:
```suggestion
            Error::Oci(ref s) => {
```

danbugs:
I'll come back to do a review of the functionality tomorrow~ lemme know if there's anything specific you'd like me to do locally to test the cgroups!

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+                let path = self.get_file(&format!("hugetlb.{}.max", limit.page_size()));
+                ensure_write_file(path, &format!("{}", limit.limit()))?;
+            }
+        }
+
+        if let Some(blkio) = res.block_io() {
+            if let Some(weight) = blkio.weight() {
+                ensure_write_file(
+                    self.get_file("io.weight"),
+                    &format!("{}", (weight - 10) * 9999 / 990),
```
Mossaka, line 446: https://github.com/containerd/cgroups/blob/724eb82fe759f3b3b9c5f07d22d2fab93467dc56/v2/utils.go#L209

Was it a mistake? Missing 1 plus here

```suggestion
                    &format!("{}", 1 + (weight - 10) * 9999 / 990),
```

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+                        Error::Others(format!(
+                            "could not convert cgroup.controllers to string: {}",
+                            e
+                        ))
+                    })?
+                    .trim();
+
+                if trimmed.len() > 0 {
+                    return Err(Error::FailedPrecondition(
+                        "hybyrid cgroup is not supported".to_string(),
```
Mossaka, line 545: just for my own understanding: why is hybrid cgroup not supported?

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+                    ));
+                }
+
+                return Ok(Box::new(CgroupV1 {
```
Mossaka, line 549: minor: I think you can remove this extra return

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+                    base: v1,
+                    path: PathBuf::from(name),
+                }));
+            }
+            return Ok(Box::new(CgroupV1::new(v1, PathBuf::from(name))));
+        }
+    }
+
+    if v2.exists() {
+        return Ok(Box::new(CgroupV2::new(v2, PathBuf::from(name))));
```
Mossaka, line 559: what if v1 is mounted directly on `/sys/fs/cgroup/unified`?

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+            for (k, v) in unified {
+                ensure_write_file(self.get_file(k), v)?;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+pub fn new(name: String) -> Result<Box<dyn Cgroup>> {
```
Mossaka, line 521: If would be great to see more unit tests for `new` function, considering its logical to determine v1 and v2 cgroups is quiet obscuring. 

For fast path, we probably need 
- one test case for detecting hybrid mode, 
- one test case with v2 mounted on v1 path, 
- one test case with v1 mounted on v2 path,
- one test case where v1 v2 both not found

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+    if v1_found && v2_found {
+        let p = PathBuf::from(base.clone()).join("cgroup.controllers");
+        if std::fs::read(p).map_err(|e| std::io::Error::from(e))?.len() > 0 {
+            return Ok(Box::new(CgroupV1 {
+                base: base,
+                path: PathBuf::from(name),
+            }));
+        }
+        return Err(Error::FailedPrecondition(
+            "hybyrid cgroup is not supported".to_string(),
```
Mossaka, line 605: Why is the logic here to detect hybyrid cgroup different from the fast path? Specifically, in the fast path, we converted `std::fs::read(v2.join("cgroup.controllers"))` to `trimmed` string and then checked length of the string. Here, we only checked `std::fs::read(p).len() > 0`.

Mossaka (changes requested):
Great work, @cpuguy83 ! Left some comments on `cgroups.rs` file~

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
Mossaka: Not asking these unit tests in this PR though. I'd like to create backlog and have another PR to increase the test coverage.

Mossaka:
For some reason, when I retried to run a testwasm container, I got the following error
```
➜  runwasi git:(pr-21) ✗ sudo ctr run --rm --runtime=io.containerd.wasmtime.v1 docker.io/library/wasmtest:v1 testwasm wasm echo hello world 
ctr: Others("could not open cgroup file /default/testwasm/cpu.shares: No such file or directory (os error 2)"): unknown
```

src/instance.rs
```diff
@@ -137,137 +136,95 @@ pub fn prepare_module(
...
+                    *ec = Some((status.status, Utc::now()));
+                    drop(lock);
+                    cvar.notify_all();
+                });
+                Ok(tid)
+            }
+            exec::Context::Child => {
+                // child process
+
+                // TODO: How to get exit code?
```
Mossaka, line 220: I believe it's totally fine to assume that Okay case returns exit code 0. The trick case is how to handle error. 

See [here](https://docs.wasmtime.dev/api/wasmtime/struct.Func.html#errors-4), wasmtime `call()` will return an `Error` type which can be `Trap` or type error in params. 

[Trap](https://docs.wasmtime.dev/api/wasmtime/enum.Trap.html) is defined as 

```rust
pub enum Trap {
    StackOverflow,
    MemoryOutOfBounds,
    HeapMisaligned,
    TableOutOfBounds,
    IndirectCallToNull,
    BadSignature,
    IntegerOverflow,
    IntegerDivisionByZero,
    BadConversionToInteger,
    UnreachableCodeReached,
    Interrupt,
    AlwaysTrapAdapter,
    OutOfFuel,
}
```

Maybe we can try mapping each enum value of `Trap` to proper exit code?

src/instance.rs
```diff
@@ -277,16 +234,23 @@ impl Instance for Wasi {
...
             "module is not running".to_string(),
         ))?;
-
-        i.interrupt();
-        Ok(())
+        fd.kill(signal as i32)
     }
 
     fn delete(&self) -> Result<(), Error> {
+        let spec = match load_spec(self.bundle.clone()) {
```
Mossaka, line 245: `load_spec` result can be saved in Instance internal state.

Mossaka (commented):
`exec.rs` lgtm!

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+    fn delete(&self) -> Result<()> {
+        let path = self.base.join(&self.path);
+        if path.exists() {
+            std::fs::remove_dir(path)?;
+        }
+        Ok(())
+    }
+
+    fn add_task(&self, pid: u32) -> Result<()> {
+        ensure_write_file(self.get_file("cgroup.procs"), &format!("{}", pid))?;
```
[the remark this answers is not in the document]
cpuguy83: Correct.
cgroupv1 is "multi hierarchy" and cgroupv2 is "unified hierarchy".

Both cgroupv1 and cgroupv2 have controllers (memory controller, cpu controller, etc).
A process can be in different heirarchies in cgroupv1 (e.g. cpu/foo and memory/bar), but in v2 there is only 1 hierarchy (foo/memory && foo/cpu, or foo/bar/memory and foo/bar/cpu).

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+                | nix::fcntl::OFlag::O_RDONLY,
+            nix::sys::stat::Mode::empty(),
+        )
+        .map_err(|e| std::io::Error::from(e).into())
+    }
+
+    fn delete(&self) -> Result<()> {
+        let path = self.base.join(&self.path);
+        if path.exists() {
+            std::fs::remove_dir(path)?;
```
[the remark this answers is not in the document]
cpuguy83: The correct way to delete a cgroup is to use rmdir (you can test this out in a terminal).
For instance if you try to `rm -rf` a cgroup dir (which will try to recursively delete every object in the dir) you will get an error because cgroups are a special filesystem and it doesn't make sense to delete files inside a cgroup. If the cgroup exists the files must also exist, so `rmdir` is how to delete the group.

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+            }
+        }
+
+        if let Some(blkio) = res.block_io() {
+            let controller_path = self.get_controller("blkio")?;
+            if let Some(weight) = blkio.weight() {
+                ensure_write_file(controller_path.join("blkio.weight"), &weight.to_string())?;
+            }
+            if let Some(weight_device) = blkio.weight_device() {
+                let mut file = OpenOptions::new()
```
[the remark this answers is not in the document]
cpuguy83: Indeed I considered this, but went with this since I'm writing to the same file multiple times.
But I can definitely switch it to `ensure_write_file` to simplify things. Trade-off here is is overhead opening/closing repeatedly vs 1 open/close for the entire thing.

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+            if let Some(quota) = cpu.quota() {
+                max = format!("{}", quota);
+            }
+            if let Some(period) = cpu.period() {
+                ensure_write_file(self.get_file("cpu.max"), &format!("{} {}", max, period))?;
+            } else {
+                ensure_write_file(self.get_file("cpu.max"), max.as_str())?;
+            }
+
+            // no realtime support
```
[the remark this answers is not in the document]
cpuguy83: It's just a note since the oci spec has "real-time" cpu constraints but cgroupv2 does not support that (though we do include in in the cgroupv1 code).
I can delete it or possibly expand on this.

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
cpuguy83: Simplicity. It could be added but for now it's either cgroupv1 or cgroupv2.

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
cpuguy83: Indeed this function is tricky... and even trickier to test in CI because we can't really easily control different cgroup modes.

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
cpuguy83: Nooooo, off by one!

crates/containerd-shim-wasm/src/sandbox/cgroups.rs
```diff
@@ -0,0 +1,674 @@
...
+        Ok(())
+    }
+}
+
+pub fn new(name: String) -> Result<Box<dyn Cgroup>> {
+    // fast path
+    let v1 = PathBuf::from("/sys/fs/cgroup");
+    let v2 = PathBuf::from("/sys/fs/cgroup/unified");
+    if v1.join("cpu").exists() {
+        let stat = statfs(v1.join("cpu").as_path()).map_err(|e| std::io::Error::from(e))?;
```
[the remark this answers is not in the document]
cpuguy83: 👀 That's amazing.

crates/containerd-shim-wasm/src/sandbox/exec.rs
cpuguy83: Oh great, I thought this wasn't quite right.
I see now I can pass the boxed reference with `.as_ref()`.

crates/containerd-shim-wasm/src/sandbox/exec.rs
```diff
@@ -0,0 +1,252 @@
...
+        let ws = wait(self.fd, false)?;
+        Ok(ws.into())
+    }
+
+    pub fn wait_no_hang(&self) -> Result<Status, Error> {
+        wait(self.fd, true)
+    }
+
+    // Send the specified signal to the process referred to by the pidfd.
+    pub fn kill(&self, sig: i32) -> Result<(), Error> {
```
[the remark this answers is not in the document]
cpuguy83: I think the usage here is safe since there is no way to mess up memory safety here except for the code inside of the `unsafe` block. There's nothing the caller can do to mess it up.

crates/containerd-shim-wasm/src/sandbox/exec.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// This is is a wrapper around clone3 which allows us to pass a pidfd
+// This works otherwise just like normal forking semantics:
+// If this is the parent, the result will be Ok(Context::Parent(pid, pidfd)), where the pid is the pid of the new process.
+// If this is the child, the result will be Ok(Context::Child).
+//
+// Code that runs in the child must not do things like access locks or other shared state.
+// The child should not depend on other threads in the parent process since the new process becomes single threaded.
+pub unsafe fn fork(cgroup: Option<&dyn Cgroup>) -> Result<Context, Error> {
```
cpuguy83, line 125: fyi, I decided to keep this as `unsafe` because the caller does need to be very careful about the context they are calling this from.

danbugs: sounds good~ it would be helpful to add a `# Safety` section in documentation comments for this function then!

danbugs:
> For some reason, when I retried to run a testwasm container, I got the following error
> 
> ```
> ➜  runwasi git:(pr-21) ✗ sudo ctr run --rm --runtime=io.containerd.wasmtime.v1 docker.io/library/wasmtest:v1 testwasm wasm echo hello world 
> ctr: Others("could not open cgroup file /default/testwasm/cpu.shares: No such file or directory (os error 2)"): unknown
> ```

I got the same error:
image

Mossaka:
> > For some reason, when I retried to run a testwasm container, I got the following error
> > ```
> > ➜  runwasi git:(pr-21) ✗ sudo ctr run --rm --runtime=io.containerd.wasmtime.v1 docker.io/library/wasmtest:v1 testwasm wasm echo hello world 
> > ctr: Others("could not open cgroup file /default/testwasm/cpu.shares: No such file or directory (os error 2)"): unknown
> > ```
> 
> I got the same error: image
> 
> EDIT: I tried manually creating the file and it does spin up, but I'm not getting the song that never ends like we do in main.

Can you elaborate what file you were manually creating?

src/instance.rs
cpuguy83: From what I recall when working with wasmtime in go, was seems to support exit codes.
We could try to capture these traps and translate though.

This isn't new code though, just moved due to whitespace changes.

src/instance.rs
cpuguy83: Maybe we can deal with this separately?
We are only calling this twice, once on start and once on delete, but storing this adds more complexity to `fn new` (which is tricky because it can't return an error) or more complexity with lazily-loading the spec.

cpuguy83:
Updated:

- Fixed issue with absolute group paths (e.g. `/foo/bar`) causing the base path of the group to be overwritten (along with tests)
- Split cgroup impls into separate modules
- Re-wrote cgroup detection logic to be more robust
- Support for v1 cgroup controllers which do not share a common root
- Support for more cgroupv1 controllers
- Added more tests around group setup

This should have a follow-up to add CI for cgroupv2 to make sure those code paths are exercised.
This will require setting up nested virt so we have control over the kernel.

Mossaka (approved):
lgtm!

Mossaka:
Hey @cpuguy83, in our last meeting, I asked you feature parity between runwasi and runc. I remember you said there are features like namespaces and OCI hooks which are not implemented in runwasi. Do we plan to add them in the future? If yes, we could create issues for them.

danbugs (approved):
LGTM!

cpuguy83:
2 LGTM's, I'm gonna go ahead and merge.

PR #31, opened 2022-12-01
add standard containerd org documentation and licensing

devigned:
- Update the license file to reflect the move to the containerd org
- Add security notifications file pointing to containerd org resources
- Add maintainers list
- Update the readme with containerd governance info 

Signed-off-by: David Justice <david@justice.dev>


.gitignore
```diff
@@ -1,4 +1,8 @@
 bin/
 test/out/img.tar
 **/target/
-!src/bin/
\ No newline at end of file
+!src/bin/
+
+# IDE related ignores
```
samuelkarp, line 6: In the core projects we've generally kept IDE-related ignores out and encouraged people to use a global gitignore file on their local machine, but I don't think we've made that a hard and fast rule. The downside to including them here is that the list can keep growing essentially unbounded as more contributors use different editors.

samuelkarp (approved):

cpuguy83 (approved):

PR #33, opened 2022-12-02
cgroups: v2: Enable all controllers from parent

### LINKED ISSUE #28: Cargo test fail with test_cgroup
**Not fail on every machines.**

I have success and fail machines both, but I am not sure what kind of information I could provide.
I could help test in my environment if you have any idea.
> So far I know if I  `sudo mkdir` under `/sys/fs/cgroup/memory`, they will generate different folder content between success and fail machine.

Run
```shell
cargo test --all test_cgroup --verbose
```

And I got those error log.
```
---- sandbox::cgroups::tests::test_cgroup stdout ----
Error: Others("failed to apply cgroup: could not open cgroup file /sys/fs/cgroup/relative/nested/containerd-wasm-shim-test_cgroup/memory.max: No such file or directory (os error 2)")

---- sandbox::cgroups::tests::test_cgroup stdout ----
running test with sudo: sandbox::cgroups::tests::test_cgroup
Error: Stdio(Kind(Other))
```

cpuguy83:
The test for nested cgroups was failing for v2 cgroups because controllers have to be explicitly enabled for child cgroups.

This change makes sure we enable those controllers for all parents of a nested cgroup *that we are creating*.
This does not change parents that we did not create.

Fixes #28

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/cgroups/mod.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/cgroups/mod.rs
@@ -178,6 +178,7 @@ fn ensure_write_file(path: std::path::PathBuf, content: &str) -> Result<()> {
     }
     let mut file = fs::OpenOptions::new()
         .write(true)
+        .create(true)
         .open(&path)
         .map_err(|e| {
             Error::Others(format!(
@@ -249,6 +250,7 @@ fn find_cgroup_mounts<T: BufRead>(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use log::debug;
     use oci_spec::runtime::Spec;
     use serde_json as json;
     use std::io::{Cursor, Write};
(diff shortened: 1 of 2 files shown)


crates/containerd-shim-wasm/src/sandbox/cgroups/mod.rs
```diff
@@ -286,13 +287,35 @@ mod tests {
...
     #[test]
     fn test_cgroup() -> Result<()> {
         if !super::super::exec::has_cap_sys_admin() {
             println!("running test with sudo: {}", function!());
             return run_test_with_sudo(function!());
         }
+
         let cg = new("containerd-wasm-shim-test_cgroup".to_string())?;
+
+        println!("cgroup: {}", cg.version());
```
devigned, line 317: Is this here for debugging?

devigned (changes requested):
Looks solid. Just one question.

crates/containerd-shim-wasm/src/sandbox/cgroups/mod.rs
cpuguy83: Yes, it's useful if this fails later to know what version of groups the test ran against.

devigned: What do you think about `log::debug!()` rather than `println!()`?

cpuguy83: Changed.

devigned (approved):
lgtm!

danbugs (approved):
LGTM

PR #36, opened 2022-12-07
Add custom buildx builder

jsturtevant:
Signed-off-by: James Sturtevant <jstur@microsoft.com>

When running `make load` I ended up not being able to build due to using the docker target in buildx:

```
error: Docker exporter feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
make: *** [Makefile:25: test/out/img.tar] Error 1
```

This fixes it by using a custom builder and adds a `clean` target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,13 @@ install:
 # TODO: build this manually instead of requiring buildx
 test/out/img.tar: test/image/Dockerfile test/image/src/main.rs test/image/Cargo.toml test/image/Cargo.lock
 	mkdir -p $(@D)
+	docker buildx rm wasmbuilder || true
+	docker buildx create --name wasmbuilder --use
 	docker buildx build --platform=wasi/wasm -o type=docker,dest=$@ -t $(TEST_IMG_NAME) ./test/image
 
 load: test/out/img.tar
 	sudo ctr -n $(CONTAINERD_NAMESPACE) image import $<
+
+clean:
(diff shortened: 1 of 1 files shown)


danbugs (approved):
LGTM

Makefile
```diff
@@ -22,7 +22,13 @@ install:
 # TODO: build this manually instead of requiring buildx
 test/out/img.tar: test/image/Dockerfile test/image/src/main.rs test/image/Cargo.toml test/image/Cargo.lock
 	mkdir -p $(@D)
+	docker buildx rm wasmbuilder || true
```
cpuguy83, line 25: This is going to mess up my build cache.
Maybe we can instruct people to set a different builder instead?

Note that you should be able to pass one in by setting $BUILDX_BUILDER already.

cpuguy83 (changes requested):

Makefile
jsturtevant: only for this builder? This only gets run once do the file marker `test/out/img.tar`

I didn't know about $BUILDX_BUILDER.  I tried it and it worked. Will be easy enough to just give that hint in the readme.

cpuguy83: I wanted to get away from requiring docker for building this.
I'll work on that.

jsturtevant: sounds good.  Closing in favor of https://github.com/containerd/runwasi/pull/41

PR #37, opened 2022-12-09
cgroup: v2: do not set subtree_control for new cg

### LINKED ISSUE #35: Others("Device or resource busy (os error 16)"): unknown
After [building ](https://github.com/containerd/runwasi#build)and [running the demo example](https://github.com/containerd/runwasi#examples), I got the following error:

```
sudo ctr run --rm --runtime=io.containerd.wasmtime.v1 docker.io/library/wasmtest:latest testwasm
ctr: Others("Device or resource busy (os error 16)"): unknown
```

## investigation
The task is marked as CREATED:

```
sudo ctr task ls
TASK          PID    STATUS
testwasm13    0      CREATED
```

but get the following error when trying to delete it:

```
sudo ctr task rm testwasm13
ERRO[0000] unable to delete testwasm13                   error="task must be stopped before deletion: created: failed precondition"
ctr: task must be stopped before deletion: created: failed precondition
```

Also get a slightly different error when trying to "stop" it:

```
sudo ctr task kill -s SIGKILL testwasm13
ctr: cannot kill non-running container, current state: Exited(TaskState { s: PhantomData }): failed precondition
```

## other info
There seems to be two issues:

- the shim is and containerd are out of sync on the state of the shim. this leads to not being able to clean up the task/container
- There is an unhandled exception in https://github.com/containerd/runwasi/blob/e266bbbb51f2a2e1e207304f87090717278bab45/src/instance.rs#L190 This causes the `Device or resource busy (os error 16)`

## versions
containerd version: `containerd containerd.io 1.6.7 0197261a30bf81f1ee8e6a4dd2dea0ef95d67ccb`
shim version: built from `main` (e266bbbb51f2a2e1e207304f87090717278bab45)
linux version:

```
lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04
Codename:       jammy
```

It does works on my WSL instance:

```
containerd containerd.io 1.6.6 10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1    

 lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.4 LTS
Release:        20.04
Codename:       focal
```

cpuguy83:
A cgroup can either have delegations (`cgroup.subtree_control` is populated) OR it can have processes in it, but not both. So we need to make sure that we don't add any controllers to the new cgroup's `cgroup.subtree_control` BUT we should continue to do so for any parents that we create.

Fixes #35

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/cgroups/cgroupv2.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/cgroups/cgroupv2.rs
@@ -4,6 +4,7 @@ use super::{
     ensure_write_file, find_cgroup_mounts, list_cgroup_controllers, new_mount_iter, safe_join,
     Cgroup, CgroupOptions, Version,
 };
+use log::debug;
 use nix::sys::statfs;
 pub use oci_spec::runtime::LinuxResources as Resources;
 use std::fs;
@@ -24,18 +25,41 @@ impl CgroupV2 {
         safe_join(self.base.clone(), self.path.join(name))
     }
 
+    fn full_path(&self) -> Result<PathBuf> {
+        safe_join(self.base.clone(), self.path.clone())
+    }
+
     // Make sure the cgroup parents exist and that cgroup controllers are delegated from the parent cgroup.
     fn ensure_parents(&self) -> Result<()> {
+        debug!("ensuring cgroup parents exist: {}", self.path.display());
         let mut full = safe_join(PathBuf::from("/"), self.base.clone())?;
 
-        for d in self.path.clone().iter() {
+        let p = self.path.clone();
+        for d in p.iter() {
             full = safe_join(full.clone(), PathBuf::from(d))?;
 
             if full.exists() {
+                debug!("skipping creation of existing cgroup: {}", full.display());
                 // This is a pre-existing cgroup, so we won't touch subtree control.
                 continue;
             }
 
+            if full.to_str() == self.full_path()?.to_str() {
+                // This is the last element in the path, so we don't need to create it.
+                // The cgroup will get created later when we try to write to a file in it.
+                //
+                // This is also neccessary because we do not want to set subtree control on this cgroup.
+                // Otherwise we will not be able to put any processes into this cgroup.
+                // A cgroup can either delegate controllers to its children or it can have processes in it, but not both.
+                // https://docs.kernel.org/admin-guide/cgroup-v2.html#no-internal-process-constraint
+                debug!(
+                    "skipping creation of last element in cgroup path: {}",
+                    full.to_str().unwrap()
+                );
+                break;
+            }
+
+            debug!("creating cgroup directory: {}", full.display());
             fs::create_dir_all(&full)?;
 
             // For every enabled controller, add it to the subtree control so the child cgroup can use it.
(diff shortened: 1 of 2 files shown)


cpuguy83:
Ok this is good to go.

crates/containerd-shim-wasm/src/sandbox/cgroups/cgroupv2.rs
```diff
@@ -26,16 +27,35 @@ impl CgroupV2 {
...
 
             if full.exists() {
+                debug!("skipping creation of existing cgroup: {}", full.display());
                 // This is a pre-existing cgroup, so we won't touch subtree control.
                 continue;
             }
 
+            if full.to_str() == safe_join(self.base.clone(), self.path.clone())?.to_str() {
+                // This is the last element in the path, so we don't need to create it.
+                // The cgrouop will get created later when we try to write to a file in it.
```
jsturtevant, line 45:
```suggestion
                // The cgroup will get created later when we try to write to a file in it.
```

crates/containerd-shim-wasm/src/sandbox/exec.rs
```diff
@@ -145,7 +145,7 @@ pub unsafe fn fork(cgroup: Option<&dyn Cgroup>) -> Result<Context, Error> {
             }
         }
     } else {
-        debug!("no CAP_SYS_ADMIN, not creating new namespaces");
+        debug!("no CAP_SYS_ADMIN, not creating setting cgroup");
```
jsturtevant, line 148: "not creating cgroup"?

crates/containerd-shim-wasm/src/sandbox/cgroups/cgroupv2.rs
```diff
@@ -26,16 +27,35 @@ impl CgroupV2 {
...
+        for d in p.iter() {
             full = safe_join(full.clone(), PathBuf::from(d))?;
 
             if full.exists() {
+                debug!("skipping creation of existing cgroup: {}", full.display());
                 // This is a pre-existing cgroup, so we won't touch subtree control.
                 continue;
             }
 
+            if full.to_str() == safe_join(self.base.clone(), self.path.clone())?.to_str() {
```
jsturtevant, line 43: Is this the only place that the full path (base + path) is calcuated? might consider making base+path a method on the implementation if reused somewhere

jsturtevant (commented):
lgtm

crates/containerd-shim-wasm/src/sandbox/exec.rs
cpuguy83: In this case it would be not setting the cgroup because the cgroup is already created and just passed into this function.

jsturtevant: then it would be?

```suggestion
        debug!("no CAP_SYS_ADMIN, not setting cgroup");
```

cpuguy83: lol, I didn't even notice that `creating` was in there after you pointed it out.

cpuguy83:
Fixed comments

devigned (approved):
lgtm!

jsturtevant (approved):

PR #38, opened 2022-12-09
Fix issue with sudo test runner messing with tty

cpuguy83:
Cargo seems to really enjoy messing with the tty.
Instead of having 2 cargos (main one that executed the test originally, then the sub-test runner) access to the same terminal, make the subprocess go through a pipe for stdout and stderr.


devigned:
Nice! I'm going to pull this down and give it a run to verify behavior.

Mossaka (approved):
lgtm!

PR #41, opened 2022-12-14
Build test image without docker.

cpuguy83:
This manually assembles the tar instead of requring buildkit/buildx/docker to create it.

Instead of using buildx to stick our single binary into a container image we can build the tar ourself.
This should work on Windows (not tested...) or Linux and only depends on cargo to build.

Closes #36


cpuguy83:
@jsturtevant for testing on Windows.

Mossaka (approved):
lgtm!

crates/wasi-demo-app/build.rs
```diff
@@ -0,0 +1,62 @@
...
+    let config = spec::ConfigBuilder::default()
+        .entrypoint(vec!["/wasi-demo-app.wasm".to_owned()])
+        .build()
+        .unwrap();
+
+    let layer_digest = try_digest(layer_path.as_path()).unwrap();
+    let img = spec::ImageConfigurationBuilder::default()
+        .config(config)
+        .os("wasi")
+        .architecture("wasm32")
```
rumpl, line 43: OCI prefers `wasm` instead of `wasm32`, see the [discussion in this PR](https://github.com/opencontainers/image-spec/pull/964#issuecomment-1289408465)

jsturtevant: When I tried this out I had to run `rustup target add wasm32-wasi`. I guess this is related?

jsturtevant:
> @jsturtevant for testing on Windows.

the sample program and image file build run on windows. When I tried loading it into containerd I got the follow error:

```
.\ctr.exe image import --all-platforms ..\..\runwasi\target\wasm32-wasi\debug\img.tar                                                                   unpacking ghcr.io/containerd/runwasi/wasi-demo-app:latest (sha256:01f42703daa223ec938b719aac05b09812f0e6b5ca8c5a69f73ef55546e4b881)...time="2022-12-19T09:11:41-08:00" level=info msg="apply failure, attempting cleanup" error="failed to extract layer sha256:8e73cf52c941fc918688f08efb674b55883a7a1b8ac36b7069af83557c7f3313: hcsshim::ProcessBaseLayer \\\\?\\C:\\ProgramData\\containerd\\root\\io.containerd.snapshotter.v1.windows\\snapshots\\1217: The system cannot find the path specified.: unknown" key="extract-614379200-CtdF sha256:8e73cf52c941fc918688f08efb674b55883a7a1b8ac36b7069af83557c7f3313"                  ctr: failed to extract layer sha256:8e73cf52c941fc918688f08efb674b55883a7a1b8ac36b7069af83557c7f3313: hcsshim::ProcessBaseLayer \\?\C:\ProgramData\containerd\root\io.containerd.snapshotter.v1.windows\snapshots\1217: The system cannot find the path specified.: unknown     
```

I don't think this is an issue with this PR per say.  I am sort of guessing but might be related to the fact that Windows doesn't have a "baselayer" or doesn't know how handle the image layer produced.  I'll have to dig in deeper to figure it out.

I don't think we should hold this up for Windows support as there is quite a bit work to do there anyways.

crates/wasi-demo-app/build.rs
rumpl: I don't think it's related. The architecture here is for the OCI image, not for cargo or the rust code

jsturtevant: i see that now, thanks!

cpuguy83: Fixed

PR #43, opened 2022-12-15
Add a cgroup wrapper which implements Drop

cpuguy83:
This allows tests to fully cleanup.
`Drop` is only implemented for this wrapper because we usually don't
want to delete cgroups just because the value went out of scope.
    
The test wrapper also deletes all parents to make sure the full tree is
deleted.


Mossaka (approved):
lgtm!

devigned (approved):
LGTM!

PR #44, opened 2022-12-20
WIP: Store compiled modules

cpuguy83:
Modules are compiled and stored in a content addressable store. Two runs of the same module are significantly faster since we only need to digest the wasm module being loaded and do not need to compile it again.

This is still a work in progress.
The store is not really concurrency safe yet (it mostly works, but multiple threads/processes could hit issues).

### DIFF
--- a/crates/containerd-shim-wasm/src/content/mod.rs
+++ b/crates/containerd-shim-wasm/src/content/mod.rs
@@ -1,0 +1,161 @@
+use anyhow::{Context, Error, Result};
+use serde::{Deserialize, Serialize};
+use sha256::try_digest;
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::PathBuf;
+
+#[derive(Deserialize, Serialize, PartialEq, Clone)]
+pub struct Digest {
+    alg: String,
+    enc: String,
+}
+
+impl Digest {
+    fn new(algorithm: String, encoded: String) -> Self {
+        Self {
+            alg: algorithm,
+            enc: encoded,
+        }
+    }
+
+    fn algorithm(&self) -> &str {
+        &self.alg
+    }
+
+    fn encoded(&self) -> &str {
+        &self.enc
+    }
+}
+
+impl std::fmt::Display for Digest {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}:{}", self.algorithm(), self.encoded())
+    }
+}
+
+impl TryFrom<&str> for Digest {
+    type Error = Error;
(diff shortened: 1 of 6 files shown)


cpuguy83:
@rumpl

crates/containerd-shim-wasm/src/content/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+            .join("metadata")
+            .join(dgst.algorithm())
+            .join(dgst.encoded());
+
+        let mut file = File::open(path).context("could not open metadata file")?;
+        let mut buf = String::new();
+        file.read_to_string(&mut buf)
+            .context("could not read metadata file")?;
+
+        Ok(serde_json::from_str(&buf).context("could not parse metadata file")?)
```
rumpl, line 104: This could be written like

```rust
let data = fs::read_to_string(path).context("could not read metadata file");
Ok(serde_json::from_str(&data).context("could not parse metadata file")?)
```

`fs::read_to_string` does the exact same thing your code does

crates/containerd-shim-wasm/src/content/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+
+pub struct ContentWriter {
+    f: File,
+    ingest_path: PathBuf,
+    target: PathBuf,
+}
+
+impl ContentWriter {
+    pub fn write(&mut self, data: &[u8]) -> Result<usize> {
+        self.f.try_clone()?.write(data).context("write")
```
rumpl, line 140: Curious why do you `try_clone` here? Can't we directly write `self.f.write(data).context("write")`?

cpuguy83: Lots of issues with borrowing if I recall correctly. Could also just be a remnant of me trying to get a different implementation working with file locking.

squillace:
@cpuguy83 does this make any allowance for _precise_ understanding of the module to check whether it's been cached? That is, it seems like it checks a sha.... which I'd be assuming is a digest. Am I right? Just want to make sure it's easy to move from a quick, possibly unsafe check to a more formal one down the line.... looks GREAT though.

cpuguy83:
@squillace This is based on the digest of the module from the image.
There does seemt o be some extra namespacing we need to do at least for wasmtime since it (IIRC) will reject modules compiled for a different version of wasmtime.

squillace:
I'm wondering whether that's really just a load problem: if you pull an arm image on an amd node, you'll get the same "I just can't today" error. Though if there's a coherent way to determine and handle it in the future.....

jsturtevant:
going to close this in favor of #405. Thanks @cpuguy83 for the inspiration!

PR #57, opened 2023-01-27
Cleanups from static code analysis

ipuustin:
I ran clippy static code analysis, and the tool had a bunch of suggestions. Some of the fixes make a lot of sense but some are mostly style issues. I split the fixes to a number of commits, so if need be, I can easily remove some fix classes where the benefit might not be clear.

One thing that came up is that there is a lot of code duplication between wasmtime and wasmedge crates. I'm thinking a utility library would be useful for sharing code. This way the possible bugs would be just in one place. :-)

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/cgroups/cgroupv1.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/cgroups/cgroupv1.rs
@@ -413,7 +413,7 @@ fn format_device(device: &LinuxDeviceCgroup) -> String {
 
     match device.minor() {
         Some(minor) => s.push_str(&minor.to_string()),
-        None => s.push_str("*"),
+        None => s.push('*'),
     }
     s.push(' ');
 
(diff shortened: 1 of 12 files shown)


rumpl:
> One thing that came up is that there is a lot of code duplication between wasmtime and wasmedge crates. I'm thinking a utility library would be useful for sharing code. This way the possible bugs would be just in one place. :-)

This is definintely something we should do, there is duplication between the two runtimes in this repository but also in the [slight and spin shims](https://github.com/deislabs/containerd-wasm-shims)

ipuustin:
> > One thing that came up is that there is a lot of code duplication between wasmtime and wasmedge crates. I'm thinking a utility library would be useful for sharing code. This way the possible bugs would be just in one place. :-)
> 
> This is definintely something we should do, there is duplication between the two runtimes in this repository but also in the [slight and spin shims](https://github.com/deislabs/containerd-wasm-shims)

Yeah, I can take a first pass at this effort. I think I'll first try to just move shared the code to a library in a reasonable way. We can then iterate on the suitable utility API interfaces.

rumpl:
youki does a great job at this with their [Executor trait](https://github.com/containers/youki/blob/main/crates/libcontainer/src/workload/mod.rs#L22-L29), basically the executor is called once everything is setup, in the default case it's a simple `execvp` and in the wasm* case all you need to do is instantiate the wasm runtime and run the module

rumpl (approved):

cpuguy83 (approved):

PR #60, opened 2023-01-31
Move call of the prestart hooks over to the lib part

rumpl:
Moving this makes pre-start hooks work for all the instances.

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/shim.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/shim.rs
@@ -958,6 +958,10 @@ where
 
         debug!("create done");
 
+        // Per the spec, the prestart hook must be called as part of the create operation
+        debug!("call prehook before the start");
+        oci::setup_prestart_hooks(&spec.hooks())?;
+
         Ok(api::CreateTaskResponse {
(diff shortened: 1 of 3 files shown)


crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -970,6 +970,19 @@ where
         }
 
         let i = self.get_instance(req.get_id())?;
+
+        if let Some(bundle) = i.cfg.get_bundle() {
```
rumpl, line 974: In the case of a cri pause container a new base instance is created without a bundle. I'm not sure I really understand this

cpuguy83: I haven't gone back and looked at the implementation (and it's been long enough I don't have it available in main memory 😄), but we pretend to start a pause container for cri since there's not really anything we can do with it.

rumpl:
I just remembered, the prestart hook is a weird one, [see this](https://github.com/opencontainers/runtime-spec/pull/1169), I'll update the code tomorrow...

cpuguy83 (approved):

PR #63, opened 2023-02-06
Inherit workspace's dependencies and other metadata

Mossaka:
This PR also modifies `use cap_std::path::PathBuf` to `std::fs::PathBuf` since it's removed from `cap_std` since `0.25`.

Signed-off-by: jiaxiao zhou <jiazho@microsoft.com>


rumpl (approved):
Nice!

PR #66, opened 2023-02-11
feat: add k3s test

defims:
k3s is newbee friendly and production available. To avoid polluting the system, the binaries of k3s and wasmedge are downloaded to the bin folder, and k3s service renamed to k3s-runwasi.
Since k3s only support linux, this test can only run in linux platform. For macos platform [multipass](https://multipass.run/) or [lima](https://github.com/lima-vm/lima) needed. For windows platform wsl2 needed.


Mossaka (approved):
lgtm! Thanks for your contribution!

PR #71, opened 2023-02-22
Update caps, protobuf deps and remove containerd-shim-protos

jsturtevant:
This updates a few of the dependencies in the project that were straight forward to update.  

I was trying to update to the ttrpc dependencies to the latest but the latest versions of ttrpc have new compiler and also bump the version of protobuf to 3.X.   We can't go to the newer versions that until we get a new release of `containerd-shim` from rust-extensions repo with the new changes.

When trying this out I noticed that the `contianerd-shim-protos` dependency was indirect but we were pulling it in directly so this fixes it be removing it.

I've also bumped `caps` which was a simple dependency update, whereas a few others required code changes.


devigned (approved):
lgtm

PR #73, opened 2023-02-22
Update nix and oci-spec dependencies

jsturtevant:
Updates nix and oci-spec dependencies by makeing the thrid party crate use the same dep as rest of the workspace.


Mossaka (approved):
lgtm!

Issue #23, opened 2022-11-18
Full Linux OCI runtime spec support


cpuguy83:
Right now we have only partial support for the OCI runtime spec.
While some things in the spec may not make sense for running wasm code itself, it is useful for sandboxing for the wasm runtime and/or the execution of the wasm for defense-in-depth as well as ensuring fewer surprises for users expecting their settings to actually apply.

Some things missing:

- [x] **cgroups**: In progress: https://github.com/deislabs/runwasi/pull/21
- [ ] **Lifecycle Hooks**: https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle https://github.com/opencontainers/runtime-spec/blob/86290f6a00fbdc6d561e14b2e6a11788a1a5f29c/config.md#posix-platform-hooks (Docker has at least some of this in their fork: https://github.com/second-state/runwasi/pull/13/files @rumpl)
- [ ] **hybrid cgroup**: In #21 there is support for cgroupv1 and cgroupv2, but not hybrid mode
- [ ] **systemd cgroup**: #21 now implements cgroupfs mode but does not support systemd cgroups
- [ ] **namespaces**: We do setup the network namespace, but other namespace support is still needed https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#namespaces
- [ ] **seccomp**: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#seccomp
- [ ] **apparmor**: https://github.com/opencontainers/runtime-spec/blob/86290f6a00fbdc6d561e14b2e6a11788a1a5f29c/config.md#linux-process
- [ ] **selinux**: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#mount-label https://github.com/opencontainers/runtime-spec/blob/86290f6a00fbdc6d561e14b2e6a11788a1a5f29c/config.md#linux-process
- [ ] **devices**: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#devices
- [ ] **sysctl**: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#sysctl
- [ ] **masked paths**: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#masked-paths
- [ ] **personality**: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#personality
[...]

PR #78, opened 2023-02-27
runwasi: Use youki to fully support OCI runtime spec in wasmedge

utam0k:
Partially Fix: https://github.com/containerd/runwasi/issues/23 (After this PR, I'm planning to create a PR to address wasmtime.)

Remaining issues
- [x] Wait for merge:
    - [x] https://github.com/containers/youki/pull/1637
    - [x] https://github.com/containers/youki/pull/1403
- [x] Pass all tests
- [x] Pass CIs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,6 @@ jobs:
       - name: "check cgroup version"
         run: "mount | grep cgroup"
       - uses: actions/checkout@v3
-        with:
-          submodules: true
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
@@ -55,8 +53,6 @@ jobs:
       - name: "check cgroup version"
         run: "mount | grep cgroup"
       - uses: actions/checkout@v3
-        with:
-          submodules: true
       - name: setup rust-wasm target
         run: rustup target add wasm32-wasi
       - name: run
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,9 +20,15 @@ jobs:
       - name: "check cgroup version"
         run: "mount | grep cgroup"
       - uses: actions/checkout@v3
+        with:
+          submodules: true
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
+      - name: Setup OCI runtime build env
+        run: |
+          sudo apt -y update
+          sudo apt install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev
       - name: Setup WasmEdge build env
         run: |
           curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash -s -- --version=0.11.2
@@ -35,10 +41,8 @@ jobs:
       - name: Validate docs
         run: ./scripts/validate-docs.sh
       - name: Run tests
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --all --verbose
+        run: |
+          sudo -E LD_LIBRARY_PATH=/home/runner/.wasmedge/lib /home/runner/.cargo/bin/cargo test --all --verbose
 
   e2e:
     needs: [build]
@@ -51,6 +55,8 @@ jobs:
       - name: "check cgroup version"
         run: "mount | grep cgroup"
       - uses: actions/checkout@v3
+        with:
+          submodules: true
       - name: setup rust-wasm target
         run: rustup target add wasm32-wasi
       - name: run
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "youki"]
 	path = youki
 	url = git@github.com:utam0k/youki.git
+    branch = runwasi
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,0 @@
-[submodule "youki"]
-	path = youki
-	url = git@github.com:utam0k/youki.git
-    branch = runwasi
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "youki"]
+	path = youki
+	url = git@github.com:utam0k/youki.git
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,8 +29,9 @@ nix = "0.26"
 cap-std = "1.0"
 thiserror = "1.0"
 libc = "0.2.138"
-oci-spec = "0.6"
+oci-spec = { version = "^0.5.5", features = ["runtime"] }
 sha256 = "1.1"
+libcontainer = { path = "youki/crates/libcontainer" }
 
 [profile.release]
 panic = "abort"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,9 +29,9 @@ nix = "0.26"
 cap-std = "1.0"
 thiserror = "1.0"
 libc = "0.2.138"
-oci-spec = { version = "^0.5.5", features = ["runtime"] }
+oci-spec = { version = "0.6.0", features = ["runtime"] }
 sha256 = "1.1"
-libcontainer = { path = "youki/crates/libcontainer" }
+libcontainer = { git = "https://github.com/containers/youki", rev = "b469fd12e8aaf2b202432cd1476cb509c751e513" }
 
 [profile.release]
 panic = "abort"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,10 +28,10 @@ chrono = "0.4"
 nix = "0.26"
 cap-std = "1.0"
 thiserror = "1.0"
-libc = "0.2.138"
+libc = "0.2.141"
 oci-spec = { version = "0.6.0", features = ["runtime"] }
 sha256 = "1.1"
-libcontainer = { git = "https://github.com/containers/youki", rev = "b469fd12e8aaf2b202432cd1476cb509c751e513" }
+libcontainer = { git = "https://github.com/utam0k/youki", rev = "03eb78d13a6493b1184aa40fb96c73d882975956" }
 
 [profile.release]
 panic = "abort"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,8 @@ thiserror = "1.0"
 libc = "0.2.141"
 oci-spec = { version = "0.6.0", features = ["runtime"] }
 sha256 = "1.1"
-libcontainer = { git = "https://github.com/utam0k/youki", rev = "03eb78d13a6493b1184aa40fb96c73d882975956" }
+# TODO: Wait for the release and use the crates.io one.
+libcontainer = { git = "https://github.com/containers/youki", rev = "edd63c84f903aa09510ef758ea6f617e0cb8b7e1" }
 
 [profile.release]
 panic = "abort"
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ ENV WASMEDGE_INCLUDE_DIR=/root/.wasmedge/include
 ENV WASMEDGE_LIB_DIR=/root/.wasmedge/lib
 ENV LD_LIBRARY_PATH=/root/.wasmedge/lib
 RUN xx-apt-get install -y gcc g++ libc++6-dev zlib1g
+RUN xx-apt-get install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev
 RUN rustup target add $(xx-info march)-unknown-$(xx-info os)-$(xx-info libc)
 RUN <<EOT
     set -ex
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ In either case you need to implement the `Instance` trait:
 ```rust
 pub trait Instance {
     /// Create a new instance
-    fn new(id: String, rootdir: String, cfg: Option<&InstanceConfig<Self::E>>) -> Self;
+    fn new(id: String, cfg: Option<&InstanceConfig<Self::E>>) -> Self;
     /// Start the instance
     /// The returned value should be a unique ID (such as a PID) for the instance.
     /// Nothing internally should be using this ID, but it is returned to containerd where a user may want to use it.
--- a/README.md
+++ b/README.md
@@ -27,15 +27,21 @@ In either case you need to implement the `Instance` trait:
 
 ```rust
 pub trait Instance {
-    // Create a new instance
-    fn new(id: String, cfg: Option<&InstanceConfig<Self::E>>) -> Self;
-    // Start the instance and return the pid
+    /// Create a new instance
+    fn new(id: String, rootdir: String, cfg: Option<&InstanceConfig<Self::E>>) -> Self;
+    /// Start the instance
+    /// The returned value should be a unique ID (such as a PID) for the instance.
+    /// Nothing internally should be using this ID, but it is returned to containerd where a user may want to use it.
     fn start(&self) -> Result<u32, Error>;
-    // Send the specified signal to the instance
+    /// Send a signal to the instance
     fn kill(&self, signal: u32) -> Result<(), Error>;
-    // Delete the instance
+    /// delete any reference to the instance
+    /// This is called after the instance has exited.
     fn delete(&self) -> Result<(), Error>;
-    // wait for the instance to exit and send the exit code and exit timestamp to the provided sender.
+    /// wait for the instance to exit
+    /// The sender is used to send the exit code and time back to the caller
+    /// Ideally this would just be a blocking call with a normal result, however
+    /// because of how this is called from a thread it causes issues with lifetimes of the trait implementer.
     fn wait(&self, send: Sender<(u32, DateTime<Utc>)>) -> Result<(), Error>;
 }
 ```
--- a/crates/containerd-shim-wasm/src/sandbox/shim.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/shim.rs
@@ -39,6 +39,8 @@ use nix::unistd::mkdir;
 use oci_spec::runtime;
 use ttrpc::context::Context;
 
+static CONTAINER_ROOT_DIR: &str = "/var/run/runwasi";
+
 type InstanceDataStatus = (Mutex<Option<(u32, DateTime<Utc>)>>, Condvar);
 
 struct InstanceData<T: Instance<E = E>, E>
@@ -736,7 +738,7 @@ where
         let cfg = InstanceConfig::new(self.engine.clone());
         InstanceData {
             instance: None,
-            base: Some(Nop::new(id, None)),
+            base: Some(Nop::new(id, CONTAINER_ROOT_DIR.into(), None)),
             cfg,
             pid: RwLock::new(None),
             status: Arc::new((Mutex::new(None), Condvar::new())),
@@ -933,7 +935,11 @@ where
         self.instances.write().unwrap().insert(
             req.get_id().to_string(),
             Arc::new(InstanceData {
-                instance: Some(T::new(req.get_id().to_string(), Some(&builder))),
+                instance: Some(T::new(
+                    req.get_id().to_string(),
+                    CONTAINER_ROOT_DIR.into(),
+                    Some(&builder),
+                )),
                 base: None,
                 cfg: builder,
                 pid: RwLock::new(None),
--- a/crates/containerd-shim-wasm/src/sandbox/shim.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/shim.rs
@@ -39,8 +39,6 @@ use nix::unistd::mkdir;
 use oci_spec::runtime;
 use ttrpc::context::Context;
 
-static CONTAINER_ROOT_DIR: &str = "/var/run/runwasi";
-
 type InstanceDataStatus = (Mutex<Option<(u32, DateTime<Utc>)>>, Condvar);
 
 struct InstanceData<T: Instance<E = E>, E>
@@ -738,7 +736,7 @@ where
         let cfg = InstanceConfig::new(self.engine.clone());
         InstanceData {
             instance: None,
-            base: Some(Nop::new(id, CONTAINER_ROOT_DIR.into(), None)),
+            base: Some(Nop::new(id, None)),
             cfg,
             pid: RwLock::new(None),
             status: Arc::new((Mutex::new(None), Condvar::new())),
@@ -935,11 +933,7 @@ where
         self.instances.write().unwrap().insert(
             req.get_id().to_string(),
             Arc::new(InstanceData {
-                instance: Some(T::new(
-                    req.get_id().to_string(),
-                    CONTAINER_ROOT_DIR.into(),
-                    Some(&builder),
-                )),
+                instance: Some(T::new(req.get_id().to_string(), Some(&builder))),
                 base: None,
                 cfg: builder,
                 pid: RwLock::new(None),
--- a/crates/containerd-shim-wasm/src/sandbox/shim.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/shim.rs
@@ -338,6 +338,7 @@ where
     instances: LocalInstances<T, E>,
     events: Arc<Mutex<EventSender>>,
     exit: Arc<ExitSignal>,
+    namespace: String,
 }
 
 #[cfg(test)]
@@ -422,6 +423,7 @@ mod localtests {
             (),
             tx,
             Arc::new(ExitSignal::default()),
+            "test_namespace".into(),
         ));
         let mut _wrapped = LocalWithDescrutor::new(local.clone());
 
@@ -447,7 +449,12 @@ mod localtests {
         // When a cri sandbox is specified we just assume it's the sandbox container and treat it as such by not actually running the code (which is going to be wasm).
         let (etx, _erx) = channel();
         let exit_signal = Arc::new(ExitSignal::default());
-        let local = Arc::new(Local::<Nop, _>::new((), etx, exit_signal));
+        let local = Arc::new(Local::<Nop, _>::new(
+            (),
+            etx,
+            exit_signal,
+            "test_namespace".into(),
+        ));
 
         let mut _wrapped = LocalWithDescrutor::new(local.clone());
 
@@ -610,7 +617,12 @@ mod localtests {
     fn test_task_lifecycle() -> Result<()> {
         let (etx, _erx) = channel(); // TODO: check events
         let exit_signal = Arc::new(ExitSignal::default());
-        let local = Arc::new(Local::<Nop, _>::new((), etx, exit_signal));
+        let local = Arc::new(Local::<Nop, _>::new(
+            (),
+            etx,
+            exit_signal,
+            "test_namespace".into(),
+        ));
 
         let mut _wrapped = LocalWithDescrutor::new(local.clone());
 
@@ -718,7 +730,12 @@ where
     E: Send + Sync + Clone,
 {
     /// Creates a new local task service.
-    pub fn new(engine: E, tx: Sender<(String, Box<dyn Message>)>, exit: Arc<ExitSignal>) -> Self
+    pub fn new(
+        engine: E,
+        tx: Sender<(String, Box<dyn Message>)>,
+        exit: Arc<ExitSignal>,
+        namespace: String,
+    ) -> Self
     where
         T: Instance<E = E> + Sync + Send,
         E: Send + Sync + Clone,
@@ -729,11 +746,12 @@ where
             instances: Arc::new(RwLock::new(HashMap::new())),
             events: Arc::new(Mutex::new(tx)),
             exit,
+            namespace,
         }
     }
 
     fn new_base(&self, id: String) -> InstanceData<T, E> {
-        let cfg = InstanceConfig::new(self.engine.clone());
+        let cfg = InstanceConfig::new(self.engine.clone(), self.namespace.clone());
         InstanceData {
             instance: None,
             base: Some(Nop::new(id, None)),
@@ -924,7 +942,7 @@ where
         }
 
         let engine = self.engine.clone();
-        let mut builder = InstanceConfig::new(engine);
+        let mut builder = InstanceConfig::new(engine, self.namespace.clone());
         builder
             .set_stdin(req.get_stdin().into())
             .set_stdout(req.get_stdout().into())
@@ -1182,8 +1200,13 @@ where
     type Instance = T;
     fn new(namespace: String, _id: String, engine: E, publisher: RemotePublisher) -> Self {
         let (tx, rx) = channel::<(String, Box<dyn Message>)>();
-        forward_events(namespace, publisher, rx);
-        Local::<T, E>::new(engine, tx.clone(), Arc::new(ExitSignal::default()))
+        forward_events(namespace.clone(), publisher, rx);
+        Local::<T, E>::new(
+            engine,
+            tx.clone(),
+            Arc::new(ExitSignal::default()),
+            namespace,
+        )
     }
 }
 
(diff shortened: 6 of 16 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "youki"]
```
utam0k, line 1: It is for the efficiency of development and debugging and will eventually be deleted.

utam0k:
Current Status.
```console
$ cargo test --no-run --message-format=json | jq -r "select(.profile.test == true) | .filenames[]" | while read -r bin; do;
 sudo LD_LIBRARY_PATH=/home/utam0k/.wasmedge/lib "$bin"
done
[...]
running 3 tests
test instance::tests::test_maybe_open_stdio ... ok
test instance::wasitest::test_delete_after_create ... ok
test instance::wasitest::test_wasi ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.14s
```

Dockerfile
```diff
@@ -16,6 +16,7 @@ ENV WASMEDGE_INCLUDE_DIR=/root/.wasmedge/include
 ENV WASMEDGE_LIB_DIR=/root/.wasmedge/lib
 ENV LD_LIBRARY_PATH=/root/.wasmedge/lib
 RUN xx-apt-get install -y gcc g++ libc++6-dev zlib1g
+RUN xx-apt-get install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev
```
ipuustin, line 19: I think libcontainer doesn't depend on libsystemd-dev or libdbus-glib-1-dev? Or does it?

crates/wasmedge/src/instance.rs
```diff
@@ -185,89 +152,92 @@ impl Instance for Wasi {
...
-        let stderr = self.stderr.clone();
 
+    fn start(&self) -> Result<u32, Error> {
         debug!("preparing module");
-        let spec = load_spec(self.bundle.clone())?;
-
-        let vm = prepare_module(engine, &spec, stdin, stdout, stderr)
-            .map_err(|e| Error::Others(format!("error setting up module: {}", e)))?;
-
-        let cg = oci::get_cgroup(&spec)?;
```
ipuustin, line 200: Are the oci functions such as `get_cgroup`, `setup_cgroup` etc. used elsewhere, or can we drop them too?

ipuustin (commented):
Looks cool! Couple of questions:
1. You say this partially fixes #23 -- can you tell which parts will be missing?
2. Does this cause any functional changes outside of the spec? I mean, does Youki use the three-process model now for running the VM? Is the namespace isolation the same?

I was preparing some seccomp support in https://github.com/ipuustin/runwasi/commits/seccomp. I'm thinking I'll probably pause that work until this PR merges and then add the unit tests and other potentially useful stuff on top of this PR.

Cargo.toml
```diff
@@ -30,8 +29,9 @@ nix = "0.26"
 cap-std = "0.26"
 thiserror = "1.0"
 libc = "0.2.138"
-oci-spec = "0.6"
+oci-spec = { version = "^0.5.5", features = ["runtime"] }
```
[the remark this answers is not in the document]
utam0k: Thanks for your comment. I'll check it later

Dockerfile
utam0k: `libcontainer` support systemd-cgroup driver, so it depends on them.

crates/wasmedge/src/instance.rs
utam0k: Since `libcontainer` supports set-up cgroup, I think don't need to take care of cgroup by runwasi.

utam0k:
:wave: Hi, @ipuustin. Thanks for your comment.

We need to support wasmtime after this PR.
> You say this partially fixes https://github.com/containerd/runwasi/issues/23 -- can you tell which parts will be missing?

Basically the same. Youki implements the oci-spec following runc's implementation, but more simply.
> Does this cause any functional changes outside of the spec? I mean, does Youki use the three-process model now for running the VM? Is the namespace isolation the same?

I agree with you. Sorry for the inconvenience.
> I was preparing some seccomp support in https://github.com/ipuustin/runwasi/commits/seccomp. I'm thinking I'll probably pause that work until this PR merges and then add the unit tests and other potentially useful stuff on top of this PR.

Cargo.toml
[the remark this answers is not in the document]
utam0k: It looks youki that was on this PR was too old.

.github/workflows/ci.yml
```diff
@@ -20,9 +20,15 @@ jobs:
       - name: "check cgroup version"
         run: "mount | grep cgroup"
       - uses: actions/checkout@v3
+        with:
+          submodules: true
```
utam0k, line 24: See: 
https://github.com/containerd/runwasi/pull/78/files#r1118520073

Dockerfile
ipuustin: Ok I see, there's the `systemd` feature for libcontainer.

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -371,15 +341,19 @@ mod wasitest {
...
+        let wasi = Arc::new(Wasi::new(
+            "test".to_string(),
+            rootdir.into_os_string().into_string().unwrap(),
+            Some(cfg),
+        ));
 
         wasi.start()?;
 
-        let w = wasi.clone();
+        // let w = wasi.clone();
```
ipuustin, line 352: Just drop commented out code -- git will remember those if need be.

ipuustin (changes requested):
I think this concept looks good! When do you think next libcontainer release will be -- should we wait until that or use the libcontainer commit from main branch?

.github/workflows/ci.yml
```diff
@@ -35,10 +39,8 @@ jobs:
       - name: Validate docs
         run: ./scripts/validate-docs.sh
       - name: Run tests
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --all --verbose
+        run: |
+          sudo -E LD_LIBRARY_PATH=/home/runner/.wasmedge/lib /home/runner/.cargo/bin/cargo test --all --verbose
```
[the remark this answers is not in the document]
utam0k: You need to run it as a root user.

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -103,171 +131,103 @@ pub fn reset_stdio() {
...
+                }
+                Err(Error::Others(e.to_string()))
+            }
+        }
     }
 
     fn delete(&self) -> Result<(), Error> {
-        let spec = match load_spec(self.bundle.clone()) {
-            Ok(spec) => spec,
+        match container_exists(self.rootdir.as_str(), self.id.as_str()) {
```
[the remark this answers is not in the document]
utam0k: If a container is present to meet OCI specifications, it must return Ok.

.github/workflows/ci.yml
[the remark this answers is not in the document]
ipuustin: Yes I see! You could maybe do something like this to run the tests with sudo:

```patch
diff --git a/crates/containerd-shim-wasm/src/sandbox/mod.rs b/crates/containerd-shim-wasm/src/sandbox/mod.rs
index e807480..72d7ac2 100644
--- a/crates/containerd-shim-wasm/src/sandbox/mod.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/mod.rs
@@ -16,5 +16,4 @@ pub use shim::{Cli as ShimCli, Local};

 pub mod oci;

-#[cfg(test)]
-mod testutil;
+pub mod testutil;
diff --git a/crates/containerd-shim-wasm/src/sandbox/testutil.rs b/crates/containerd-shim-wasm/src/sandbox/testutil.rs
index a2c1f86..baa61ee 100644
--- a/crates/containerd-shim-wasm/src/sandbox/testutil.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/testutil.rs
@@ -41,6 +41,7 @@ pub fn run_test_with_sudo(test: &str) -> Result<()> {
         .map_err(Error::from)
 }

+#[macro_export]
 macro_rules! function {
     () => {{
         fn f() {}
@@ -53,4 +54,4 @@ macro_rules! function {
     }};
 }

-pub(crate) use function;
+pub use function;
diff --git a/crates/containerd-shim-wasmedge/src/instance.rs b/crates/containerd-shim-wasmedge/src/instance.rs
index 64c9987..7e1b8f5 100644
--- a/crates/containerd-shim-wasmedge/src/instance.rs
+++ b/crates/containerd-shim-wasmedge/src/instance.rs
@@ -256,6 +256,9 @@ mod wasitest {
     use std::sync::mpsc::channel;
     use std::time::Duration;

+    use containerd_shim_wasm::function;
+    use containerd_shim_wasm::sandbox::exec::has_cap_sys_admin;
+    use containerd_shim_wasm::sandbox::testutil::run_test_with_sudo;
     use oci_spec::runtime::{ProcessBuilder, RootBuilder, SpecBuilder};
     use tempfile::{tempdir, TempDir};

@@ -383,6 +386,10 @@ mod wasitest {
     #[test]
     #[serial]
     fn test_wasi() -> Result<(), Error> {
+        if !has_cap_sys_admin() {
+            println!("running test with sudo: {}", function!());
+            return run_test_with_sudo(function!());
+        }
         let dir = tempdir()?;
         let path = dir.path();
         let wasmbytes = wat2wasm(WASI_HELLO_WAT).unwrap();
@@ -401,6 +408,10 @@ mod wasitest {
     #[test]
     #[serial]
     fn test_wasi_error() -> Result<(), Error> {
+        if !has_cap_sys_admin() {
+            println!("running test with sudo: {}", function!());
+            return run_test_with_sudo(function!());
+        }
         let dir = tempdir()?;
         let wasmbytes = wat2wasm(WASI_RETURN_ERROR).unwrap();
```

crates/containerd-shim-wasmedge/src/instance.rs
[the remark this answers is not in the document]
ipuustin: Sure, I was just thinking that they are called one after another, so this could maybe be a place for optimization. But this is a very minor thing and I'm fine either way.

utam0k:
> I think this concept looks good! When do you think next libcontainer release will be -- should we wait until that or use the libcontainer commit from main branch?

@ipuustin Thanks for your comment. We may release it soon. However, I would like to merge once with this commit as specified. This is because I am afraid of conflicts. I have already done several major conflict resolutions on this PR.

.github/workflows/ci.yml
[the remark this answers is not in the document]
utam0k: @ipuustin I love this idea. I have applied for it.

crates/containerd-shim-wasmedge/src/executor.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub stdin: Option<RawFd>,
+    pub stdout: Option<RawFd>,
+    pub stderr: Option<RawFd>,
+}
+
+impl Executor for WasmEdgeExecutor {
+    fn exec(&self, spec: &Spec) -> Result<()> {
+        // parse wasi parameters
+        let args = get_args(spec);
+        let mut cmd = args[0].clone();
```
ipuustin, line 29: Hmm, what if `args` is empty?

crates/containerd-shim-wasmedge/src/executor.rs
```diff
@@ -0,0 +1,117 @@
...
+            .func("_start")
+            .expect("Not found '_start' func in the 'main' module instance")
+            .call(&mut vm, params!())
+        {
+            Ok(_) => std::process::exit(0),
+            Err(_) => std::process::exit(137),
+        };
+    }
+
+    fn can_handle(&self, _spec: &Spec) -> Result<bool> {
```
ipuustin, line 86: This is an interesting function especially if we will later on do things such as map workloads to runtimes which can support the features the workload uses.

ipuustin (approved):
I think this looks good now. Going forward we need to add a bunch of tests trying out the various OCI spec features, which I suppose will be good for both this project and libcontainer too.

crates/containerd-shim-wasmedge/src/executor.rs
utam0k: Maybe we need to validate it?

crates/containerd-shim-wasmedge/src/executor.rs
utam0k: Initially, this trait is designed for youki.
https://github.com/containers/youki/blob/2260c1ea776584140c659d166888addaa73b936d/crates/youki/src/workload/wasmedge.rs#L51-L63

README.md
devigned, line 1: Thank you so much for adding to the comments. The additional context is super helpful for a naive observer approaching the topic.

devigned (changes requested):
This PR looks awesome. Thank you so much, @utam0k!!

Before it is merged, I think we would like to resolve: https://github.com/containerd/runwasi/pull/78#discussion_r1140069567, and I'd like @cpuguy83 to give it a review.

Really fantastic work here. Thank you again.

utam0k:
@devigned Thanks for your review. PTAL

crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -39,6 +39,8 @@ use nix::unistd::mkdir;
 use oci_spec::runtime;
 use ttrpc::context::Context;
 
+static CONTAINER_ROOT_DIR: &str = "/var/run/runwasi";
```
cpuguy83, line 42: I can't help but think this should be specified by the implementation, ie `/run/io.containerd.wasmedge.v1` OR from the spec?

cpuguy83: It's also not clear to me what this is used for

cpuguy83 (changes requested):

devigned (approved):
lgtm!

I'm +1 pending @cpuguy83 feedback. Great work, @utam0k!

crates/containerd-shim-wasm/src/sandbox/shim.rs
utam0k: This directory is the same role as runc; it stores files for managing the state of the container and signaling start.
If runwasi is to be considered only as an OCI runtime, I don't think the name containerd should be included in the directory. I think that including it may compromise its versatility.
https://github.com/opencontainers/runc/blob/efad7a3b80f4585065f93aad3e91989da5012d63/man/runc.8.md?plain=1#L108-L111

cpuguy83: In our case containerd is passing down a state dir to the shim that we should be using rather than coming up with our own.

I can't exactly recall how we setting that dir right now but since this crate is only a library I think we should leave any sort of state dir to a higher level.

cpuguy83:
Requesting t hat we keep all other merges on hold until this is in.
Looks like we need to regenerate Cargo.lock due to a merge conflict.

crates/containerd-shim-wasm/src/sandbox/shim.rs
utam0k: @cpuguy83 Thanks for your comment 🙇 That is sort of an ideal, but when it comes to adjusting to reality it takes time. Many OCI Runtimes already require a state command. This command is inevitably stateful. Also, it is an OCI specification that `create` command and `start` command are two different commands: `create` creates the container process and the `start` command sent a message to that process to start the container. In other words, the `start` command must leave an interface somewhere to interact with the container created by create. That interface is left in this directory.
https://containers.github.io/youki/user/liboci_cli.html

utam0k:
> Requesting t hat we keep all other merges on hold until this is in. Looks like we need to regenerate Cargo.lock due to a merge conflict.

Sorry for a lof of changes. I have resolved it.

crates/containerd-shim-wasm/src/sandbox/shim.rs
cpuguy83: @utam0k Yes I understand that.
For us, containerd passes down the state.
As an example when you create a container with containerd, it's not using the default runc state dir because containerd (or more precisely whoever started the container) is managing that path.

If it's docker `/run/docker/runtime-runc/moby` (where `moby` is the containerd namespace) because docker customizes the path when it invokes containerd. 

```console
# jq . /run/containerd/io.containerd.runtime.v2.task/moby/01e48c19c2af1da405f209b79fcc10542372ab82fb081ad80f713826797e6ded/options.json
{
  "binary_name": "runc",
  "root": "/var/run/docker/runtime-runc",
  "systemd_cgroup": true
}
```

If it's ctr it's `/run/containerd/runc/default/test` (where `default` is the containerd namespace and `test` is the container id).
ctr does not customize this path so it's just the default.

```console
# jq . /run/containerd/io.containerd.runtime.v2.task/default/test/options.json 
{}
```

So we should expect the actual shim implementation to pass down this state dir from higher up the stack.
The above mentioned directories is the OCI bundle path.
So we should be able to check for a state dir in this file and 

We can choose a default path, but it also must be namespaced by the containerd namespace or else we can end up with conflicts.

cpuguy83: The other part of this is we have a long running daemon managing state and should not need to persist it across calls.

When the daemon dies so do the containers it is managing.

utam0k: @cpuguy83 Make sense a lot 💯 I have updated my PR!

utam0k:
@cpuguy83 PTAL

Mossaka (commented):
Mostly LGTM! I have some questions and most of them are just for me to understand `libcontainer` more. I would greatly appreicate your answers.

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -103,171 +134,158 @@ pub fn reset_stdio() {
...
 
-    debug!("opening stdout");
-    let stdout = maybe_open_stdio(&stdout_path).context("could not open stdout")?;
-    if stdout.is_some() {
-        unsafe {
-            STDOUT_FD = Some(dup(STDOUT_FILENO));
-            dup2(stdout.unwrap(), STDOUT_FILENO);
-        }
-    }
+fn determine_rootdir<P: AsRef<Path>>(bundle: P) -> Result<PathBuf, Error> {
```
cpuguy83, line 142: We also need to account for the containerd namespace otherwise we could end up with 2 containers with the same ID but different namespaces writing to the same state dir.

cpuguy83 (changes requested):
I think as a whole I was hoping to see this integration at a lower level than youki's Executor interface so that we can have more control over the runtime environment and reduce on-disk state since we aren't managing linux processes (outside of what's needed to deal with cgroups).
I think for now (after the requested changes are made) this will work out well.

It will be interesting to see projects like slight and spin adopt similar changes since they are more function handler based and spinning up a new process per function is not what would be desired.
But again, as is this is probably fine and we can always adjust as needed.

cpuguy83:
@rumpl

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -103,171 +134,158 @@ pub fn reset_stdio() {
...
-                let _ret = match vm.run_func(Some("main"), "_start", params!()) {
-                    Ok(_) => std::process::exit(0),
-                    Err(_) => std::process::exit(137),
-                };
-            }
-        }
+
+        let syscall = create_syscall();
+        let mut container = ContainerBuilder::new(self.id.clone(), syscall.as_ref())
+            .with_executor(vec![Box::new(WasmEdgeExecutor {
```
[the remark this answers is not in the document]
utam0k: It can be done; youki actually does. 
https://github.com/containers/youki/blob/d1e19fdbe3229999c287949e31615b4251c90d17/crates/youki/src/workload/executor.rs#L3-L13

At an abstract level, you can implement what is to be executed in the container.

Cargo.toml
```diff
@@ -29,8 +29,9 @@ nix = "0.26"
 cap-std = "1.0"
 thiserror = "1.0"
 libc = "0.2.138"
-oci-spec = "0.6"
+oci-spec = { version = "0.6.0", features = ["runtime"] }
 sha256 = "1.1"
+libcontainer = { git = "https://github.com/containers/youki", rev = "b469fd12e8aaf2b202432cd1476cb509c751e513" }
```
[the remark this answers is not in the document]
utam0k: Yes, it did but we haven't released the latest version runwasi needs yet. 
https://crates.io/crates/libcontainer

.github/workflows/ci.yml
```diff
@@ -23,6 +23,10 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
+      - name: Setup OCI runtime build env
```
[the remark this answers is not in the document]
utam0k: Make sense

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -103,171 +134,158 @@ pub fn reset_stdio() {
...
+                stdin: maybe_open_stdio(self.stdin.as_str()).context("could not open stdin")?,
+                stdout: maybe_open_stdio(self.stdout.as_str()).context("could not open stdout")?,
+                stderr: maybe_open_stdio(self.stderr.as_str()).context("could not open stderr")?,
+            })])?
+            .with_root_path(self.rootdir.clone())?
```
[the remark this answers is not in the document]
utam0k: @Mossaka Sorry, may I ask you to elaborate a bit more on your question?

utam0k:
> I think as a whole I was hoping to see this integration at a lower level than youki's Executor interface so that we can have more control over the runtime environment and reduce on-disk state since we aren't managing linux processes (outside of what's needed to deal with cgroups). I think for now (after the requested changes are made) this will work out well.
> 
> It will be interesting to see projects like slight and spin adopt similar changes since they are more function handler based and spinning up a new process per function is not what would be desired. But again, as is this is probably fine and we can always adjust as needed.

@cpuguy83 
I don't know much about `slight` and `spin`, so I can't comment much.
Just, the biggest weakness of this architecture is its strong dependence on `youki`/`libcontainer`. There are many ways to think about this. The license is quite loose, so if something goes wrong, you can fork in the worst case. I can also participate in the development of runwasi if needed, but only in my spare time (youki is my hobby too, and being vendor neutral is what I would like). I look forward to a future with runwasi.

utam0k:
@Mossaka @cpuguy83 Thanks for your review. I know it is hard to review because the changes are so significant. PTAL

crates/containerd-shim-wasmedge/src/instance.rs
utam0k: I made it consider the namespace.
https://github.com/containerd/runwasi/pull/78/commits/3fc44d2df7a0e3d52a174d92e6316909b87c7c62

crates/containerd-shim-wasmedge/src/instance.rs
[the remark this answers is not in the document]
Mossaka: Thanks for the link! It looks like three out of 4 executors are wasm executors. I am assuming the default executor is the one that can execute normal container workloads, right?

rumpl:
I didn't have the time to take a look at the code yet, will look closely tomorrow but one thing that comes to mind is: why are we using this at the wasmedge shim level? I would have thought that the runwasi library would be the one that supports the OCI runtime spec and all that the shims themselves would have to do is to handle the module that needs to run.

crates/containerd-shim-wasmedge/src/instance.rs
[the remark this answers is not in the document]
rumpl: Yeah, the `DefaultExecutor` is the one that executes the container's process https://github.com/containers/youki/blob/d1e19fdbe3229999c287949e31615b4251c90d17/crates/libcontainer/src/workload/default.rs#L14

rumpl:
I just tested this with docker and I get an error:

```console
docker run --runtime=io.containerd.wasmedge.v1 was
docker: Error response from daemon: failed to create container: unknown.
ERRO[0000] error waiting for container:   
```

In the daemon logs I don't see anything that could help

```console
DEBU[2023-03-28T17:31:39.512641886+02:00] event                                         module=libcontainerd namespace=moby topic=/tasks/create
ERRO[2023-03-28T17:31:39.512738900+02:00] failed to process event                       container=80470a7712ba6ba97f29169ae886d45620569d8fd5a99a52cb7c5ffb391eb078 error="could not find container 80470a7712ba6ba97f29169ae886d45620569d8fd5a99a52cb7c5ffb391eb078: No such container: 80470a7712ba6ba97f29169ae886d45620569d8fd5a99a52cb7c5ffb391eb078" event=create event-info="{80470a7712ba6ba97f29169ae886d45620569d8fd5a99a52cb7c5ffb391eb078 80470a7712ba6ba97f29169ae886d45620569d8fd5a99a52cb7c5ffb391eb078 0 0 0001-01-01 00:00:00 +0000 UTC <nil>}" module=libcontainerd namespace=moby
```

I'll keep digging some more.

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -103,171 +135,170 @@ pub fn reset_stdio() {
...
+                    }
+                }
+            } as u32;
+            let mut ec = lock.lock().unwrap();
+            *ec = Some((status, Utc::now()));
+            drop(ec);
+            cvar.notify_all();
+        });
+
+        Ok(0)
```
rumpl, line 260: The docs say that this value should be a unique ID:

https://github.com/containerd/runwasi/blob/fff59017205828e1835ea7f699b7a0c5e7b924be/crates/containerd-shim-wasm/src/sandbox/instance.rs#L105-L108

rumpl:
Little update, I don't know yet why this fails. With docker the container creation fails, I'm trying to figure out where and why it fails. Somewhere we get a `ECHILD` error deep inside youki's code.

It doesn't really surprise me though, the container creation sets up all the OCI stuff but we are already setting some of the things in our sandbox...

rumpl:
Right, in this case it's the `waitpid` that fails:

https://github.com/containers/youki/blob/04de7bf3b144da81f991d5756d5b284d4c87cec3/crates/libcontainer/src/process/container_main_process.rs#L110-L119

Removing these lines from youki make it work (for a hello-world example):
```console
$ docker run --runtime=io.containerd.wasmedge.v1 rumpl/was
Hello, world!
```

ipuustin:
> Right, in this case it's the `waitpid` that fails:
> 
> https://github.com/containers/youki/blob/04de7bf3b144da81f991d5756d5b284d4c87cec3/crates/libcontainer/src/process/container_main_process.rs#L110-L119
> 
> Removing these lines from youki make it work (for a hello-world example):

You mean `waitpid()` returns an error? The waitpid docs say:

> [ECHILD]
> The calling process has no existing unwaited-for child processes.

Is this the errno that `waitpid()` sets? Youki could check if the process has already been waited by runwasi, because as far as I understand the waitpid check is for synchronization purposes. (The same check maybe should be added to runwasi too so that there wouldn't be a race condition.)

rumpl:
> Is this the errno that `waitpid()` sets?

Yes, `waitpid` returns the `ECHILD` error in that code with docker

utam0k:
@rumpl hmm... Please give me instructions on how to reproduce this error.

rumpl:
@utam0k You can download the 24.0.0-beta-1 version of dockerd from download.docker.com (for example for Debian it would be here: https://download.docker.com/linux/debian/dists/bullseye/pool/test/amd64/docker-ce-cli_24.0.0~beta.1-1~debian.11~bullseye_amd64.deb, you'll have to look around for your dist or use the static ones).

Then you can:

* Edit the `/etc/docker/daemon.json` file and add `{ "features": { "containerd-snapshotter": true } }`, 
* Run the daemon with `sudo ./dockerd -D -H unix:///tmp/docker.sock --data-root /tmp/dockerd --pidfile /tmp/docker.pid`
* Create a context that will connect to this daemon: `docker context create test --docker "host=unix:///tmp/test.sock"`
* Use that context: `docker context use test` 
* Run a wasm workload with docker.

Once you finished you can go back to your original context: `docker context use default`

squillace:
hey @utam0k I pinged you on discord, but dropping it here, too: we'd love to work with you and @rumpl to get this working on Docker Desktop -- if we can get that, we can merge and move on to showing the beautiful magic. Is that something you can work on? We totally understand if not -- life is complex -- but then we might need to reopen the repo to merge other prs and try this one again later..... 

What do you think?

utam0k:
> hey @utam0k I pinged you on discord, but dropping it here, too: we'd love to work with you and @rumpl to get this working on Docker Desktop -- if we can get that, we can merge and move on to showing the beautiful magic. Is that something you can work on? We totally understand if not -- life is complex -- but then we might need to reopen the repo to merge other prs and try this one again later.....
> 
> What do you think?

I have already replied to that via discord.

utam0k:
> > Is this the errno that `waitpid()` sets?
> 
> Yes, `waitpid` returns the `ECHILD` error in that code with docker

@rumpl How did you know `ECHILD`?

utam0k:
@rumpl cc: @rumpl  
I send out the PR to address that case.
https://github.com/containers/youki/pull/1777

utam0k:
@rumpl PTAL

crates/containerd-shim-wasmedge/src/instance.rs
utam0k: @rumpl 
I have updated this branch to fix it! Thanks for your review.

Cargo.toml
```diff
@@ -28,9 +28,10 @@ chrono = "0.4"
 nix = "0.26"
 cap-std = "1.0"
 thiserror = "1.0"
-libc = "0.2.138"
-oci-spec = "0.6"
+libc = "0.2.141"
+oci-spec = { version = "0.6.0", features = ["runtime"] }
 sha256 = "1.1"
+libcontainer = { git = "https://github.com/utam0k/youki", rev = "03eb78d13a6493b1184aa40fb96c73d882975956" }
```
rumpl, line 34: Since [this PR](https://github.com/containers/youki/pull/1777) is merged now maybe this can be containers/youki rather than your fork?

0xE282B0:
Same result here:
> docker: Error response from daemon: failed to load spec: unknown.
> ERRO[0000] error waiting for container:                 

The last thing that is executed is the [ContainerBuilder[...].build()?;](
https://github.com/utam0k/runwasi/blob/libcontainer-wasmedge/crates/containerd-shim-wasmedge/src/instance.rs#L224-L233) when using docker. However, when running the same container with only containerd it is working.

@utam0k, could it be something in the bundle causing that?

rumpl:
Just spent some more time with this, is annoying that anyhow only shows the last error in the chain, the actual error I get is:

```
Apr 06 18:01:45 nuque containerd[528]: [ERROR] ERROR: failed to load spec
Apr 06 18:01:45 nuque containerd[528]: [ERROR] because: failed to validate runtime spec
Apr 06 18:01:45 nuque containerd[528]: [ERROR] because: runtime spec has incompatible version '1.1.0-rc.1'. Only 1.0.X is supported
```

0xE282B0:
@rumpl, You're right, bug hunting it's a bit cumbersome. Fortunately, the error you're mentioning was the last that blocks the container from running. Now I have at least a simple Wasm container running with moby.

@utam0k, I don't know what implications the version change has for youki but when I [disable the version check](
https://github.com/0xE282B0/youki/commit/b3fe971082569afd4801c824a6d4b6d04b52bebf) it is working for me. Shall we open an issue for youki to support `1.1.x` as well?

utam0k:
@rumpl @0xE282B0 :wave: Thanks for your check.
I will check it this weekend.

utam0k:
@0xE282B0 @rumpl I have put the commit to fix it!

0xE282B0:
LGTM, I'm able to run a simple Wasm container with Moby v24.0.0-beta.1 on Ubuntu 22.10 x64. 

The log is looking similar to the old containerd-shim-wasmedge version:


Well done! 🥳

I'll test it with different Kubernetes distributions on Tuesday, but when it is working with Docker there should not be a big problem.

Mossaka (approved):

cpuguy83 (approved):
LGTM

We can follow up with any necessary fixes.

utam0k:
@cpuguy83 Thanks for the hard PR review 🙇

cpuguy83:
Thanks for putting in all the effort! Great work.

PR #80, opened 2023-03-01
Make namespace setup conditional on OS

cpuguy83:
This code is only valid for linux, so only compile it there.


cpuguy83:
@jsturtevant

Mossaka (approved):

jsturtevant:
lgtm

Issue #49, opened 2023-01-09
Windows support


jsturtevant:
Currently we cannot use this project on Windows.  This is an umbrella issue to track Windows support. There may be more items added as the work progresses.

The project doesn't currently build:

```
 cargo build
   Compiling zstd-safe v5.0.2+zstd.1.5.2
   Compiling zstd-sys v2.0.4+zstd.1.5.2
   Compiling ittapi-sys v0.3.2
   Compiling wasmtime-runtime v2.0.2
   Compiling cranelift-wasm v0.89.2
   Compiling psm v0.1.21
   Compiling uapi v0.2.10
   Compiling uapi-proc v0.0.5
   Compiling num-traits v0.2.15
   Compiling ttrpc v0.6.1
   Compiling ittapi v0.3.2
   Compiling cap-fs-ext v0.26.1
error[E0425]: cannot find function `geteuid` in crate `libc`
   --> C:\Users\jstur\.cargo\registry\src\github.com-1ecc6299db9ec823\uapi-proc-0.0.5\src\lib.rs:16:34
    |
16  |             root: unsafe { libc::geteuid() == 0 },
    |                                  ^^^^^^^ help: a function with a similar name exists: `getpid`

Some errors have detailed explanations: E0412, E0422, E0425, E0432, E0433.
For more information about an error, try `rustc --explain E0412`.
error: could not compile `ttrpc` due to 58 previous errors
error: failed to run custom build command for `uapi v0.2.10
```

At a minimum the following tasks need to be completed:

- [ ] https://github.com/containerd/ttrpc-rust/issues/132
- [ ] https://github.com/containerd/rust-extensions/issues/4
- [ ] [Ensure can build test image on windows](https://github.com/containerd/runwasi/pull/41#issuecomment-1357997502)
- [ ] refactor usage of nix in runwasi
- [ ] refactor the cgroup setup to run only on Windows
- [ ] Add Windows e2e tests

squillace:
thanks, @jsturtevant !!!

jsturtevant:
This is the PR that enables can import the test image to windows (https://github.com/microsoft/hcsshim/pull/1688)

jsturtevant:
When parsing the runtime spec in runwasi it fails, tracking in https://github.com/containers/oci-spec-rs/issues/126

jprendes:
[...]

PR #81, opened 2023-03-02
Update project to latest shim-protos and ttrpc using protobuf 3.x

jsturtevant:
In preparation for adding windows #49 which is built using latest [ttrpc libraries](https://github.com/containerd/ttrpc-rust/pull/172) which use protobuf 3, this PR updates the project to use latest dependencies and moves the protocols to using protobuf 3.  

note:  There isn't a release of the [shim-protos](https://github.com/containerd/rust-extensions/blob/main/crates/shim-protos/Cargo.toml) project with these changes yet so it pins a rev.  I am going to open an issue on https://github.com/containerd/rust-extensions/ to see if we can get a release (or maybe wait till the windows changes propagate up). Opened this up to show required changes which allow for using Windows.


jsturtevant:
This is taking much longer than anticipated, things are moving along but slowly.  I am finding it difficult to maintain this patch and work on the Windows implementation with the various changes going in for the youki work. 

Any thoughts on pinning to a rev vs a realse so we can start bumping the other dependencies and moving forward on protobuf 3?

devigned:
> This is taking much longer than anticipated, things are moving along but slowly. I am finding it difficult to maintain this patch and work on the Windows implementation with the various changes going in for the youki work.
> 
> Any thoughts on pinning to a rev vs a realse so we can start bumping the other dependencies and moving forward on protobuf 3?

I'm good with pinning to a rev as long as we have an issue to track the follow up work. What do others think?

jsturtevant:
> I'm good with pinning to a rev as long as we have an issue to track the follow up work. What do others think?

I've refreshed the PR and opened https://github.com/containerd/runwasi/issues/162 to track updating getting this on a release

cpuguy83 (approved):

PR #83, opened 2023-03-06
Add cargo publish step to release pipeline

### LINKED ISSUE #13: Feature: Update GH Actions to Release the Crates to crate.io

cpuguy83:
Example successful run: https://github.com/cpuguy83/runwasi-fork/actions/runs/4347852402'
crates.io package: https://crates.io/crates/containerd-shim-wasm/versions

Note I yanked the version because it is technically incorrect since I did not set the version in Cargo.toml.
And here is a run where I added validation to ensure the Cargo.toml version matches the tag: https://github.com/cpuguy83/runwasi-fork/actions/runs/4348574593

To support this without having to create an API token from my personal account, I setup a bot account @containerd-runwasi-release-bot which has 2FA enabled.

Additionally once the crates.io package was created I manually set the owners of the package:

image

This will need to be done for the first release on crates.io for each new crate.

Also just to note, containerd-shim-wasm must first be released then the other crates need to be updated to include an explicit version of that crate before we can release those.

Closes #13

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,10 +20,35 @@ jobs:
       crate: ${{ steps.parse-ref.outputs.crate }}
       version: ${{ steps.parse-ref.outputs.version }}
     steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.ref }}
       - id: parse-ref
+        name: Parse ref
         run: |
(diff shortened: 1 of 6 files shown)


Mossaka (approved):
lgtm!

rumpl (approved):

PR #85, opened 2023-03-07
Fix code docs

cpuguy83:
1. Docs were using godoc style docs rather than rustdoc, so were not being picked up by rustdoc
2. Added some missing docs
3. Gave containerd-shim-wasm crate its own README.md so it makes more sense on crates.io

Example rustdoc:

image


rumpl (approved):

devigned (approved):
lgtm!

PR #86, opened 2023-03-07
release: Do not create empty archives

cpuguy83:
When there are no binary artifacts we should not be generating tarballs since they will just be empty.


Mossaka (approved):

Dockerfile
```diff
@@ -60,7 +60,11 @@ FROM build AS build-tar
 WORKDIR /build/release
 ARG CRATE
 ARG TARGETOS TARGETARCH TARGETVARIANT
-RUN tar -C /build/bin -czf /build/release/${CRATE}-${TARGETOS}-${TARGETARCH}${TARGETVARIANT}.tar.gz .
+RUN <<EOF
+if [ -n "$(find /build/bin -type f -exec echo {} \;)" ]; then
+    tar -C /build/bin -czf /build/release/${CRATE}-${TARGETOS}-${TARGETARCH}${TARGETVARIANT}.tar.gz .
```
ipuustin, line 65: Maybe put double quotes also around the file name in case any variable ever has spaces etc:
 
    tar -C /build/bin "/build/release/${CRATE}-${TARGETOS}-${TARGETARCH}${TARGETVARIANT}.tar.gz" .

ipuustin (approved):
LGTM!

Mossaka (approved):
lgtm!

PR #87, opened 2023-03-08
Bump wasmtime to 6.0

cpuguy83:
Because there is a shared dep (wasmtime-fiber) between the wasmtime shim and the wasmedge shim, and that dep links to native code, these normally have to be updated in sync, but wasmtime-fiber is a transitive dependency in wasmedge and requires updating upstream before we can do anything here, which is not ideal, especially going forward. We could split the shims out of the workspace... but then that's almost as painful to deal with given all the other shared dependencies.

For now this changes the wasmtime shim to disable the `async` feature which makes means wasmtime-fiber is no longer a dependency in the wasmtime shim.
We also don't use any async functions so this shouldn't really be changing anything.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,17 +70,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


Mossaka (approved):
lgtm!

jsturtevant (approved):
lgtm 

I still think it maybe valuable down the line some point to have independent dependencies for the individual shims but doesn't seem to be necessary at this point.

Issue #72, opened 2023-02-22
Update wasmtime deps to latest version (6.x)


jsturtevant:
I tried to update the `wasmtime` dep to the latest and found that `runwasmedge` is blocking the upgrade do to a dep on `wasmtime-fiber`:

```
[ERROR rust_analyzer::lsp_utils] rust-analyzer failed to load workspace: Failed to read Cargo metadata from Cargo.toml file /home/jstur/projects/runwasi/Cargo.toml, Some(Version { major: 1, minor: 67, patch: 1 }): Failed to run `"cargo" "metadata" "--format-version" "1" "--features" "generate_bindings" "--manifest-path" "/home/jstur/projects/runwasi/Cargo.toml" "--filter-platform" "x86_64-unknown-linux-gnu"`: `cargo metadata` exited with an error:     Blocking waiting for file lock on package cache
    Updating crates.io index
error: failed to select a version for `wasmtime-fiber`.
    ... required by package `runwasmedge v0.1.0 (/home/jstur/projects/runwasi/crates/wasmedge)`
versions that meet the requirements `^2.0` are: 2.0.2, 2.0.1, 2.0.0

the package `wasmtime-fiber` links to the native library `wasmtime-fiber-shims`, but it conflicts with a previous package which links to `wasmtime-fiber-shims` as well:
package `wasmtime-fiber v6.0.0`
    ... which satisfies dependency `wasmtime-fiber = "=6.0.0"` of package `wasmtime v6.0.0`
    ... which satisfies dependency `wasmtime = "^6.0"` of package `runwasmtime v0.1.0 (/home/jstur/projects/runwasi/crates/wasmtime)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the links ='wasmtime-fiber' value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

failed to select a version for `wasmtime-fiber` which could resolve this conflict


[...]

PR #89, opened 2023-03-15
Split shims from workspace

ipuustin:
Move the shims binaries to separate dependency management. This it not a fully good thing, because it increases the manual effort needed for dependency management. The positive side is that it removes conflicts between wasmedge and wasmtime dependencies. Note that `cargo build` no longer builds the shims -- you need to use `make build` for that.

Fixes: https://github.com/containerd/runwasi/issues/72


Mossaka:
Isn't the only conflicted dependency, shared by both wasmedge and wasmtime, `wasmtime-fiber` has been pinned by PR #87? Thus we don't really need to split them into two shims from the workspace anymore. I guess we still might ran into the dependency conflicts problem in the future.

ipuustin:
Yeah, as long as we don't use `async` feature in wasmtime, things will work for now. However I think we'll eventually hit more snags like this, especially if we add new runtimes with new deps to the repository. I think this boils down to how we want to organize the repo -- I see three options:

1. Separate crates for shims (this PR). Downside is more manual work updating the deps. Also if the runwasi lib deps start to conflict with shim runtime deps, then this doesn't help with that.
2. Solving the dependency problems one-by-one when they appear (current model). This might become easier or more difficult when we move to libcontainer. This may also require that we work with the runtime lib projects to keep them in sync? 
3. Separate git repos and thus different workspaces for the different shims to keep the dependencies separate. The benefit would be that in the future the shims could have their release cycles detached from runwasi release cycle. The downside would be a lot more tricky API management and the added complexity in code and repository management.

cpuguy83:
> The benefit would be that in the future the shims could have their release cycles detached from runwasi release cycle

This one we already have separate since we are namespacing tags per crate as described in https://github.com/containerd/runwasi/blob/main/RELEASE.md

That said, I don't think it would be horrible to split up the workspace if we end up having more problems.

I think for the wasmtime-fiber issue we should probably talk to the wasmtime folks about this:

https://github.com/bytecodealliance/wasmtime/blob/13be5618a7fce47ae1069be011979538fe7809db/crates/fiber/Cargo.toml#L10-L14

```
# We link to some native code with symbols that don't change often, so let Cargo
# know that we can't show up multiple times in a crate graph. If this is an
# issue in the future we should tweak the build script to set `#define`
# directives or similar to embed a version number of this crate in symbols.
```

cpuguy83:
Opened https://github.com/bytecodealliance/wasmtime/issues/6083 to discuss this.

ipuustin:
Closing this (at least for now).

PR #92, opened 2023-03-21
testutil: normalize test names.

ipuustin:
Noticed that when we use `run_test_with_sudo` from wasmedge shim (such as in https://github.com/containerd/runwasi/pull/78), the test name passed to the function contains `::{{closure}}` at the end, meaning that the tests aren't triggered by the test runner. The name normalization fixes it. I'm thinking we might at some point add more checks and validation, so I made a separate function.


Mossaka (approved):
lgtm!

cpuguy83:
Thanks!
We are going to hold off merges until the youki wasmedge PR is merged.

cpuguy83 (approved):

PR #93, opened 2023-03-21
Minor cargo clippy fixes.

ipuustin:
Ran `cargo clippy`, fix a few very small issues reported by the tool. We could aim to have an empty clippy result (by adding comments to the places we don't want to fix) and add running clippy as a CI check?


rumpl (approved):

Mossaka:
> We could aim to have an empty clippy result (by adding comments to the places we don't want to fix) and add running clippy as a CI check?

Sounds good to me. We can run `cargo clippy --all-targets --all-features --workspace -- -D warnings` to fail the CI if there are any clippy warnings.

PR #95, opened 2023-03-21
CI: Add rustfmt and cargo clippy to CI

Mossaka:
This PR follows the discussion from https://github.com/containerd/runwasi/pull/93


.github/workflows/ci.yml
```diff
@@ -10,6 +10,24 @@ env:
...
 
 jobs:
+  fmt:
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: Swatinem/rust-cache@v2
+      - uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
jsturtevant, line 20: This will be a rolling version.  It means PR might pass a check when merging but main branch might fail later and need to be updated.

Mossaka: Can you explain why will the main branch might fail later if the CI passes?

jsturtevant: if this runs on 1.68 and passes, later 1.69 comes out with new clippy rules. Then locally update to 1.69, pull main, run clippy and it will fail.  

This can be annoying for contributors as they might have a PR open and rebased on the main branch but now this starts failing when the new tool chain comes out. Either they need to submit a sperate PR and fix then rebase or fix it in same PR to get it in.

If you do release branches this can be a similar issue, where before the fix goes into the branch the formats need to be applied for the new version.

Mossaka: Ah I see. What do you recommend to mitigate this issue? I think what we can do is that whenever a new release of clippy comes out (together with a new cargo release), we manually submit a PR to fix the main branch.

ipuustin:
I made a PR (https://github.com/Mossaka/runwasi/pull/2) for amending to this PR for marking all the existing clippy issues as "allowed". We can fix them later and we will get the CI to catch any new clippy issues.

.github/workflows/ci.yml
jsturtevant: IMO we should be pinning the version of Rust that we build against. This creates a little bit more work from a maintenance perspective but it allows more control, sets up for repeatable builds and avoids the pitfalls I mentioned above.

Mossaka:
> I made a PR ([Mossaka#2](https://github.com/Mossaka/runwasi/pull/2)) for amending to this PR for marking all the existing clippy issues as "allowed". We can fix them later and we will get the CI to catch any new clippy issues.

Thanks for your contribution! Merged it in

cpuguy83:
Needs a rebase.

.github/workflows/ci.yml
Mossaka: Done. I've pinned the rust version to 1.69.0

Mossaka:
Rebased, PTAL @cpuguy83

jsturtevant:
LGTM

.github/workflows/ci.yml
```diff
@@ -10,6 +10,25 @@ env:
...
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: Swatinem/rust-cache@v2
+      - uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: 1.69.0
+          components: rustfmt, clippy
+      - name: Setup WasmEdge build env
+        run: |
```
cpuguy83, line 23: Looks like we need all the system libs like in line 48 here.
Probably not build-essential but likely everything else.

cpuguy83 (changes requested):
Build is failing due to missing system libs

.github/workflows/ci.yml
Mossaka: Done.

cpuguy83:
Looks like a fmt error:

image

Mossaka:
Alright, should be passing now as cargo clippy passed on my local env.

cpuguy83 (approved):

PR #111, opened 2023-05-04
Enable WasmEdge roofs support and start writing OCI tests for Wasmedge and Wasmtime

ipuustin:
Now that we have support for OCI runtime spec for Wasmedge shim, we should try to do two things:

 1. find out which OCI spec features are actually reasonable to test for runwasi shims
 2. create a set of Wasm modules which exercise those OCI spec features so that we can test them

Those OCI spec features which don't really map to Wasm modules we can try to test on the runtime level, checking for example which capabilities the runtime process has?

This PR adds an external directory for storing/building the Wasm modules and an integration test directory for Wasmedge shim for trying out the OCI spec features. I added some seccomp and device file tests for seeing how this would work in practice. Support for Wasmtime is commented out because the OCI support (https://github.com/containerd/runwasi/pull/142) hasn't been merged yet, but could be useful in testing the PR.

So, I think the first question is: which OCI features are meaningful and should be tested for Wasm/WASI containers? For example it could be argued that the Wasm _modules_ should never see the device files, but the Wasm _runtimes_ should, so that they could expose the features through the WASI APIs. The second question is that is this approach of having the test Wasm modules in-tree the right way forward? The problem is that it appears to be difficult or impossible to run `cargo build` for two architectures at the same time (making the tests depend on the Wasm module build).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,10 @@ jobs:
           args: --all --verbose
       - name: Validate docs
         run: ./scripts/validate-docs.sh
+      - name: Setup rust-wasm target and compile test modules
+        run: |
+          rustup target add wasm32-wasi
+          make test/wasm-modules
       - name: Run tests
         run: |
           make test
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,9 @@ test/k8s: test/k8s/cluster
 test/k8s/clean: bin/kind
 	bin/kind delete cluster --name $(KIND_CLUSTER_NAME)
 
+test/wasm-modules:
+	cargo build -p wasi-modules-for-testing --target wasm32-wasi
+
 .PHONY: bin/k3s
 bin/k3s:
 	mkdir -p bin && \
--- a/test/oci-tests/tests/common/mod.rs
+++ b/test/oci-tests/tests/common/mod.rs
@@ -1,0 +1,100 @@
+use std::fs::{create_dir, read_to_string, File, OpenOptions};
+use std::io::prelude::*;
+use std::os::unix::fs::OpenOptionsExt;
+use std::path::{Path, PathBuf};
+use std::sync::mpsc::channel;
+use std::time::Duration;
+
+use libc::SIGKILL;
+use oci_spec::runtime::Spec;
+use serde::{Deserialize, Serialize};
+use tempfile::tempdir;
+
+use containerd_shim_wasm::sandbox::instance::Wait;
+use containerd_shim_wasm::sandbox::{EngineGetter, Error, Instance, InstanceConfig};
+
+#[derive(Serialize, Deserialize)]
+struct Options {
+    root: Option<PathBuf>,
+}
+
+pub static WASM_FILENAME: &str = "./file.wasm";
+
+pub(crate) fn get_external_wasm_module(name: String) -> Result<Vec<u8>, Error> {
+    let manifest_dir = env!("CARGO_MANIFEST_DIR");
+    let target = Path::new(manifest_dir)
+        .join("../../target/wasm32-wasi/debug")
+        .join(name.clone());
+    std::fs::read(target).map_err(|e| {
+            Error::Others(format!(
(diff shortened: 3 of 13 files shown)


Mossaka (commented):
Thanks for contributing more tests for OCI!

I quickly skimmed through the test cases and noticed that the larger part of them are not Wasmedge specific - which means the same set of tests can be port to test the wasmtime shim as well. 

After #142 is merged upstream, I am happy to take this task of making these tests available across shims.

Makefile
```diff
@@ -58,7 +58,7 @@ target/wasm32-wasi/$(TARGET)/img.tar: target/wasm32-wasi/$(TARGET)/wasi-demo-app
 	cd crates/wasi-demo-app && cargo build $(RELEASE_FLAG) --features oci-v1-tar
 
 load: target/wasm32-wasi/$(TARGET)/img.tar
-	sudo ctr -n $(CONTAINERD_NAMESPACE) image import --all-platforms $<
+	sudo ctr -n $(CONTAINERD_NAMESPACE) image import --all-platforms target/wasm32-wasi/debug/img.tar
```
cpuguy83, line 61: Why this change?
With the change it doesn't take into account `$TARGET` but `$<` does.

ipuustin: It was a leftover from a rebase... fixed now.

ipuustin:
> Thanks for contributing more tests for OCI!
> 
> I quickly skimmed through the test cases and noticed that the larger part of them are not Wasmedge specific - which means the same set of tests can be port to test the wasmtime shim as well.
> 
> After #142 is merged upstream, I am happy to take this task of making these tests available across shims.

Thanks! In fact the Wasmtime support is already present in the tests but commented out, because the OCI support is not there yet. If you like, you can try rebasing #142 on top of this PR, removing the comments (marked with `TODO:`) and running the tests.

test/oci-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    drop(stdout);
+
+    spec.save(dir.path().join("config.json"))?;
+
+    let mut cfg = InstanceConfig::new(WasmEdgeWasi::new_engine()?, "test_namespace".into());
+    let cfg = cfg
+        .set_bundle(dir.path().to_str().unwrap().to_string())
+        .set_stdout(dir.path().join("stdout").to_str().unwrap().to_string());
+
+    let wasi = Arc::new(WasmEdgeWasi::new("test".to_string(), Some(cfg)));
```
jsturtevant, line 133: I believe this returns the `trait Instance`?  If so we could invert this.  Create the `config` and the `instance` outside the module and make this function generic `run_test_with_spec` and pass in the module.  

This would make the tests written below to not be specific calls to each runtime `let (output, retval) = common::run_wasmedge_test(&spec, &bytes)?;` making it more of a framework.

.github/workflows/ci.yml
```diff
@@ -62,6 +62,10 @@ jobs:
           args: --all --verbose
       - name: Validate docs
         run: ./scripts/validate-docs.sh
+      - name: Setup rust-wasm target and compile test modules
+        run: |
+          rustup target add wasm32-wasi
```
jsturtevant, line 67: if you put this in the make target it makes, it will just skip if already present and makes it so some one runs that step they don't miss it

jsturtevant (commented):
This is a great, It feels like you're on the verge of creating a framework for testing different module specs.  I've made one suggestion to leverage our `Instance` trait and invert the logic a bit that would make it so we could easily re-use the tests across different runtimes.  Let me know if that makes sense

Mossaka:
Is this PR ready for review? @ipuustin

ipuustin:
> Is this PR ready for review? @ipuustin

It will be shortly -- I'll still add the Wasmtime OCI tests, because the support should be there now with the libcontainer integration.

jsturtevant:
#306 built upon this work and is now merged.  Thanks @ipuustin for the initial work!

PR #117, opened 2023-05-10
deps: do not use chrono default features.

ipuustin:
Disable `chrono` crate default features. This is because of a [potential vulnerability](https://rustsec.org/advisories/RUSTSEC-2020-0071) in `time` crate, which is a `chrono` dependency. Runwasi shims in this repository don't appear to be affected by the vulnerability since we don't set environmental variables.


Mossaka (approved):
lgtm!

PR #118, opened 2023-05-14
chore: bump wasmedge to 0.12.1 and wasmedge-sdk to 0.8.1

CaptainVincent:
1. bump new version
2. remove async feature relative to #72
    > According to the information I received from my colleague, there is no urgent need for a fork at the moment, so let me disable it here first.

### DIFF
--- a/crates/containerd-shim-wasmedge/src/executor.rs
+++ b/crates/containerd-shim-wasmedge/src/executor.rs
@@ -7,7 +7,7 @@ use std::os::unix::io::RawFd;
 
 use wasmedge_sdk::{
     config::{CommonConfigOptions, ConfigBuilder, HostRegistrationConfigOptions},
-    params, Vm,
+    params, VmBuilder,
 };
 
 static mut STDIN_FD: Option<RawFd> = None;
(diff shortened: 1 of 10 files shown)


crates/containerd-shim-wasmedge/src/executor.rs
```diff
@@ -42,17 +42,17 @@ impl Executor for WasmEdgeExecutor {
             .build()?;
 
         // create a vm with the config settings
-        let mut vm = Vm::new(Some(config))?;
+        let mut vm = VmBuilder::new().with_config(config).build()?;
 
         // initialize the wasi module with the parsed parameters
-        let mut wasi_instance = vm.wasi_module()?;
-        wasi_instance.initialize(
+        let wasi_module = vm.wasi_module_mut().ok_or("Not found wasi module").unwrap();
```
ipuustin, line 48: Maybe use `?` instead of `unwrap?`?

ipuustin (commented):
I think this looks good, only one small nit.

Removing the async feature means that it isn't now present in either Wasmedge or Wasmtime shim. In theory we could enable one of them using the current wasmtime-fiber approach, or if wasmtime-fiber is going to be forked/replaced by Wasmedge soon enough we can just wait.

crates/containerd-shim-wasmedge/src/executor.rs
CaptainVincent: Thanks your feedback.  Done.

cpuguy83:
Thanks!

PR #121, opened 2023-05-17
wasmedge: do not leak stdio file descriptors

ipuustin:
Since we are using the `std{in|out|err}` from containerd as raw file descriptors, we need to close them after we pass them to the container process.

### DIFF
--- a/crates/containerd-shim-wasmedge/src/instance.rs
+++ b/crates/containerd-shim-wasmedge/src/instance.rs
@@ -14,6 +14,7 @@ use libc::{dup2, SIGINT, SIGKILL, STDERR_FILENO, STDIN_FILENO, STDOUT_FILENO};
 use log::{debug, error};
 use nix::errno::Errno;
 use nix::sys::wait::{waitid, Id as WaitID, WaitPidFlag, WaitStatus};
+use nix::unistd::close;
 use serde::{Deserialize, Serialize};
 use wasmedge_sdk::{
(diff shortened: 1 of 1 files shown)


utam0k (approved):
👍

Mossaka (approved):
lgtm!

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -235,6 +238,12 @@ impl Instance for Wasi {
         let pid = container.pid().unwrap();
         container.start()?;
 
+        // Close the fds now that they have been passed to the container process
+        // so that we don't leak them.
+        stdin.map(|fd| unsafe { close(fd) });
```
Mossaka, line 243: Is there a way to close the fd without using `unsafe`?

cpuguy83: [nix](https://docs.rs/nix/latest/nix/index.html)::[unistd](https://docs.rs/nix/latest/nix/unistd/index.html)::[close](https://docs.rs/nix/latest/nix/unistd/fn.close.html#)

ipuustin: Ok fixed! I also moved the FD closing up before `start()`, because the process exists already after the `build()` step. Later we can investigate if we can pass the actual FD ownership to the runtime so that we wouldn't need to explicitly close the FDs.

cpuguy83 (approved):

Issue #97, opened 2023-03-28
Benchmarking 


Mossaka:
This issue is served as a place for discussing various ways / ideas we can benchmark `runwasi` the wasm shims that was proposed by @ipuustin 

- One idea is that we can write a simple wasm program (Fibonacci) and execute in `runwasi` and a native program executing in `runc`. This provides a base benchmark of comparising the performance of wasi program vs. native runc processes. It is not meant to benchmark the performance of WASI in general. 
- Having the base benchmark set, we can observe the performance difference for each version increments. For example, we can observe how much speed increase / descrease for version 0.2 vs. 0.3 
- Another idea of benchmarking is testing how "dense" of wasm pods can we go for a node. It is often advertised that wasm modules can increase CPU utilization and thus increasing the density of running pods per node. We can verify this point to push the containerd runtime to the extreme by running thousands of pods at the same time. 

Feel free to add ideas and thoughts on this topic! Any suggestion is welcome 🙏 

- [x] #612  
  - [x] #620
  - [x] #621
  - [x] #612
  - [x] #626
- [x] #613
- [x] #614
- [x] #615
  - [x] #734 
- [x] #616
- [x] Benchmark the end-to-end wasi-demo-app workloads #764 
- [x] #771
- [ ] reach out to the CNCF containerd maintainers and ask what managed services can we use for monitoring

Mossaka:
Close this one as #126 adds benchmarking support.

Mossaka:
Wanted to reopen this issue because I think #126 does not fully address the scope of the issue mentioned above. A few things that will make runwasi benchmarking stories better

- add #126 to the CI
- run runc and crun in the CI to give comparison results
- dense test

Mossaka:
Check out SpinKube perf test: https://github.com/fermyon/spinkube-performance

PR #126, opened 2023-05-30
benchmarks: add initial benchmarking support.

ipuustin:
Add a submodule "webassembly-benchmarks". The benchmark contains a set of wasm files from libsodium. Only the benchmarks which are fast to run (take under 5 seconds to complete on a desktop computer using WasmEdge) were selected for benchmarking.

The benchmark suite initializes the Wasm runtime, runs the benchmark, and cleans up the container. This is done for both WasmEdge and Wasmtime shim crates. Note that the results for the two runtimes shouldn't really be directly compared right now, because Wasmtime shim hasn't been converted to youki/libcontainer-rs yet and that changes the performance due to the extended namespacing. What this helps with is to see what is the performance impact of any given change, for example having the Youki integration done, using a seccomp filter, or setting cgroup options.

The benchmarking library used is Criterion. Run the benchmarks like this:

    # cargo bench

If you have gnuplot binary installed, Criterion will generate an HTML report with graphs in target/criterion/report/index.html file. Running the suite takes bit over ten minutes on my desktop.

This PR adds two new direct dependencies: Criterion (Apache2.0/MIT) as dev-dependency and webassembly-benchmarks (ISC) as a submodule.

Note that this PR is just the first step in the benchmarking path -- it does only pretty low-level benchmarking using a limited set of tests. Later on we can add comparison to the normal containers (by compiling and running the libsodium benchmarks as regular processes with the native Youki executor), add or remove Wasm benchmark modules, and improve the test setup. We could for example try running hunderds of containers at once to see how "dense" containers on a node would perform.

This is related to https://github.com/containerd/runwasi/issues/97 (and should fix the second point in the list).

### DIFF
--- a/benches/containerd-shim-benchmarks/benches/webassembly-benchmarks.rs
+++ b/benches/containerd-shim-benchmarks/benches/webassembly-benchmarks.rs
@@ -1,0 +1,362 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+
+use std::borrow::Cow;
+use std::fs::{create_dir, File, OpenOptions};
+use std::io::Write;
+use std::os::unix::prelude::OpenOptionsExt;
+use std::path::{Path, PathBuf};
+use std::sync::mpsc::channel;
+use std::sync::Arc;
+use std::time::Duration;
+
+use anyhow::bail;
+use chrono::{DateTime, Utc};
+
+use containerd_shim_wasm::sandbox::exec::has_cap_sys_admin;
+use containerd_shim_wasm::sandbox::instance::Wait;
+use containerd_shim_wasm::sandbox::{EngineGetter, Error, Instance, InstanceConfig};
+
+use containerd_shim_wasmedge::instance::{reset_stdio, Wasi as WasmEdgeWasi};
+use containerd_shim_wasmtime::instance::Wasi as WasmtimeWasi;
+use libc::SIGKILL;
+use oci_spec::runtime::{ProcessBuilder, RootBuilder, Spec, SpecBuilder};
+use serde::{Deserialize, Serialize};
+use tempfile::{tempdir, TempDir};
+
+/*
+    For benchmarking try to choose cases which run fast enough -- the idea is to
+    get a rough idea of the performance rather than run a hours-long benchmark
+    suite.
+
+    Because of this, only select the benchmarks which run in under 5 seconds on
+    a desktop computer using WasmEdge. Note that this selection is pretty
+    arbitrary and we can add or remove benchmarks easily, also from other
+    sources. This is the selection algorithm:
+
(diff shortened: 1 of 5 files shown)


ipuustin:
The sources for the benchmark modules are here: https://github.com/jedisct1/libsodium/tree/master/test/default

benches/containerd-shim-benchmarks/benches/webassembly-benchmarks.rs
```diff
@@ -0,0 +1,362 @@
...
+    let wasi = Arc::new(WasmtimeWasi::new("test".to_string(), Some(cfg)));
+
+    wasi.start()?;
+
+    let w = wasi.clone();
+    let (tx, rx) = channel();
+    let waiter = Wait::new(tx);
+    w.wait(&waiter).unwrap();
+
+    let res = match rx.recv_timeout(Duration::from_secs(600)) {
```
Mossaka, line 125: 10 minutes of timeout seems too long for me...

Mossaka (approved):
LGTM! Some questions regarding the benchmark: 

- What areas do this set of wasm modules test? CPU, Memory etc. 
-  Does this set of wasm modules require WASI at all?

benches/containerd-shim-benchmarks/benches/webassembly-benchmarks.rs
ipuustin: That's right. I reduced the timeout to one minute. That should be plenty too...

ipuustin:
> LGTM! Some questions regarding the benchmark:
> 
> * What areas do this set of wasm modules test? CPU, Memory etc.

Depends on the test. For example, the `hash` tests should be all about memory access. The encryption tests are probably more computation-intensive. What we are missing is an "empty" Wasm module that would just return. With that we could measure the setup/teardown time. The benefit of using the libsodium tests is that we are getting something that looks more like a real-life workload compared to just computing fibonacci sequence.

> * Does this set of wasm modules require WASI at all?

No, these tests don't use WASI.

Mossaka (approved):
LGTM! 

I can merge this in when all CI pipelines are passed.

cpuguy83:
Some linting issues, also conflict in Cargo.lock.
```
error: redundant clone
   --> crates/containerd-shim-wasmtime/src/instance.rs:355:21
    |
355 |         let w = wasi.clone();
    |                     ^^^^^^^^ help: remove this
    |
note: cloned value is neither consumed nor mutated
   --> crates/containerd-shim-wasmtime/src/instance.rs:355:17
    |
355 |         let w = wasi.clone();
    |                 ^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
    = note: `-D clippy::redundant-clone` implied by `-D warnings`

error: could not compile `containerd-shim-wasmtime` (lib test) due to previous error
warning: build failed, waiting for other jobs to finish...
```

PR #127, opened 2023-05-30
Bump wasmtime to 9.0.2

Mossaka:
This supercedes #122 

This PR also adds "wasmtime", "wasi-common" and "wasmtime-wasi" deps to be ignored by the dependabot because it is usually required to update all together, but dependabot has tried to update individual dep which failed the pipelines.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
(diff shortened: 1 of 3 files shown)


ipuustin (approved):
LGTM. I'm ok with excluing the 'wasmtime-*' deps, but in a perfect world they would have their internal dependencies set correctly so we wouldn't need to do any special handling for them.

Mossaka:
They are seperate published crates, but updating one needs update to another one.

PR #130, opened 2023-06-07
Update `youki` to include capabilities fix

jprendes:
Update youki dependency to include [a fix](https://github.com/containers/youki/commit/2ff8b97e676ebf848842df8fbc01f1b5c5c7b08c) for a capabilities issue that prevented it from running inside of Docker Desktop


crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -28,6 +28,7 @@ clone3 = "0.2"
 libc = { workspace = true }
 caps = "0.5"
 proc-mounts = "0.3"
+libcontainer = { workspace = true }
```
ipuustin, line 31: I think it might be bit of an issue that the `containerd-shim-wasm` crate needs to depend on `libcontainer` for error handling. Doesn't that imply that if an (external) shim uses some other error type, runwasi would also need to depend on the library that provides it to be consistent? I'm not really sure what would be the right solution here, and this PR isn't the problem, I'm just wondering if this could be done better.

ipuustin (approved):
This looks good to me. I think we can figure out the error handling question separately if there is a need to.

crates/containerd-shim-wasm/Cargo.toml
jprendes: I would be interested in moving `youki` out of the WasmEdge implementation into a lower level of `runwasi` so that all shims can benefit from it. In that case having a dependency on `libcontainer` would be less of a concern.
I will open an Issue to discuss it further.

jprendes: Ok, it's a bit more complex thank I thought.

Adding this variant to the `Error` enum means that all the shims (not only those using `youki`)  inherit youki's runtime dependencies. This is basically `libdbus` (and transitively, `libsystemd`).

If we plan to move all shims to `youki`, then this is a reasonable cost to pay.
Otherwise I could place `youki`'s errors in `Error::Any`. It's less descriptive, but would do the trick without adding the extra dependencies.

Cargo.toml
```diff
@@ -28,11 +28,11 @@ chrono = { version = "0.4", default-features = false, features = ["clock"] }
 nix = "0.26"
 cap-std = "1.0"
 thiserror = "1.0"
-libc = "0.2.145"
+libc = "0.2.146"
 oci-spec = { version = "0.6.0", features = ["runtime"] }
 sha256 = "1.1"
 # TODO: Wait for the release and use the crates.io one.
-libcontainer = { git = "https://github.com/containers/youki", rev = "edd63c84f903aa09510ef758ea6f617e0cb8b7e1" }
+libcontainer = { git = "https://github.com/containers/youki", rev = "3a790dffe9d58f186ead31a1f5e71587ce67124c" }
```
Mossaka, line 35: Personally, I would like to have a comment there saying what branch is this commit in. It can be as simple as "# main".

crates/containerd-shim-wasm/Cargo.toml
Mossaka: Just for this PR, is it possible to make containerd-shim-wasm not depend on youki?

Mossaka (commented):
Thanks a lot! Added a few comments.

crates/containerd-shim-wasm/Cargo.toml
jprendes: Yes, I think that's sensible. I'll use `Error::Any` for now.

jprendes:
I will rebase this PR after #132 is merged

cpuguy83:
Merge conflict in Cargo.lock

jprendes:
Rebased to fix merge conflict.

ipuustin:
This is important also because the oci-spec-rs 0.6.1 appears not to be compatible with the previous libcontainer SHA.

jprendes:
The version of `youki` in the last push includes the [commit from 2 days ago](https://github.com/containers/youki/commit/545b13da8636bcc8911a0bebbad4dac547cc9698) that updates `oci-spec-rs` to `0.6.1`, but I don't see any SHA related changes in that commit.

ipuustin:
The youki commit where the oci-spec version is bumped is required for oci-spec 0.6.1 support for runwasi too... otherwise runwasi compilation will fail for missing Time namespace support:

```
   Compiling libcontainer v0.0.5 (https://github.com/containers/youki?rev=edd63c84f903aa09510ef758ea6f617e0cb8b7e1#edd63c84)
error[E0004]: non-exhaustive patterns: `LinuxNamespaceType::Time` not covered
   --> /home/ipuustin/.cargo/git/checkouts/youki-135a19ffee0653b9/edd63c8/crates/libcontainer/src/namespaces.rs:33:11
    |
33  |     match namespace_type {
    |           ^^^^^^^^^^^^^^ pattern `LinuxNamespaceType::Time` not covered
    |
note: `LinuxNamespaceType` defined here
   --> /home/ipuustin/.cargo/registry/src/github.com-1ecc6299db9ec823/oci-spec-0.6.1/src/runtime/linux.rs:792:5
    |
769 | pub enum LinuxNamespaceType {
    | ---------------------------
...
792 |     Time = 0x00000080,
    |     ^^^^ not covered
    = note: the matched value is of type `LinuxNamespaceType`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
    |
40  ~         LinuxNamespaceType::Mount => CloneFlags::CLONE_NEWNS,
41  ~         LinuxNamespaceType::Time => todo!(),
    |
```

ipuustin (approved):
Looks good now!

PR #132, opened 2023-06-08
Fixes to tests to make cargo clippy succeed

ipuustin:
Add `delete()` to Wasmtime tests. Also clean up the code a little to make `cargo clippy` succeed.

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/instance.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/instance.rs
@@ -179,7 +179,6 @@ impl Instance for Nop {
     fn start(&self) -> Result<u32, Error> {
         Ok(std::process::id())
     }
-
     fn kill(&self, signal: u32) -> Result<(), Error> {
         let code = match signal as i32 {
             SIGKILL => 137,
@@ -208,7 +207,6 @@ impl Instance for Nop {
 #[cfg(test)]
 mod noptests {
     use std::sync::mpsc::channel;
-    use std::sync::Arc;
     use std::time::Duration;
 
     use libc::SIGHUP;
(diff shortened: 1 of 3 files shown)


crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -219,12 +219,9 @@ mod noptests {
     fn test_nop_kill_sigkill() -> Result<(), Error> {
         let nop = Arc::new(Nop::new("".to_string(), None));
```
jprendes, line 220: IIUC, you don't need this `Arc` either.
Same for all the other ones in the PR (inc. the `wasi` one).

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -372,6 +371,8 @@ mod wasitest {
         let output = read_to_string(dir.path().join("stdout"))?;
         assert_eq!(output, "hello world\n");
 
+        wasi.delete()?;
```
jprendes, line 374: Would it make sense to have as part of `trait Drop`?

jprendes (commented):
Thanks for working on this!

Mossaka (approved):
LGTM! Happy to merge this once comments are resolved.

crates/containerd-shim-wasm/src/sandbox/instance.rs
ipuustin: Hmm, I think you are right. Removed the Arc wrapping from tests now.

crates/containerd-shim-wasmtime/src/instance.rs
ipuustin: I'm not sure! It would look nicer but we would lose the error handling if `delete()` fails. Also it would be a breaking change in the API. I would maybe not add that change into this PR but think about it later.

jprendes (approved):
LGTM!

PR #133, opened 2023-06-09
Update the docs to include install deps

jsturtevant:
Updates the docs to be clear about dependecies and streamlines a few files so there is concistency with make file.


Mossaka (approved):
LGTM! Thanks, @jsturtevant

Issue #108, opened 2023-04-28
Distribute using OCI artifacts / allow shim to pull artifact


radu-matei:
This is a feature request to allow distribution of apps targeting `runwasi` using OCI artifacts, and _not_ as a container images.

Currently, all applications targeting `runwasi` need to be distributed as a container image, with the structure carefully constructed. This is not an ideal long-term solution for a number of reasons, deduplication of Wasm files and static assets being one of them, particularly in the context of https://github.com/bytecodealliance/registry/pull/87 and https://hackmd.io/50rfwV6BTJWN8VZBhdAN_g.

Ideally, pulling the artifact could be done by the actual shim implementation — and implementations could continue to default to wrapping the Wasm app in a container image; however, some shim implementations would benefit greatly from using their existing mechanism of distributing apps using OCI artifacts (see https://developer.fermyon.com/spin/distributing-apps.md#publishing-a-spin-application-to-a-registry).

Thoughts?

cc @rumpl, @squillace, @Mossaka, @devigned.

jsturtevant:
This has a bit of cross-over of the conversation on https://github.com/deislabs/containerd-wasm-shims/issues/89.  In particular, we were looking for ways to enable containerd to do the [pulling ](https://github.com/deislabs/containerd-wasm-shims/issues/89#issuecomment-1523804934) or [unpacking](https://github.com/deislabs/containerd-wasm-shims/issues/89#issuecomment-1526058470).

There was also @devigned comment that aligns with your idea that we should be using [OCI artifacts and not container images.](https://github.com/containerd/runwasi/issues/106#issuecomment-1523433276)

squillace:
I absolutely believe this is the proper path. I do wish to support scratch images if possible; but it introduces arch and os issues that have to be navigated and that's precisely "handling the current" and not moving to the proper solution. Artifacts are that solution in my opinion.

cpuguy83:
To note, there's been some discussion for having containerd ask the shim to pull in order to better support confidential compute.
I think wasm tends to fall in the same category there along with VM's.
It wouldn't be awful to let the shim handle pull so we don't have to teach containerd  how to deal with stuff... but then every shim needs to implement that.

squillace:
Is there a way to surface a shared ability? Some crate?

radu-matei:
[...]

PR #135, opened 2023-06-09
Add support for building Containerd OCI Artifacts

jsturtevant:
Address #108 

This adds an experimental tool for creating a WASM OCI Artifacts based on the proposal in https://docs.google.com/document/d/11shgC3l6gplBjWF1VJCWvN_9do51otscAm0hBDGSSAc.  Containerd and runwasi don't currently support the OCI artifact format that is generated.

### DIFF
--- a/crates/oci-tar-builder/README.md
+++ b/crates/oci-tar-builder/README.md
@@ -1,19 +1,56 @@
 ## OCI Tar Builder
 
 This is a library that can be used to build OCI tar archives. It is used by the `wasi-demo-app` crate to build the OCI tar archive that is used to run the demo app.
-The currently implementation is to support encapsulating the `wasi-demo-app` wasm module into an OCI tar. It may be possible to use this for other purposes, but that not currently a goal of the project.
+The currently implementation is to support encapsulating the `wasi-demo-app` wasm module as an OCI tar.
 
 ### Contributing
 
 We welcome contributions to this to make it more robust, useful, and generally better.
 
-### Usage
+### Library Usage
 
 The library is currently not published to crates.io, so you will need to add the following to your `Cargo.toml`:
 
 ```toml
 [dependencies]
 oci-tar-builder = { git = "https://github.com/containerd/runwasi.git" }
+```
+
+See [wasi-demo-app build script](../wasi-demo-app/build.rs) for an example.
+
+### Executable usage
+
+There is an experimental executable that uses the library and can package a wasm module as an OCI artifact.  This isn't currently usable with containerd/runwasi today.  See the [OCI artifacts in containerd](https://docs.google.com/document/d/11shgC3l6gplBjWF1VJCWvN_9do51otscAm0hBDGSSAc) for more information.
+
+To generate the package and import to a registry using a tool such as [regctl](https://github.com/regclient/regclient/blob/main/docs/regctl.md#image-commands): 
+
+```
+cargo run --bin oci-tar-builder -- --name wasi-demo-app --module ./target/wasm32-wasi/debug/wasi-demo-app.wasm -o ./bin
+regctl image import localhost:5000/wasi-demo-oci:module ./bin/wasi-demo-app.tar        
+```
+
+View the manifest created, notice that the media types are `application/vnd.w3c.wasm.module.v1+wasm` which are subject to change.
+
+```
+regctl manifest get localhost:5000/wasi-demo-oci:module
+Name:                                localhost:5000/wasi-demo-oci:module
+MediaType:                           application/vnd.oci.image.manifest.v1+json
+Digest:                              sha256:869fb6029e26713160d7626dce140f1275f591a694203509cb1e047e746daac8
+Annotations:
+  io.containerd.image.name:          localhost:5000/wasi-demo-app
+  org.opencontainers.image.ref.name: 5000/wasi-demo-app
+Total Size:                          2.565MB
+
+Config:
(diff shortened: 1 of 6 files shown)


devigned (approved):
lgtm!

cpuguy83 (approved):

crates/oci-tar-builder/src/bin.rs
```diff
@@ -0,0 +1,101 @@
...
+                "wasm" => {
+                    builder.add_layer_with_media_type(
+                        &path,
+                        "application/vnd.wasm.component.v1+wasm".to_string(),
+                    );
+                }
+                "yml" => {
+                    builder.add_layer_with_media_type(
+                        &path,
+                        "application/vnd.wasm.component.config.v1+json".to_string(),
```
ipuustin, line 44: Mapping `yml` files to `json` media type? It would work in the other direction because yaml is a superset of json.

crates/oci-tar-builder/src/bin.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    let full_image_name = args.repo.unwrap_or("localhost:5000".to_string()) + "/" + &args.name;
+    builder.add_config(img, full_image_name);
+
+    let p = out_dir.join(args.name + ".tar");
+    let f = File::create(p).unwrap();
+    builder.build(f).unwrap();
+}
+
+/// Simple program to greet a person
```
ipuustin, line 83: Hmm, is this comment right?

crates/oci-tar-builder/src/lib.rs
```diff
@@ -68,13 +73,17 @@ impl Builder {
...
         for layer in self.layers.iter() {
-            let dgst = try_digest(layer.as_path()).context("failed to digest layer")?;
-            let meta = metadata(layer).context("could not get layer metadata")?;
+            let dgst = try_digest(layer.0.as_path()).context("failed to digest layer")?;
+            let meta = metadata(layer.0.clone()).context("could not get layer metadata")?;
             let oci_digest = "sha256:".to_owned() + &dgst;
 
+            let mut media_type = MediaType::ImageLayer;
+            if !layer.1.is_empty() {
+                media_type = MediaType::Other(layer.1.clone());
```
ipuustin, line 82: Should the `layer.1` media types be enums instead of strings? Or is the idea that we are just a postman for the types?

crates/oci-tar-builder/src/bin.rs
```diff
@@ -0,0 +1,101 @@
...
+            spec::RootFsBuilder::default()
+                .diff_ids(vec![])
+                .build()
+                .unwrap(),
+        )
+        .build()
+        .context("failed to build image configuration")
+        .unwrap();
+
+    let full_image_name = args.repo.unwrap_or("localhost:5000".to_string()) + "/" + &args.name;
```
ipuustin, line 75: If you use a hardcoded default URL, you need to document it somewhere.

crates/oci-tar-builder/src/bin.rs
```diff
@@ -0,0 +1,101 @@
...
+    builder.add_config(img, full_image_name);
+
+    let p = out_dir.join(args.name + ".tar");
+    let f = File::create(p).unwrap();
+    builder.build(f).unwrap();
```
ipuustin, line 80: Clean up the file instead of just panicking?

crates/oci-tar-builder/src/bin.rs
```diff
@@ -0,0 +1,101 @@
...
+                    "Skipping Unknown file type: {:?} with extension {:?}",
+                    path,
+                    path.extension().unwrap()
+                ),
+            }
+        }
+    }
+
+    let config = spec::ConfigBuilder::default()
+        .entrypoint(vec!["".to_owned()])
```
ipuustin, line 57: Why is the entrypoint set? This can be perfectly right, I'm just curious. :-)

crates/oci-tar-builder/README.md
```diff
@@ -1,13 +1,13 @@
 ## OCI Tar Builder
 
 This is a library that can be used to build OCI tar archives. It is used by the `wasi-demo-app` crate to build the OCI tar archive that is used to run the demo app.
-The currently implementation is to support encapsulating the `wasi-demo-app` wasm module into an OCI tar. It may be possible to use this for other purposes, but that not currently a goal of the project.
+The currently implementation is to support encapsulating the `wasi-demo-app` wasm module as an OCI tar.
```
ipuustin, line 4: Typo: "currently" -> "current"

ipuustin (commented):
I'm late to this review because this is already merged, but consider these as improvement ideas for the next patch set...

crates/oci-tar-builder/src/lib.rs
jsturtevant: since the various types for wasm aren't set in stone and the library might be used for other reasons, I was just passing those values through. Maybe in the future we might want to have stricter checks here?

crates/oci-tar-builder/src/bin.rs
jsturtevant: woops!  cleaning up. this was left over from the example in  https://docs.rs/clap/latest/clap/#example

crates/oci-tar-builder/src/bin.rs
jsturtevant: I don't think it is nessary, I'll drop it for now

PR #140, opened 2023-06-12
fix: remove an extra space in default container root dir

Mossaka:
This PR removes an extra space that causes the default container root dir to not be a valid / absolute path.


jsturtevant:
LGTM

ipuustin (approved):
LGTM! I'm not sure if having a default container root directory makes sense as a concept though?

PR #54, opened 2023-01-25
Update README.md API Instance definition.

ipuustin:
Update Instance trait to match the code reality.

PR #142, opened 2023-06-13
Use youki's libcontainer APIs to implement wasmtime shim.

Mossaka:
- Implement Executor trait for wasmtime executor.
- Use libcontainer's lifecycle APIs to implement runwasi's Instance
  trait.


utam0k:
@Mossaka Do you want me to review this PR?

Mossaka:
> @Mossaka Do you want me to review this PR?

I would appreciate your feedback once this PR is ready to review :)

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -0,0 +1,133 @@
...
+    ) -> anyhow::Result<(Store<wasi_common::WasiCtx>, wasmtime::Func)> {
+        // already in the cgroup
+        let env = oci_wasmtime::env_to_wasi(spec);
+        log::info!("setting up wasi");
+
+        let path = wasi_dir(".", OpenOptions::new().read(true))?;
+        let mut wasi_builder = WasiCtxBuilder::new()
+            .args(args)?
+            .envs(env.as_slice())?
+            .preopened_dir(path, "/")?;
```
Mossaka, line 68: This path contains a wasm binary. This might not be desired in the future. 

Major concenrs: james work, preload binary, and include static assets not included in the binary. 

rootfs
  - root
  - .wasm

ipuustin: Note that here https://github.com/ipuustin/runwasi/commit/7e72c3ca10a0454d4e220aa35db26f710fb03a17#diff-36f92d3bdc22f6005e8d13ff459d9d135364b34748de944da89785ecaa8d9e0aR58 is how we could enable rootfs file access for WasmEdge shim too. We need to have a discussion what we want to do about it -- should we only expose /dev and /proc files to the runtime, or to the container too?

Mossaka:
This PR is ready to review for now. The `test_wasi` test is still failing with the following message:

```bash
Error: Any(failed to receive. "waiting for init ready". BrokenChannel

Caused by:
    channel connection broken)
```

crates/containerd-shim-wasmtime/src/executor.rs
Mossaka: I created an issue for this. Closing it now.

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -0,0 +1,133 @@
...
+        spec: &Spec,
+        args: &[String],
+    ) -> anyhow::Result<(Store<wasi_common::WasiCtx>, wasmtime::Func)> {
+        // already in the cgroup
+        let env = oci_wasmtime::env_to_wasi(spec);
+        log::info!("setting up wasi");
+
+        let path = wasi_dir(".", OpenOptions::new().read(true))?;
+        let mut wasi_builder = WasiCtxBuilder::new()
+            .args(args)?
```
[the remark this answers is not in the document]
jsturtevant: could you explain a bit more?

jsturtevant (commented):
Looks great, I am looking with an eye towards how this would work when Windows is added and I think this actually makes a few things easier.  The awkward part might be re-using the `WasmtimeExecutor` directly vs via a youki `Container` since Windows isn't supported in youki but I think it will work.

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -81,200 +46,178 @@ fn load_spec(bundle: String) -> Result<oci::Spec, Error> {
...
-    Ok((wctx, module, method.to_string()))
-}
-
 impl Instance for Wasi {
     type E = wasmtime::Engine;
-    fn new(_id: String, cfg: Option<&InstanceConfig<Self::E>>) -> Self {
-        let cfg = cfg.unwrap(); // TODO: handle error
+    fn new(id: String, cfg: Option<&InstanceConfig<Self::E>>) -> Self {
+        // TODO: there are failure cases e.x. parsing cfg, loading spec, etc.
+        // thus should make `new` return `Result<Self, Error>` instead of `Self`
```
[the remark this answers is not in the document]
ipuustin: This was discussed before: https://github.com/containerd/runwasi/pull/54#issuecomment-1403269766

IMHO we should aim to remove as many `unwrap()` calls as possible from shim "main thread", because a library should not panic easily.

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -523,24 +441,68 @@ mod wasitest {
...
+    use tempfile::tempdir;
+
+    use super::*;
+
+    #[test]
+    fn test_maybe_open_stdio() -> Result<(), Error> {
+        let f = maybe_open_stdio("")?;
+        assert!(f.is_none());
+
+        let f = maybe_open_stdio("/some/nonexistent/path")?;
```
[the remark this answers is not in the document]
utam0k: 👍

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -81,200 +46,178 @@ fn load_spec(bundle: String) -> Result<oci::Spec, Error> {
...
+        thread::spawn(move || {
+            let (lock, cvar) = &*code;
+
+            let status = match waitid(WaitID::Pid(pid), WaitPidFlag::WEXITED) {
+                Ok(WaitStatus::Exited(_, status)) => status,
+                Ok(WaitStatus::Signaled(_, sig, _)) => sig as i32,
+                Ok(_) => 0,
+                Err(e) => {
+                    if e == Errno::ECHILD {
+                        0
```
utam0k, line 98: I assume this code came from my code, but it has to plant the log in this case.

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -0,0 +1,133 @@
...
+        true
+    }
+
+    fn name(&self) -> &'static str {
+        EXECUTOR_NAME
+    }
+}
+
+impl WasmtimeExecutor {
+    fn prepare_function(
```
utam0k, line 55: We know at a glance that this is a function.
```suggestion
    fn prepare(
```

crates/containerd-shim-wasmedge/src/instance.rs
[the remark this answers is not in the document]
Mossaka: This is a personal preference. I moved the test down because I felt unit tests should be placed at the bottom of the file. I tried to search for style guide on this but could not find anything. With that said, I haven o strong opinions on where the tests go and I am happy to revert this change.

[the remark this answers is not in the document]
Mossaka: Nevermind, I misunderstood your comment. Yes I will move this to the library crate.

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: Could you elaborate more about "plant the log"? Do you want me to log a message when `Errno::ECHILD` was returned?

crates/containerd-shim-wasmtime/src/executor.rs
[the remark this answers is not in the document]
Mossaka: This is from a discussion at the CG meeting this week. Runwasi currently has no ways to specify, explicitly, what file and directories a wasm container have access to.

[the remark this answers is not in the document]
jsturtevant: ah yes, Do we have an issue to track this?

crates/common/Cargo.toml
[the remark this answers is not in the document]
jsturtevant: ah that makes sense, Maybe it could be hidden behind a cargo `feature` in that crate?  Would that be a good use case for using features?

crates/containerd-shim-wasmtime/src/executor.rs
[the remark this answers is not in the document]
Mossaka: No, but now we have it. #157

crates/common/Cargo.toml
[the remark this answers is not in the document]
Mossaka: I don't know if this is better than a common library. What's the major concern for having an extra crate in the repo?

crates/containerd-shim-wasmtime/src/instance.rs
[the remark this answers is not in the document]
Mossaka: We can raise an issue for tracking this one. No need to change the API in this PR.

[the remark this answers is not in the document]
Mossaka: Created one and resolving this convo.

crates/common/Cargo.toml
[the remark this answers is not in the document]
jsturtevant: Mostly a knee jerk reaction from experience, since it is being used by both shims and related to containerd-shim-wasm.  Ends up being one more dependency someone will need to manage somewhere.  In particular if one relies on the other and you end up having to update them both at the same time anyways. I didn't look hard at how closely related these are and how interdependent (they seem to have disappeared from the current code pushed so maybe there is something here). 

As a side note, I try to avoid modules/crates that are called `common` as they end up being a dumping ground for functionality.  And maybe this was part of the reaction as well, if it was something like `wasm-libcontainer` maybe make more sense?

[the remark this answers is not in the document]
Mossaka: Okay I will try the `feature` suggestion!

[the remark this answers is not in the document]
Mossaka: Done!

crates/containerd-shim-wasm/src/wasm_libcontainer/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Checks if the container exists.
+pub fn container_exists<P: AsRef<Path>>(root_path: P, container_id: &str) -> Result<bool> {
+    let container_root = construct_container_root(root_path, container_id)?;
+    Ok(container_root.exists())
+}
+
+/// containerd can send an empty path or a non-existant path
+/// In both these cases we should just assume that the stdio stream was not setup (intentionally)
+/// Any other error is a real error.
+pub fn maybe_open_stdio(path: &str) -> Result<Option<RawFd>, Error> {
```
jsturtevant, line 33: this doesn't feel `libcontainer` specific?

crates/containerd-shim-wasm/src/wasm_libcontainer/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+};
+
+/// Loads the container state from the given root path.
+pub fn load_container<P: AsRef<Path>>(root_path: P, container_id: &str) -> Result<Container> {
+    let container_root = construct_container_root(root_path, container_id)?;
+    if !container_root.exists() {
+        bail!("container {} does not exist.", container_id)
+    }
+
+    Container::load(container_root)
```
jsturtevant, line 20: Following up on the conversation around modules and crates, now that the source is back.  

This is the only line that feels like it is specific to libcontainer.  The other functions are useful regardless (I needed to canonicalize the root path in windows too).  

I like pulling those other functions into this crate but not sure they need to be behind a flag.  And since it is one line to load the root maybe don't need duplicate this function?

Might be missing something, just some thoughts

Mossaka: Yeah maybe I can get rid of this entirely. Let me do some refactoring work there

Mossaka: This commit https://github.com/containerd/runwasi/pull/142/commits/b606b25373ddd3b933099064dd49fc18fe280359 removes the feature flag entirely.

Mossaka: I still feel that, although now these functions don't depend on `libcontainer`, the assumption that there is a `root_path` to a container runtime's state is pretty libcontainer dependent, no?

crates/containerd-shim-wasmtime/src/instance.rs
utam0k: >  when Errno::ECHILD was returned?

Yes 👍

Mossaka:
I am trying to fix the failing integration tests on this one.

crates/containerd-shim-wasm/src/wasm_libcontainer/mod.rs
jsturtevant: From what I can tell the only part that is libcontainer dependent is adding the `container_id` to the end of the cannocal path, since ` Container::load(container_root)` and adding `cointainer_id` are both one line, I don't know that sharing logic in a common script is worth it, and I think it would be better to just inline those in the respective shims.

Mossaka:
Hey there! The tests should all be passing now! Could you please take another look?

Mossaka:
Umm this is strange, I could not re-produce the pipeline error locally at my ubuntu 22.04 machine https://github.com/containerd/runwasi/actions/runs/5395848131/jobs/9798781514?pr=142

utam0k:
@Mossaka How about using this? You can use bpftrace on the CI env.
https://github.com/marketplace/actions/debugging-with-tmate

Mossaka:
All the tests are passing 🥳

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -1,280 +1,284 @@
...
+                stdin,
+                stdout,
+                stderr,
+                engine,
+            })])?
+            .with_root_path(self.rootdir.clone())?
+            .as_init(&self.bundle)
+            .with_systemd(false)
+            .build()?;
+        Ok(container)
```
arschles, line 256: Not entirely sure the return type of `build()`, but will this work?

```suggestion
        ContainerBuilder::new(self.id.clone(), syscall.as_ref())
            .with_executor(vec![Box::new(WasmtimeExecutor {
                stdin,
                stdout,
                stderr,
                engine,
            })])?
            .with_root_path(self.rootdir.clone())?
            .as_init(&self.bundle)
            .with_systemd(false)
            .build()
```

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -1,280 +1,284 @@
...
+    }
+}
+
+impl EngineGetter for Wasi {
+    type E = wasmtime::Engine;
+    fn new_engine() -> Result<Engine, Error> {
+        let engine = Engine::default();
+        Ok(engine)
```
arschles, line 264:
```suggestion
        Ok(Engine::default())
```

arschles (changes requested):

crates/containerd-shim-wasm/src/sandbox/instance_utils.rs
```diff
@@ -0,0 +1,83 @@
...
+        return Ok(None);
+    }
+    match OpenOptions::new().read(true).write(true).open(path) {
+        Ok(f) => Ok(Some(f.into_raw_fd())),
+        Err(err) => match err.kind() {
+            ErrorKind::NotFound => Ok(None),
+            _ => Err(err.into()),
+        },
+    }
+}
```
danbugs, line 47: ~consider making some of these `fn`s `pub(crate)`.

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -0,0 +1,111 @@
...
+        let mut store = Store::new(&self.engine, wctx);
+
+        log::info!("instantiating instance");
+        let i = linker.instantiate(&mut store, &module)?;
+
+        log::info!("getting start function");
+        let f = i
+            .get_func(&mut store, method)
+            .ok_or_else(|| anyhow!("module does not have a wasi start function".to_string()))?;
+        Ok((store, f))
```
danbugs, line 109:
```suggestion
        log::info!("instantiating instance");
        let instance = linker.instantiate(&mut store, &module)?;

        log::info!("getting start function");
        let start_func = instance
            .get_func(&mut store, method)
            .ok_or_else(|| anyhow!("module does not have a WASI start function".to_string()))?;
        Ok((store, start_func))
```

nit

danbugs (commented):
LGTM, Joe – just some small comments here and there.

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use crate::oci_wasmtime::{self, wasi_dir};
+
+const EXECUTOR_NAME: &str = "wasmtime";
+
+pub struct WasmtimeExecutor {
+    pub stdin: Option<RawFd>,
+    pub stdout: Option<RawFd>,
+    pub stderr: Option<RawFd>,
+    pub engine: Engine,
```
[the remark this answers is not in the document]
Mossaka: I am not sure what benefits would it gain by using readonly at the cost of adding an extra dependency. Would you like to elaborate more about it?

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -0,0 +1,111 @@
...
+        let wasi_builder = WasiCtxBuilder::new()
+            .args(args)?
+            .envs(env.as_slice())?
+            .inherit_stdio()
+            .preopened_dir(path, "/")?;
+
+        if let Some(stdin) = self.stdin {
+            let _ = dup(STDIN_FILENO);
+            let _ = dup2(stdin, STDIN_FILENO);
+        }
```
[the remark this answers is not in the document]
Mossaka: Ah yes, great suggestion!!

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: Unfortunately this does not work as the `build` returns a `LibcontainerError`

crates/containerd-shim-wasm/src/wasm_libcontainer/mod.rs
Mossaka: Do you think that `get_instance_root`, `instance_exists` and `maybe_open_stdio` are worth in a common script?

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: I have added a log messaging saying "no child process" when `Errno::ECHILD` was returned in both the shims.

Mossaka: I am not entirely sure when people see this log message, that actions they can do to mitigate. Or is this an issue at all in the first place? I would appreciate if anyone can help me to understand it.

crates/containerd-shim-wasmtime/src/executor.rs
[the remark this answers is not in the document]
arschles: at current, they can be mutated by anybody inside or out of the module and crate. that seems like it could be problematic in the long term. another option, if you don't like the extra dependency, could be just make these module- or crate-private and provide getters as necessary.

arschles (approved):
LGTM

crates/containerd-shim-wasmtime/src/executor.rs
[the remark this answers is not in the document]
Mossaka: Ah I see. I am a bit hesitant because I believe this file / module is only used as a binary, not intended to use as a library, and hence mutation by anybody outside of the crate is less of a concern.

Mossaka:
Could you please take another look? 🙏 @utam0k , @ipuustin , @cpuguy83

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub stderr: Option<RawFd>,
+    pub engine: Engine,
+}
+
+impl Executor for WasmtimeExecutor {
+    fn exec(&self, spec: &Spec) -> Result<(), ExecutorError> {
+        let args = oci::get_args(spec);
+        if args.is_empty() {
+            return Err(ExecutorError::InvalidArg);
+        }
```
utam0k, line 29: Let's check the number of args too.

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -0,0 +1,111 @@
...
+            dup(STDERR_FILENO)?;
+            dup2(stderr, STDERR_FILENO)?;
+        }
+
+        log::info!("building wasi context");
+        let wctx = wasi_builder.build();
+
+        log::info!("wasi context ready");
+        let start = args[0].clone();
+        let mut iterator = start.split('#');
```
utam0k, line 86:
```suggestion
        let mut iterator = args
            .first()
            .context("args must have at least one argument.")?
            .split('#');
```

crates/containerd-shim-wasm/src/sandbox/instance_utils.rs
```diff
@@ -0,0 +1,83 @@
+//! Common utilities for the containerd shims.
```
utam0k, line 1: 👍 😍

crates/containerd-shim-wasmtime/src/executor.rs
Mossaka: Thanks!

utam0k (approved):
Looks good to me, but This PR introduces a big change, so I recommend getting one more approval.

cpuguy83 (approved):

Mossaka:
Thanks everyone for reviewing!! 🥂🙌

Issue #131, opened 2023-06-08
Use youki libcontainer crate for all shims


jprendes:
The work in #78 enabled `youki`'s `libcontainer` for the WasmEdge shim.
The Issue #110 intents to repeat the same work for the wasmtime shim.
During #78, @rumpl 's [comment](https://github.com/containerd/runwasi/pull/78#issuecomment-1485569345) mentioned that we could enable `youki`'s `libcontainer` in a lower level in `runwasi`, so that all shims can benefit from it.
Any thoughts?

ipuustin:
I did try to use libcontainer functions for adding support for OCI devices and seccomp. The seccomp code is at https://github.com/ipuustin/runwasi/commits/seccomp and the devices code is here: https://github.com/ipuustin/runwasi/commits/oci-devices. My finding was that libcontainer can very well be used on this level, but it is a lot of effort and will bring a lot of duplicated code to runwasi. I mean, the seccomp code I needed to add in the seccomp branch alone is hundreds of lines of code, a lot of it `unsafe`.

However, what would make sense IMHO would be to go through the libcontainer initialization code and figure out what exactly is needed for running Wasm/WASI modules securely. For example, do we really need a PID namespace if the Wasm module will not even have the concept of a process? Or does it make sense to have the Masked Paths OCI feature, if we anyway control precisely which files are available? If the result of this study is that the optimal Wasm initialization path is a lot different from traditional container initialization, we might want to contribute to libcontainer a different way for loading the container, or maybe just do it using the lower-level primitives as you suggest.

I'm myself curious to what we could do in runwasi shared mode using just threads for running the Wasm VMs -- would we get anywhere close to a working "container"?

ipuustin:
Most of the seccomp code is in containerd-shim-wasm (see https://github.com/ipuustin/runwasi/commit/78c3c4d9e3eef9d04e0835e81576de96fc62b41b). It's just being called from the WasmEdge and Wasmtime shims. The reason there needs to be some code in shims is that the open FD for seccomp notification needs to be passed over socket to main process and then to the external notifier process.

cpuguy83:
> However, what would make sense IMHO would be to go through the libcontainer initialization code and figure out what exactly is needed for running Wasm/WASI modules securely

The main thing here is we are using linux containerization as an extra boundary, plus it also helps wasm play nicely with other linux containers/processes.

[...]

PR #147, opened 2023-06-16
Handle Wasm Images with wasm media types

jsturtevant:
fixes: #108 (requires changes in containerd) 

This adds the ability to handle oci artifacts as described in the proposal in https://docs.google.com/document/d/11shgC3l6gplBjWF1VJCWvN_9do51otscAm0hBDGSSAc. 

~Containerd doesn't currently support passing the annotations for the OCI artifact format and needs the following PR: https://github.com/containerd/containerd/pull/8699~

This pr relies on https://github.com/containerd/containerd/pull/9142 which has been merged upstream


jsturtevant:
This should likely go in after #142, will need some refactoring to support the executor. But wanted to get some eyes on the general direction.

crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -32,16 +33,24 @@ where
...
+    containerd_address: Option<String>,
 }
 
 impl<E> InstanceConfig<E>
 where
     E: Send + Sync + Clone,
 {
     pub fn new(engine: E, namespace: String) -> Self {
+        let os_args: Vec<_> = env::args_os().collect();
+        let containerd_address = match containerd::parse(&os_args[1..]) {
```
jsturtevant, line 46: I tried this in a few places (95d595541eeb47d67261c46223d837da6186093e / 89bb778251fd3f551849d350a11a79350712af5a) and landed here as a decent place but happy to hear feedback. 

I think really this shouldn't be something we need to do but should be passed in when the shim is created (https://github.com/containerd/runwasi/pull/147/files#diff-8f8704e2bb5644be97cfbd99c736cb7f432da2de20292de2b0f69e4193bd0f64R1360)

crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -1357,6 +1357,7 @@ where
     type T = Local<I, E>;
 
     fn new(_runtime_id: &str, id: &str, namespace: &str, _config: &mut shim::Config) -> Self {
+        // Ideally this function passes in either the containerd address or the flags from the cli
```
jsturtevant, line 1360: https://github.com/containerd/rust-extensions/pull/150

jsturtevant:
This no longer works after #131 since the entrypoint points to the wasm module which is no longer in the container image.  Will need to re-think how to handle that scenario.  Similar situation to #194

crates/containerd-shim-wasm/src/sandbox/shim.rs
jsturtevant: fixed in #192

crates/containerd-shim-wasm/src/sandbox/instance.rs
jsturtevant: fixed in #192

jsturtevant:
this is working using https://github.com/containerd/containerd/pull/8699 and https://github.com/bytecodealliance/registry/pull/146.  See demo: https://www.youtube.com/watch?v=SEwBcEFgXUM&feature=youtu.be

Currently blocked on  https://github.com/containers/youki/issues/2220, I am using a custom build that disables the verification for now.

jsturtevant:
after https://github.com/containerd/runwasi/pull/290 this should be unblocked will give it try soon

crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
```diff
@@ -84,6 +92,10 @@ impl<E: Engine> Executor<E> {
 }
 
 fn is_linux_container(spec: &Spec) -> Result<()> {
+    if spec.is_wasm_oci() {
```
jsturtevant, line 95: we might be able to move this up to `inner()` above

crates/containerd-shim-wasm/src/sandbox/oci.rs
```diff
@@ -4,13 +4,35 @@ use std::collections::HashMap;
...
+    let mut spec = Spec::load(bundle.join("config.json").to_str().unwrap())?;
+    spec.canonicalize_rootfs(&bundle)?;
+    Ok(spec)
+}
+
+pub fn get_oci_artifact(spec: &Spec) -> Option<String> {
+    match spec.annotations().clone() {
+        Some(annotations)
+            if annotations
+                .contains_key("application/vnd.bytecodealliance.wasm.component.layer.v0+wasm") =>
```
jsturtevant, line 25: Is there a canonical location for this in another library or should I move this to a const some where in the project.

jsturtevant:
@devigned @0xE282B0 @jprendes this is ready for review, I am going to work on adding an integration test tomorrow

crates/containerd-shim-wasm/src/sandbox/oci.rs
jsturtevant: put it in the oci package for now

crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
jsturtevant: moved this up

jsturtevant:
There is now a kind e2e test for oci images using the contained PR that has support.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,179 @@
+#![cfg(unix)]
```
jsturtevant, line 1: opened #205

crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -20,18 +20,21 @@ crossbeam = { workspace = true }
 libc = { workspace = true }
 log = { workspace = true }
 oci-spec = { workspace = true }
+artifact-spec = { package = "oci-spec", git="https://github.com/containers/oci-spec-rs", rev="c109fe19b2548a86f20e48b18831bf17eb72c494", features = ["runtime"] }
```
devigned, line 23: I believe this will prevent us from being able to publish crates since it is not a crate dependency.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,196 @@
...
+        match resp {
+            Ok(layer_option) => layer_option,
+            Err(_) => None,
+        }
+    }
+
+    // load module will query the containerd store to find an image that has an ArtifactType of WASM OCI Artifact
+    // If found it continues to parse the manifest and return the layers that contains the WASM modules
+    // and possibly other configuration artifacts
+    pub fn load_modules(&mut self, containerd_id: String) -> Option<Vec<oci::OciArtifact>> {
```
devigned, line 140: Many of the fn's in this file return option and log warnings. I'm wondering if it would be better to return a result and bubble the error up rather than logging. WDYT?

crates/containerd-shim-wasm/src/sandbox/oci.rs
```diff
@@ -7,10 +7,23 @@ use std::os::unix::process::CommandExt;
 use std::process;
 
 use anyhow::Context;
+use oci_spec::image::Descriptor;
 pub use oci_spec::runtime::Spec;
 
 use super::error::Result;
 
+pub const COMPONENT_ARTIFACT_TYPE: &str = "application/vnd.bytecodealliance.component.v1+wasm";
+pub const MODULE_ARTIFACT_TYPE: &str = "application/vnd.bytecodealliance.module.v1+wasm";
```
devigned, line 16: ✔️

crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
```diff
@@ -26,11 +27,12 @@ pub(crate) struct Executor<E: Engine> {
     engine: E,
     stdio: Stdio,
     inner: OnceCell<InnerExecutor>,
+    oci_artifacts: Option<Vec<OciArtifact>>,
 }
 
 impl<E: Engine> LibcontainerExecutor for Executor<E> {
     fn validate(&self, spec: &Spec) -> Result<(), ExecutorValidationError> {
-        // We can handle linux container. We delegate wasm container to the engine.
+        // We can handle linux container. We delergate wasm container to the engine.
```
devigned, line 35:
```suggestion
        // We can handle linux container. We delegate wasm container to the engine.
```

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -36,13 +36,27 @@ impl Engine for WasmtimeEngine {
...
         log::info!("wasi context ready");
         let WasiEntrypoint { path, func } = ctx.wasi_entrypoint();
-        let path = path
-            .resolve_in_path_or_cwd()
-            .next()
-            .context("module not found")?;
 
-        log::info!("loading module from file {path:?}");
-        let module = Module::from_file(&self.engine, &path)?;
+        let module = match ctx.oci_artifacts() {
```
devigned, line 40: Might be nice to have some shared utilities to dry up the repeated code blocks.

crates/oci-tar-builder/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition.workspace = true
 tar = { workspace = true }
 sha256 = { workspace = true }
 log = { workspace = true }
-oci-spec =  { workspace = true }
+oci-spec = { git="https://github.com/containers/oci-spec-rs", rev="c109fe19b2548a86f20e48b18831bf17eb72c494", features = ["runtime"] }
```
devigned, line 10: Should this be a workspace version?

crates/wasi-demo-app/Cargo.toml
```diff
@@ -8,7 +8,7 @@ tar = { workspace = true, optional = true }
 sha256 = { workspace = true, optional = true }
 log = { workspace = true, optional = true }
 env_logger = { workspace = true, optional = true }
-oci-spec = { workspace = true, optional = true }
+oci-spec = { git="https://github.com/containers/oci-spec-rs", rev="c109fe19b2548a86f20e48b18831bf17eb72c494", features = ["runtime"], optional=true }
```
devigned, line 11: Should this be a workspace version?

devigned (changes requested):
Great work on this @jsturtevant! I think it looks pretty solid. Just a few comments and thoughts.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: I considered this.  I was trying to consider the failure state, we don't know if this is artifact until we've done some basic parsing.  

If it isn't an artifact or we fail to gather the info from contianerd, should we hard error and not start the container or should we continue on and try to load like it is a container as a best effort (not an artifact)?

I opted to continue on as if it wasn't a container image since current validation checks would fail and the container would restart. So if there was a temporary reason, we weren't able to gather the correct information it would get tried again.

I can see the other approach here, is if we can't gather the info then we should fail fast.  thoughts?

crates/oci-tar-builder/Cargo.toml
jsturtevant: same reasoning here https://github.com/containerd/runwasi/pull/147/files#r1346187095

crates/containerd-shim-wasmtime/src/instance.rs
jsturtevant: open to ideas... I not sure the best way to do this and maybe it's to early? it is likely the processing will change once we start to move over to preview2, which some runtimes will support and some won't?

Another thing to consider is implementing something like pre-compliation (#44) and how each runtime composes components together.  Should that be done before and only the compiled/composed module passed in?

crates/containerd-shim-wasm/Cargo.toml
jsturtevant: One thing I could do is just parse the fields required to do the work and only use this commit in the test image projects but I did open https://github.com/containers/oci-spec-rs/pull/147

jsturtevant:
ci failure is  `cp: cannot stat '/var/lib/rancher/k3s/agent/etc/containerd/config.toml': No such file or directory` https://github.com/containerd/runwasi/issues/322#issuecomment-1751267331

crates/containerd-shim-wasm/Cargo.toml
jsturtevant: https://github.com/containerd/runwasi/pull/352

jsturtevant: updated

crates/containerd-shim-wasm/src/sandbox/containerd.rs
devigned: Thank you for explaining. I get now that the executor is optionally being populated with the artifact so that in a later stage it can decide how to load with best effort.

I don't know that you have to fail fast. I'm suggesting returning a typed error to the caller so the caller would have the information to decide to fail, log, or do nothing. It's more a nit about hiding errors.

I was suggesting a signature of:
```
pub fn load_modules(&mut self, containerd_id: String) -> Result<Option<Vec<oci::OciArtifact>>>
```

Then instance.rs would be responsible for logging or handling that error. You could still pass in the None to the executor. However, it seems a bit odd to pass in None to the executor, but I guess you have to get the artifact into the executor somehow.

crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
```diff
@@ -62,17 +70,22 @@ impl<E: Engine> LibcontainerExecutor for Executor<E> {
...
             inner: Default::default(),
+            oci_artifacts,
         }
     }
 
     fn inner(&self, spec: &Spec) -> &InnerExecutor {
+        let is_wasm = self.oci_artifacts.is_some();
         self.inner.get_or_init(|| {
-            if is_linux_container(spec).is_ok() {
+            // if the spec has oci annotations we know it is wasm so short circuit checks
```
jprendes, line 85: **question:** I'm not sure what the right thing to do is here.
If we have wasm annotations, we don't need to check for linux containers, I agree.
But a runtime may want to do other checks before deciding it can handle it? e.g., in the future they may want to check if a component is a _reactor_ or a _command_?

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -20,13 +22,38 @@ pub trait RuntimeContext {
...
 }
 
 pub struct WasiEntrypoint {
     pub path: PathBuf,
     pub func: String,
 }
 
+pub(crate) struct WasiContext<'a> {
+    pub spec: &'a Spec,
+    pub oci_artifacts: Option<Vec<OciArtifact>>,
```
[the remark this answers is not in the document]
jprendes: If `WasiContext` becomes _zero-copy_, then I would argue that we only need to implement `RuntimeContext` for `WasiContext` and not for `Spec`.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: > I don't know that you have to fail fast. I'm suggesting returning a typed error to the caller so the caller would have the information to decide to fail, log, or do nothing. It's more a nit about hiding errors.

Make sense, I'll make this update

crates/wasi-demo-app/Cargo.toml
```diff
@@ -8,7 +8,7 @@ tar = { workspace = true, optional = true }
 sha256 = { workspace = true, optional = true }
 log = { workspace = true, optional = true }
 env_logger = { workspace = true, optional = true }
-oci-spec = { workspace = true, optional = true }
+oci-spec = { workspace = true, features = ["runtime"], optional=true }
```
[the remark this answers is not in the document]
jsturtevant: ah, typo from when I needed to pull in custom oci-spec

crates/containerd-shim-wasm/src/sandbox/oci.rs
jsturtevant: This will be removed.  The images produced with wasm module/component layers will not be considered artifacts since for now they use the `config.mediatype` of `application/vnd.oci.image.config.v1+json` to be compatible with containerd, see dissussion in https://github.com/opencontainers/image-spec/pull/1137 and https://github.com/opencontainers/image-spec/pull/1141

.github/workflows/ci.yml
```diff
@@ -75,7 +84,22 @@ jobs:
...
       runtime: ${{ matrix.runtime }}
-      
+      test-command: make test/k8s-${{ matrix.runtime }}
+
+  e2e-k8s-oci:
+    name: ${{ matrix.runtime }}
+    needs: [build-ubuntu, test-image-oci]
+    strategy:
+      matrix:
+        os: ["ubuntu-22.04"]
```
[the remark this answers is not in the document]
jsturtevant: this is being dropped in #359

Makefile
```diff
@@ -101,13 +101,33 @@ dist: $(RUNTIMES:%=dist-%);
...
+.PHONY: test-image/clean
 test-image/clean:
 	rm -rf target/wasm32-wasi/$(OPT_PROFILE)/
 
+.PHONY: test-image/oci/clean
+test-image/oci/clean:
+	rm -rf target/wasm32-wasi/$(OPT_PROFILE)/img-oci.tar
+
+.PHONY: demo-app
+demo-app: target/wasm32-wasi/$(OPT_PROFILE)/wasi-demo-app.wasm
```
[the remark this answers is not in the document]
jsturtevant: I was using it locally during one of my workflows. I'm fine dropping it if not useful, its really an alias for `target/wasm32-wasi/$(OPT_PROFILE)/wasi-demo-app.wasm` which is hard to type out :-)

crates/containerd-shim-wasm/src/container/context.rs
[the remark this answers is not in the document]
jsturtevant: SGTM, I had done something similar to the suggested changes originally

jprendes:
Needs rebase

Makefile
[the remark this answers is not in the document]
jprendes: Happy to keep it :-)

Makefile
```diff
@@ -122,15 +142,30 @@ dist/img.tar:
...
 
+dist/img-oci.tar: target/wasm32-wasi/$(OPT_PROFILE)/img-oci.tar
+	@mkdir -p "dist/"
+	cp "$<" "$@"
+
 load: dist/img.tar
 	sudo ctr -n $(CONTAINERD_NAMESPACE) image import --all-platforms $<
 
+load/oci: dist/img-oci.tar
+	sudo ../containerd/bin/ctr -n $(CONTAINERD_NAMESPACE) image import --all-platforms $<
```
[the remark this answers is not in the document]
jprendes: Can we assume that the ctr version in the ci runners is up to date?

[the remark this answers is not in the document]
jsturtevant: The tests I've added for now in CI run in kind with a containerd version that has the patches required. I've made note of required versions in the `readme.md`.  Once `kind` releases a new version it should have the patches and we can drop the `dockerfile.oci`

crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
jsturtevant: good point, I was trying to avoid the extra cycles of having to parse the executable files, but maybe it makes more sense to have this in the default `can_handle` of `trait Engine` so it can be overridden if required?

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,148 @@
...
+            let name = image_name.to_string();
+            let req = GetImageRequest { name };
+            let req = with_namespace!(req, self.namespace);
+            let digest = ImagesClient::new(self.inner.clone())
+                .get(req)
+                .await
+                .map_err(|err| ShimError::Others(err.to_string()))?
+                .into_inner()
+                .image
+                .ok_or(ShimError::Others(format!(
```
jprendes, line 76: Can we switch this (and the others) `ok_or` with `ok_or_else` to avoid the `format!` when it's not needed ?

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
```diff
@@ -39,8 +41,16 @@ impl<E: Engine> SandboxInstance for Instance<E> {
         let rootdir = determine_rootdir(&bundle, &namespace, rootdir)?;
         let stdio = Stdio::init_from_cfg(cfg)?;
 
+        // check if container is OCI image with wasm layers and attempt to read the module
+        let modules = containerd::Client::connect(cfg.get_containerd_address(), &namespace)?
+            .load_modules(&id)
+            .unwrap_or_else(|e| {
+                log::warn!("Error obtaining wasm layers for container {}.  Will attempt to use files inside containerd image. Error: {}", id.clone(), e);
```
jprendes, line 48: **fix:** unneeded clone

```suggestion
                log::warn!("Error obtaining wasm layers for container {}.  Will attempt to use files inside containerd image. Error: {}", id, e);
```

or even better:
```suggestion
                log::warn!("Error obtaining wasm layers for container {id}.  Will attempt to use files inside containerd image. Error: {e}");
```

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -90,47 +98,51 @@ impl Client {
...
             .as_ref()
-            .ok_or(ShimError::Others("manifest is not an OCI Artifact".to_string()))?;
+            .ok_or(ShimError::Others("failed to extract OS".to_string()))?
+            .os();
 
-        match artifact_type {
-            MediaType::Other(s) if s == COMPONENT_ARTIFACT_TYPE || s == MODULE_ARTIFACT_TYPE => {
-                log::info!("manifest with OCI Artifact of type {s}");
+        match os {
+            Os::Other(s) if s == "wasm" => {
```
jprendes, line 128: Isn't the os `wasi` / `wasip1`, while `wasm` is the arch?

Also, some images use `wasm32`, but that's ok because this would only apply to new images.

jprendes (approved):
Thanks for all your work on this!
Main comment below is around the OS vs ARCH for the image.

jprendes:
CI failure is due to missing `protoc` from the build images.

crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
jprendes: I am happy to keep it as is.
We might want to revisit this in the future.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: yup! fixed.

jsturtevant:
not sure why the kind-oci test is failing, I ran into #362 when trying locally and will revisit tomorrow now that it is fixed

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,154 @@
...
+        let manifest = self.read_content(digest)?;
+        let manifest = manifest.as_slice();
+        let manifest = ImageManifest::from_reader(manifest)?;
+
+        let arch = manifest
+            .config()
+            .platform()
+            .as_ref()
+            .ok_or_else(|| ShimError::Others("failed to extract platform".to_string()))?
+            .architecture();
```
jprendes, line 131: It would be great to expose the platform in the `WasiContext`.
We could use the os part (`wasip1`/`wasip2`) to choose between modules or components, and `os_features` might be useful to know what components we need.

Also, that could also be used to choose between wasm container vs native container, right?

jsturtevant: Like this idea of passing OS and Arch to the wasicontext.  For features, my understanding is we have some flexibility and we should follow up on how we decide to hand those.

jprendes: To be clear, I'm not implying we should start using `os_features` right now.
The details on that would need wider concensus and review.
But if we expose the `Platform` object (instead of just the OS / Arch strings), we open the door to that option.

jsturtevant: agreed!

crates/containerd-shim-wasm/src/sandbox/containerd.rs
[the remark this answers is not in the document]
Mossaka: Realized I commented on the old commits. Closing this

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    // load module will query the containerd store to find an image that has an ArtifactType of WASM OCI Artifact
+    // If found it continues to parse the manifest and return the layers that contains the WASM modules
+    // and possibly other configuration artifacts
+    pub fn load_modules(&mut self, containerd_id: String) -> Option<Vec<oci::OciArtifact>> {
+        let image_name = self.get_oci_image(containerd_id.clone())?;
+        let container_digest = self.get_image_content_sha(image_name)?;
+        let manifest = match self.read_content(container_digest) {
+            Ok(m) => m,
+            Err(_) => {
```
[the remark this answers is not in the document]
Mossaka: Realized I commented on the old commits. Closing this

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,196 @@
...
+                log::warn!("failed to read manifest");
+                return None;
+            }
+        };
+        let manifest = match ImageManifest::from_reader(manifest.as_slice()) {
+            Ok(m) => m,
+            Err(_) => {
```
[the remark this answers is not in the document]
Mossaka: Realized I commented on the old commits. Closing this

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,196 @@
...
+                log::info!("manifest is not an OCI Artifact");
+                return None;
+            }
+        }
+
+        let mut oci_artifacts = Vec::<oci::OciArtifact>::new();
+        for layer in manifest.layers().iter() {
+            let module = match self.read_content(layer.digest().clone()) {
+                Ok(m) => m,
+                Err(_) => {
```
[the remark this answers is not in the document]
Mossaka: Realized I commented on the old commits. Closing this

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,196 @@
...
+                return None;
+            }
+            None => {
+                log::info!("manifest is not an OCI Artifact");
+                return None;
+            }
+        }
+
+        let mut oci_artifacts = Vec::<oci::OciArtifact>::new();
+        for layer in manifest.layers().iter() {
```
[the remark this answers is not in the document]
Mossaka: Realized I commented on the old commits. Closing this

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -36,13 +36,24 @@ impl Engine for WasmtimeEngine {
...
+                    .next()
+                    .context("module not found")?;
+
+                Module::from_file(&self.engine, path)?
+            }
+            [module] => {
+                log::info!("loading module wasm OCI layers");
+                Module::from_binary(&self.engine, &module.layer)?
+            }
+            [..] => bail!("only a single module is supported when using images with OCI layers"),
```
Mossaka, line 54: Do we intend to support multiple images in an OCI artifact in the future?

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,196 @@
...
+use containerd_client::tonic::transport::Channel;
+use containerd_client::tonic::{self, Streaming};
+use containerd_client::with_namespace;
+use tokio::runtime::Runtime;
+use tonic::Request;
+
+use crate::sandbox::error::{Error as ShimError, Result};
+use crate::sandbox::oci::{self, OciArtifact};
+
+pub struct Client {
```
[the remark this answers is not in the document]
jprendes: This doesn't need to be `pub`, right? Maybe just `pub(super)` or `pub(crate)`.

crates/containerd-shim-wasmtime/src/instance.rs
jprendes: I was assuming that was there for components.

jsturtevant: The shims don't currently have the component features turned on, so wouldn't know how to handle multi components.  As a follow up we can enable multiple components

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,196 @@
...
+    inner: Channel,
+    rt: Runtime,
+    namespace: String,
+}
+
+// sync wrapper implementation from https://tokio.rs/tokio/topics/bridging
+impl Client {
```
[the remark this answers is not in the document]
jsturtevant: The user here is our Sandbox instance which is sync. The idea was since this the only use I wanted to avoid polluting the sandbox instance with the async related code there. But this module has evolved a bit from when I first implemented it and has become simpler.  It might not be much to wrap it at that layer.  WDYT? I don't know that I have an opinion here.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,196 @@
...
+            }
+            None => {
+                log::info!("manifest is not an OCI Artifact");
+                return None;
+            }
+        }
+
+        let mut oci_artifacts = Vec::<oci::OciArtifact>::new();
+        for layer in manifest.layers().iter() {
+            let module = match self.read_content(layer.digest().clone()) {
```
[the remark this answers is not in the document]
jsturtevant: this is no longer there

crates/containerd-shim-wasm/src/sandbox/containerd.rs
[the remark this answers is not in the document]
Mossaka: Thanks for the explaination. I am okay with keeping it as is

Mossaka (dismissed):
lgtm! Thanks for this awesome work!

jprendes:
CI is failling with
```
[ERROR] failed to initialize container process: wasmedge executor can't handle spec
```

Makefile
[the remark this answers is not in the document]
jsturtevant: i've added a check until these changes propagate to the all the distros, to avoid any confusion in local dev

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: I've passed this on to the WasiContext, but not sure I like it the current way I did it.  input welcome

jprendes (commented):
Thanks @jsturtevant !
I left some minor comments.

Mossaka (dismissed):
lgtm!

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,176 @@
...
+        | MediaType::ImageLayerNonDistributableZstd
+        | MediaType::ImageLayerZstd => true,
+        MediaType::Other(s)
+            if s.as_str()
+                .starts_with("application/vnd.docker.image.rootfs.") =>
+        {
+            true
+        }
+        _ => false,
+    }
```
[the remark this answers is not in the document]
jsturtevant: The reason I didn't do just wasm layer types is because runtimes like spin/slight (probably others) will need to pass there configuration files.  By leaving this more open ended, initially, we allow for those other types. In the case of spin they also have a media type for a module that doesn't match the `application/vnd.bytecodealliance.wasm` format.  

I would like to tighten this up in the future once we get to the point where "everything is component" approach.  The extra config files, static files, etc would just be component that the runtime knows how to use.  But feels to early to be restrictive here.

I am not too concerned about this logic change all that often.  It is a borrowed, idea from containerd's [IsLayerType](https://github.com/containerd/containerd/blame/18c9e7ec4c692cc00b7ac75d86d60685c270d48a/images/mediatypes.go#L127) and that hasn't had a change in 4 years.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -0,0 +1,176 @@
...
+        let platform: Platform = serde_json::from_slice(image_config)?;
+        match platform.architecture() {
+            Arch::Wasm => {
+                log::info!("found manifest with WASM OCI image format.");
+            }
+            _ => {
+                log::info!("manifest is not in WASM OCI image format");
+                return Ok(([].to_vec(), platform));
+            }
+        }
```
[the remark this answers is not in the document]
jsturtevant: didn't know you could do `let <> else`, so many different ways in rust 😱

jsturtevant:
heads up this has the commit from #362

crates/containerd-shim-wasm/src/sandbox/containerd.rs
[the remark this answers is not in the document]
jprendes: It's a relatively new addition to the language (since Rust 1.65) :-)

jsturtevant:
> heads up this has the commit from #362

I dropped it, as it shouldn't be needed.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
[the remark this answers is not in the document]
jprendes: We could add in the `Instance` trait a method to declare what layers it cares about (with a reasonable default impl).

Something like:
```rust
fn supported_layers_types() -> &'static [&'static str] {
}
```

or maybe something like:
```rust
fn supports_layer_type(ty: MediaType) -> bool {
    match ty {
        MediaType::Other(s) => s.starts_with("application/vnd.bytecodealliance.wasm."),
        _ => false,
    }
}
```

If you think the current approach is better I'm happy with that as well, I don't want to block it over this :-)

jprendes (approved):
LGTM

PR #156, opened 2023-06-23
Add `LinuxContainerExecutor` to Wasmtime and WasmEdge shims

Mossaka:
This PR adds a `LinuxContainerExecutor` to both the Wasmtime and Wasmedge shims. It enables the shims to be able to execute containers side by side with wasms. 

Some notable changes
1. It adds a new `libcontainer_instance` module to `containerd-shim-wasm` which provides a linux container executor implementation that is shared by both wasmtime and wasmedge shims. 
2. It adds this executor to both the shims. 
3. It adds nginx image to integration tests.


jsturtevant:
> It enables the wasmtime shim to be able to execute linux container if the wasm binary was not found on the rootfs.

We may need to revisit this since #147 would mean the wasm binary would definitely not be in the rootfs

Mossaka:
> > It enables the wasmtime shim to be able to execute linux container if the wasm binary was not found on the rootfs.
> 
> We may need to revisit this since #147 would mean the wasm binary would definitely not be in the rootfs

Yeah we can revisit this once #147 is merged.

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -81,200 +54,208 @@ fn load_spec(bundle: String) -> Result<oci::Spec, Error> {
...
+        }
+        if let Some(stderr) = stderr {
+            unsafe {
+                STDERR_FD = Some(dup(STDERR_FILENO));
+                dup2(stderr, STDERR_FILENO);
+            }
+        }
+
+        let container = ContainerBuilder::new(self.id.clone(), syscall.as_ref())
+            .with_executor(vec![wasmtime_executor, default_executor])?
```
Mossaka, line 229: could this have performance impact on linux containers?

utam0k: Which part do you care about? I think it's probably no problem.

Mossaka: This comment was made after a conversation with my collegue, Simon, at Microsoft. He suggested that in a pod spec where it contains multiple containers, and one wasm container, the executor selection logic, which checks if a container is a wasm container or not, will run for every single containers. This might have a performance impact on the startup time.

utam0k: @Mossaka That is correct. Each check process should be short. However, in general, that process should be small enough to be acceptable for users. This is because most of the time during container startup from docker/kubelet/podman... is spent on the container image.

Mossaka:
Could I please have some review comments on this PR? 🙏🙏

crates/containerd-shim-wasmtime/src/instance.rs
cpuguy83: What I'm thinking here is we could have 2 executors with the following order:

1. Read platform specific magic headers (Linux ELF / Windows PE)
2. Wasmtime exectuor

And in our wasmtime executor, assuming order depenendce in youki, we can assume anything that's not platform specific is wasm. Alternatively we can specifically inspect the wasm files (like we would need to do for ELF and PE) to check if they are wasm.

Mossaka: What benefits do we get by putting linux / windows executor prior to wasmtime executor?

cpuguy83: Mostly so we don't need to check if something is wasm.

jprendes: Excuse my ignorance, but would it be possible to check the image platform early on?
If it's `.*/wasm(32)?` assume it's a wasm container, otherwise a linux container.

cpuguy83: That could work assuming we have the data... like in an annotation or something?

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -39,8 +39,35 @@ impl Executor for WasmtimeExecutor {
...
+        let start = args[0].clone();
+        let mut iterator = start.split('#');
+        let mut cmd = iterator.next().unwrap().to_string();
+        let stripped = cmd.strip_prefix(std::path::MAIN_SEPARATOR);
+        if let Some(strpd) = stripped {
+            cmd = strpd.to_string();
+        }
+
+        let mut path = PathBuf::from(cmd);
+        if path.is_relative() {
```
jprendes, line 58: Does this handle `PATH` as `libcontainer` does for the entrypoint?
This is not needed if we don't validate the binary file.

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -39,8 +39,35 @@ impl Executor for WasmtimeExecutor {
...
+        }
+
+        // TODO: do we need to validate the wasm binary?
+        // ```rust
+        //   let bytes = std::fs::read(path).unwrap();
+        //   wasmparser::validate(&bytes).is_ok()
+        // ```
+
+        path.extension()
+            .map(|ext| ext == "wasm" || ext == "wat")
```
jprendes, line 69: **thought:** This adds a new requirement on the images. Before the extension didn't matter.

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -243,13 +244,35 @@ impl Wasi {
...
 
+        let wasmtime_executor = Box::new(WasmtimeExecutor {
+            engine,
+            stdin,
+            stdout,
+            stderr,
+        });
+        let default_executor = Box::<DefaultExecutor>::default();
+
+        if let Some(stdin) = stdin {
```
jprendes, line 255: **question:** Are these needed just for testing?

Makefile
```diff
@@ -67,10 +67,16 @@ bin/kind: test/k8s/Dockerfile
 test/k8s/_out/img: test/k8s/Dockerfile Cargo.toml Cargo.lock $(shell find . -type f -name '*.rs')
 	mkdir -p $(@D) && $(DOCKER_BUILD) -f test/k8s/Dockerfile --iidfile=$(@) --load  .
 
+.PHONY: test/py-flask
+test/py-flask:
+	$(DOCKER_BUILD) -t py-flask-app:latest -f $@/Dockerfile --load $@
```
jprendes, line 72: **suggestion:** Do we need a custom image? can't we use something like the `nginx` or `hello-world` images?

test/py-flask/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+    pip3 install -r requirements.txt
+
+COPY . /app
+
+ENTRYPOINT ["python3"]
+CMD ["app.py"]
+
+FROM builder as dev-envs
+
+RUN apk update && \
```
jprendes, line 17: **question:** is this needed?

arschles (commented):
just one suggestion. otherwise LGTM

crates/containerd-shim-wasmtime/src/executor.rs
Mossaka: That's right, maybe we can probe the binary and check if the binary file is a wasm file.

Makefile
Mossaka: That's a good suggestion!

test/py-flask/Dockerfile
Mossaka: No longer if we don't have custom image

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -243,13 +244,35 @@ impl Wasi {
...
+                STDOUT_FD = Some(dup(STDOUT_FILENO));
+                dup2(stdout, STDOUT_FILENO);
+            }
+        }
+        if let Some(stderr) = stderr {
+            unsafe {
+                STDERR_FD = Some(dup(STDERR_FILENO));
+                dup2(stderr, STDERR_FILENO);
+            }
+        }
```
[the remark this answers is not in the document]
Mossaka: I don't believe they are the same logic - the old code will run `dup2` for individual descriptor, and skil if that descriptor is `None`. The new code skips completely if one of the descriptor is None, no?

crates/containerd-shim-wasmtime/src/instance.rs
ipuustin: Yeah, I think we could have a list of heuristics, going from explicit to implicit and quick to slow:
  1. Check the annotation -- if the user says this is a Wasm binary, believe them.
  2. Check the file extension -- if it's `.wasm`, consider it a Wasm binary.
  3. Check the ELF headers
  4. Maybe check Wasm binary format?

Note that possibly that in the (far) future we need to have information about which WASI interfaces are used by the module, so that we can use the right Wasm runtime with the right settings. It could be that we need to standardize some annotation for that purpose and we can use that for the binary check too. But that's not in the scope for this PR. :-)

jprendes: I am happy to keep this as is. We can evolve it in later PRs.
I also imagine this will change a fair bit when we start using OCI artifacts, and we can revisit this then.

Mossaka:
Sorry for getting back to your comments late. I am going to resolve them either tonight or Monday morning. Thanks for your reviews!

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: Yeah I prefer to keep it as is and I'd like to raise an issue for further discussion on how we can improve this. I'd like to get @jsturtevant thoughts on this as well.

jsturtevant: > That could work assuming we have the data... like in an annotation or something?

this could be done via an additional annotation short term in https://github.com/containerd/containerd/pull/8699. Longer term I could see a section in the [runtimeconfig](https://github.com/opencontainers/runtime-spec/blob/main/spec.md#platforms) for WASI (like it is for Windows/linux)

> I'd like to raise an issue for further discussion on how we can improve this

This sounds reasonable.  This is a good first step in that direction to get folks to try it out.

crates/containerd-shim-wasmtime/src/executor.rs
Mossaka: This does not handle `PATH`.

crates/containerd-shim-wasmtime/src/executor.rs
Mossaka: I am going to keep it as is and can keep iterating through it in future PRs.

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: These are needed by the default container runtime to redirect stdio. I will update the code with comments.

Makefile
Mossaka: Done.

test/py-flask/Dockerfile
Mossaka: Removed

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -39,8 +39,35 @@ impl Executor for WasmtimeExecutor {
...
+        if let Some(strpd) = stripped {
+            cmd = strpd.to_string();
+        }
+
+        let mut path = PathBuf::from(cmd);
+        if path.is_relative() {
+            path = std::env::current_dir().unwrap().join(path);
+        }
+
+        // TODO: do we need to validate the wasm binary?
```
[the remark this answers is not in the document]
Mossaka: That's a good suggestion. I'd leave it as is for now and defer the validation to future RPs.

Mossaka:
I tried to run container and wasm container in a pod, but the hello-world container crashed. I checked containerd log messages and I got

```
Aug 01 00:12:55 containerd-wasm-control-plane containerd[108]: time="2023-08-01T00:12:55.171882679Z" level=error msg="Failed to pipe stdout of container \"b0060fd6583b82000e6062e62f13a38a2d47d910fde1e338bf4f84d17c5107a0\"" error="read /proc/self/fd/146: file already closed"
```

It looks like there are some issues with how youki's `DefaultExecutor` pipe std I/O to containerd?

Mossaka:
Even though I got the above error and the pod was not in READY state in k8s, I could still see pod logs

```
➜  runwasi git:(issue64) ✗ k logs wasi-demo-7cb6778b8f-gpd6p

Hello from Docker!
This message shows that your installation appears to be working correctly.
...
```

Mossaka:
I've updated this PR to first use Default Executor and check the the executable if it's a linux binary. Then it falls to the wasm executor.

Mossaka:
Okay, to fully verify that a path is executable in Linux, I need to verify a few things:
1. Check if the path containers "/", if is, then regard it as a full path. Otherwise, iterate through $PATH and prefix it to path. 
2. Check if the path exists
3. Check if the path executable permission is set to 1
4. Check if the path contains shebang (#!), if it is, then it's a script.
5. Check if the path contains ELF magic number, if it is, then it's a Linux ELF file.
6. If both no, then path is not an executable.

Mossaka:
Wasm binary also has magic number (`00 61 73 6D`), but the text format of the Wasm modules don't. For this PR, I will leave the verification logic for Wasm to be file extensions like `.wasm` and `.wat`.

Mossaka:
🥳 integration tests passed!!

test/k8s/deploy.yaml
```diff
@@ -24,4 +24,8 @@ spec:
       containers:
       - name: demo
         image: ghcr.io/containerd/runwasi/wasi-demo-app:latest
-        imagePullPolicy: Never
\ No newline at end of file
+        imagePullPolicy: Never
+      - name: nginx
```
jsturtevant, line 28: does this work on k3s?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -240,16 +221,23 @@ impl Wasi {
...
+            stderr,
+        });
         let container = ContainerBuilder::new(self.id.clone(), syscall.as_ref())
-            .with_executor(vec![Box::new(WasmtimeExecutor {
-                stdin,
-                stdout,
-                stderr,
-                engine,
-            })])?
+            .with_executor(vec![default_executor, wasmtime_executor])?
```
jsturtevant, line 236: hmm If we end up making the binary check optional on the builder (https://github.com/containers/youki/issues/2220) then it would mean the check wouldn't happen for linux containers as well.

crates/containerd-shim-wasmtime/src/executors/wasi.rs
```diff
@@ -39,8 +39,35 @@ impl Executor for WasmtimeExecutor {
...
+            return false;
+        }
+
+        let start = args[0].clone();
+        let mut iterator = start.split('#');
+        let mut cmd = iterator.next().unwrap().to_string();
+        let stripped = cmd.strip_prefix(std::path::MAIN_SEPARATOR);
+        if let Some(strpd) = stripped {
+            cmd = strpd.to_string();
+        }
```
jsturtevant, line 55: can synchronize this with https://github.com/containerd/runwasi/pull/219 to centralize at some point

crates/containerd-shim-wasmtime/src/executors/container.rs
```diff
@@ -0,0 +1,175 @@
...
+const EXECUTOR_NAME: &str = "default";
+
+#[derive(Default)]
+pub struct LinuxContainerExecutor {
+    pub stdin: Option<RawFd>,
+    pub stdout: Option<RawFd>,
+    pub stderr: Option<RawFd>,
+}
+
+impl Executor for LinuxContainerExecutor {
```
jsturtevant, line 21: Is there a reason we can't use the default executor or wrap it?

crates/containerd-shim-wasmtime/src/instance.rs
jsturtevant: Wondering if the `can_handle` function on the executor is the correct place for that check? It looks like we are mostly checking the spec, to determine which executor to choose.  Another option might be to check and build the container with the right options?  that would solve "check for binary" problem.

crates/containerd-shim-wasmtime/src/executors/wasi.rs
```diff
@@ -39,8 +39,35 @@ impl Executor for WasmtimeExecutor {
...
+
+        let mut path = PathBuf::from(cmd);
+        if path.is_relative() {
+            path = std::env::current_dir().unwrap().join(path);
+        }
```
jprendes, line 60: **question:**
Why do we need to strip the `/`?
Doesn't that make the `is_relative` check below be always true?
And we don't need any of that if we are just going to check the extension.

crates/containerd-shim-wasmtime/src/executors/wasi.rs
```diff
@@ -39,8 +39,35 @@ impl Executor for WasmtimeExecutor {
...
+
+        // TODO: do we need to validate the wasm binary?
+        // ```rust
+        //   let bytes = std::fs::read(path).unwrap();
+        //   wasmparser::validate(&bytes).is_ok()
+        // ```
+
+        path.extension()
+            .map(|ext| ext == "wasm" || ext == "wat")
+            .unwrap_or(false)
```
jprendes, line 70: **nit:**
```suggestion
        path.extension()
            .is_some_and(|ext| ext == "wasm" || ext == "wat")
```

crates/containerd-shim-wasmtime/src/executors/wasi.rs
```diff
@@ -39,8 +39,35 @@ impl Executor for WasmtimeExecutor {
...
+            path = std::env::current_dir().unwrap().join(path);
+        }
+
+        // TODO: do we need to validate the wasm binary?
+        // ```rust
+        //   let bytes = std::fs::read(path).unwrap();
+        //   wasmparser::validate(&bytes).is_ok()
+        // ```
+
+        path.extension()
```
jprendes, line 68: **question:** Do we care about being case sensitive?
```suggestion
        path.extension()
            .map(|ext| ext.to_lowercase())
```

crates/containerd-shim-wasmtime/src/executors/container.rs
```diff
@@ -0,0 +1,175 @@
+use libcontainer::workload::{Executor, ExecutorError, EMPTY};
```
jprendes, line 1: **thought:** I believe this would better suited in `containerd-shim-wasm` since there's nothing `wasmtime` specific here. That way we can reuse it for `wasmedge`.

jprendes (approved):
Thanks for working on this!
I left a few non-blocking comments.
Feel free to merge as is (I understand you want to merge soon). We can tidy up later :-)

test/k8s/deploy.yaml
Mossaka: `make test/k3s` uses WasmEdge shim only, which I haven't implemented DefaultExecutor to yet. So the answer is no, it doesn't work, until we implement the same thing for wasmedge shim.

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: > Wondering if the can_handle function on the executor is the correct place for that check? 

I am wondering that too. At least my implementation of `can_handle` in the default executor is very similar to youki's verify_binary except that it also verifies if the executable is a Linux ELF file or a script file.

crates/containerd-shim-wasmtime/src/executors/wasi.rs
Mossaka: Sounds good.

crates/containerd-shim-wasmtime/src/executors/container.rs
Mossaka: Mostly because I want to implement `can_handle` for the `LinuxContainerExecutor`. The default executor just returns `true` for `can_handle` https://github.com/containers/youki/blob/4dc30bf7c02be584eddfc182d4e96b7cbbb3474c/crates/libcontainer/src/workload/default.rs#L46-L48

crates/containerd-shim-wasmtime/src/executors/container.rs
```diff
@@ -0,0 +1,175 @@
...
+        Ok(_) => {}
+        Err(err) => {
+            log::info!("failed to read shebang of {}: {}", executable, err);
+            return false;
+        }
+    }
+    buffer == [0x23, 0x21] // #!
+}
+
+fn check_elf(path: PathBuf, executable: &str) -> bool {
```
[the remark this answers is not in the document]
Mossaka: Done

crates/containerd-shim-wasmtime/src/executors/container.rs
```diff
@@ -0,0 +1,175 @@
...
+        let metadata = metadata.unwrap();
+        let permissions = metadata.permissions();
+        if !metadata.is_file() || permissions.mode() & 0o001 == 0 {
+            log::info!("{} is not a file or has no execute permission", executable);
+            return false;
+        }
+
+        // check the shebang and ELF magic number
+        // https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
+        if !check_shebang(path.clone(), executable) && !check_elf(path.clone(), executable) {
```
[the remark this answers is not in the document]
Mossaka: Done

crates/containerd-shim-wasmtime/src/executors/container.rs
```diff
@@ -0,0 +1,175 @@
...
+        return false;
+    }
+    let mut file = file.unwrap();
+    match file.read_exact(&mut buffer) {
+        Ok(_) => {}
+        Err(err) => {
+            log::info!("failed to read shebang of {}: {}", executable, err);
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
Mossaka: Done thanks for the suggestion!

crates/containerd-shim-wasmtime/src/executors/container.rs
```diff
@@ -0,0 +1,175 @@
...
+            for p in path.split(':') {
+                let path = PathBuf::from(p).join(executable);
+                if path.exists() {
+                    found = true;
+                    found_path = path;
+                    break;
+                }
+            }
+            if !found {
+                return false;
```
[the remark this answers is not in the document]
Mossaka: Nice finding! Could you please fire an issue against the youki's repo. I will raise an issue here in runwasi and link it.

crates/containerd-shim-wasmtime/src/executors/container.rs
```diff
@@ -0,0 +1,175 @@
...
+        // payload through execvp, so it should never reach here.
+        unreachable!();
+    }
+
+    fn can_handle(&self, spec: &Spec) -> bool {
+        let args = spec
+            .process()
+            .as_ref()
+            .and_then(|p| p.args().as_ref())
+            .unwrap_or(&EMPTY);
```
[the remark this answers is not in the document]
Mossaka: Done.

crates/containerd-shim-wasmtime/src/executors/wasi.rs
Mossaka: You are right, I implemented it as I were passing `path` to wasmparser to validate the wasm bytes. I can remove them.

crates/containerd-shim-wasmtime/src/executors/container.rs
Mossaka: You are 100% right about it. Once #226 is merged, I will move these to `containerd-shim-wasm`, until then, I don't want to introduce the deps on `libcontainer` yet.

crates/containerd-shim-wasmtime/src/executors/container.rs
jsturtevant: why not wrap it so we aren't duplicating? I guess I mostly concerned about fixes going into youki like https://github.com/containerd/runwasi/pull/156#discussion_r1284160469 and not getting those in here.  It seems youki should be the place for the main implementation of executing containers not runwasi.

if all we want different is the can_run then again I am wondering if this is the right place for that type of desicion (https://github.com/containerd/runwasi/pull/156#discussion_r1284842502)

Mossaka: Changes that go into `verify_binary` should happen at a different layer as where `Executor`'s `can_handle` is invoked. 

> It seems youki should be the place for the main implementation of executing containers not runwasi.

I agree, I can wrap a `DefaultExecutor` instead if the `LinuxContainerExecutor` to reduce duplicate code.

Mossaka: > I am wondering if this is the right place for that type of desicion (https://github.com/containerd/runwasi/pull/156#discussion_r1284842502)

I think this totally makes sense, but it can be deferred because it's not a major concern of this PR.

Mossaka: Just push a new commit that wraps `DefaultExecutor` instead the `LinuxContainerExecutor`.

crates/containerd-shim-wasmtime/src/executors/container.rs
```diff
@@ -1,60 +1,19 @@
...
-const EXECUTOR_NAME: &str = "default";
+use std::io::Read;
+use std::{fs::OpenOptions, path::PathBuf};
 
 #[derive(Default)]
 pub struct LinuxContainerExecutor {
-    pub stdin: Option<RawFd>,
-    pub stdout: Option<RawFd>,
-    pub stderr: Option<RawFd>,
+    pub default_executor: DefaultExecutor,
```
jsturtevant, line 11: I think this could be private since we are handling it? then the shim wouldn't need to know about the default at all?

crates/containerd-shim-wasmtime/src/executors/container.rs
jsturtevant: > > I am wondering if this is the right place for that type of desicion ([#156 (comment)](https://github.com/containerd/runwasi/pull/156#discussion_r1284842502))
> 
> I think this totally makes sense, but it can be deferred because it's not a major concern of this PR.

I think it is though.  the decision making is the core part of the PR and has direct impact on https://github.com/containers/youki/issues/2220

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: If we make binary check optional on the builder, then the `can_handle` function will be used for checking the binary for linux containers. I think that's the right place for this check. 

It looks like the `Executor` trait and its `can_handle` function were removed from the main branch of `youki`.

crates/containerd-shim-wasmtime/src/executors/container.rs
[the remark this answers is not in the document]
cpuguy83: I'm a bit concerned about the path handling here.
We need to make sure things like symlinks aren't traversing the host filesystem.
I also found rust's path joining really inadequate to using with potentially hostile values.

crates/containerd-shim-wasmtime/src/instance.rs
jsturtevant: > It looks like the Executor trait and its can_handle function were removed from the main branch of youki

😮 interesting, does this change anything?

crates/containerd-shim-wasmtime/src/executors/container.rs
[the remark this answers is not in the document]
Mossaka: Do you want to address the path handling in this PR, or okay with raising an issue for future iterations? @cpuguy83

cpuguy83:
Looking through youki a bit, I think `can_handle` is not part of the executor interface and instead it is expected that we call `can_handle` from within `exec()` and return an appropriate error if it can't be used.

0xE282B0:
@cpuguy83, yes that was refactored lately. There is https://github.com/containerd/runwasi/issues/209 to adapt to the new interface.

Mossaka:
>  I think can_handle is not part of the executor interface and instead it is expected that we call can_handle from within exec() and return an appropriate error if it can't be used.

Yeah, as @0xE282B0 mentioned, this was the latest breaking change post `v0.1.0` of libcontainer and we are still depending on the released version so we will deal with the refactoring later at #209.

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: I assume there needs some work to refactor to the latest change of youki. #209

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: So this PR made a decision to 
1. check Linux binary
2. check wasm binary

In checking Linux binary, it performs
1. arguments should not be empty
2. find binary from PATH or the path of the binary contains "/" (this follows youki's `verify_binary` impl)
3. check execute permission and existance of the file
4. check the shebang and ELF magic number

In checking wasm binary, it performs
1. check entrypoint path and truncate "#"
2. check extension is "wasm" or "wat"

crates/containerd-shim-wasmtime/src/instance.rs
utam0k: Sorry for confusing you. Don't worry.
https://github.com/containers/youki/issues/2220#issuecomment-1676356315

Mossaka:
I've updated this PR to add linux container executors to both Wasmtime and WasmEdge shims.

test/k8s/deploy.yaml
Mossaka: I have updated this PR to include nginx on k3s.

crates/containerd-shim-wasmtime/src/executors/container.rs
Mossaka: I like how the direction is going in https://github.com/containers/youki/issues/2220

crates/containerd-shim-wasmtime/src/executors/container.rs
Mossaka: Done.

cpuguy83 (approved):

Mossaka:
Hey thanks everyone for reviewing this PR! I realized it's very difficult to squash commits before merge so I did a squash merge all together to make main branch clean. We can set a policy to only allow squash merge, wdyt? @jsturtevant

crates/containerd-shim-wasmtime/src/executors/container.rs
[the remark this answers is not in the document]
utam0k: @cpuguy83 @Mossaka FYI: I have addressed this issue you mentioned: https://github.com/containers/youki/pull/1911/files

PR #176, opened 2023-07-12
fix: update Makefile build and check

### LINKED ISSUE #169: make build encountered an error
I tried to use the `make build` command and got the following error:

image

I think this error may have occurred when installing wasmedge. 

I tried to use the official command: `curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash -s -- -e all -v 0.12.1
`, and the `make build` was executed successfully.

Now it looks like make bin/wasmedge doesn't install the required dependencies.

dierbei:
FIx #169


Mossaka (commented):
Mostly LGTM!

I wonder if these two env vars can be set in the wasmedge crate in `build.rs` file. This allows people to not worry about the env vars when only building wasmtime crate.

jprendes:
> I wonder if these two env vars can be set in the wasmedge crate in build.rs file.

Unfortunately that's not posible. The env-vars are needed for the `wasmedge-sys`'s `build.rs`, which normally even runs before `containerd-shim-wasmedge`'s `build.rs`.

Another option is to use [the `[env]` section in cargo config](https://doc.rust-lang.org/cargo/reference/config.html#env). However I'm not sure if individual crates in a workspace can have different cargo configs.

These env vars are also required in the `fix` target for the same reason as `check`. To avoid repetition I would prefer these as [exported variables](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html) towards the top of the Makefile

Mossaka:
> To avoid repetition I would prefer these as [exported variables](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html) towards the top of the Makefile

I agree!

Mossaka:
Another question: for contributions who wish to develop `runwasi` locally, should we recommend them to install WasmEdge in local? That is, we will default WasmEdge Rust SDK to search for the expected libraries in `/usr/local`.

jprendes:
WasmEdge will soon release static libraries https://github.com/WasmEdge/WasmEdge/issues/2660 . I would like to move the shim to use that instead.

jprendes:
I've also been tinkering with "vendoring" libwasmedge.so in #175

dierbei:
@jprendes If I understand correctly.

I should need to add `export WASMEDGE_INCLUDE_DIR=$(PWD)/bin/wasmedge/include` at the top of the Makefile.

Similar to the snippet below：
https://github.com/containerd/runwasi/blob/4362de0e81cafc0e5c793eab353787a93cb9e3df/Makefile#L5-L6

This will have a default value for wasmedge.

And we can use `make build WASMEDGE_INCLUDE_DIR=/my-wasmedge/path` to override the default.

jprendes:
Yes, pretty much. But you need to use `?=` rather than just `=` to allow overriding, just like it's done with `CONTAINERD_NAMESPACE`

jprendes (approved):
LGTM

dierbei:
@jprendes Changes have been made, please review.

Mossaka (approved):
LGTM!

PR #177, opened 2023-07-12
Do not run clippy on auto-generated files

jprendes:
The files generated by the `ttrpc-compiler` do not pass some stylistic clippy's checks.
This means that we need to run `make fix` after running `make build` to prevent CI failures.

This PR prevents clippy from running on the auto-generated.
This means that `make check` won't fail on those files.
This also prevents `make fix` from modifying the auto-generated files.


Mossaka (approved):
LGTM!

rumpl (approved):

PR #185, opened 2023-07-19
Use rust-toolchain.toml and Bump rust version to 1.71

utam0k:


Dockerfile
```diff
@@ -1,6 +1,8 @@
 # syntax=docker/dockerfile:1
 
-ARG RUST_VERSION=1.69
+# Make sure to keep in sync with the version in rust-toolchain.toml"
```
rumpl, line 3: I have one last smallest of nits, sorry
```suggestion
# Make sure to keep in sync with the version in rust-toolchain.toml
```

rumpl (approved):
:+1:

Dockerfile
utam0k: 👍 fixed

PR #186, opened 2023-07-19
fix: add make deps to release CI

Mossaka:


cpuguy83 (approved):

cpuguy83 (changes requested):
Actually, beyond CI failing, I'm not a fan of this since its encoding ubuntu/debian specific things into the makefile.
It'd be fine to include this in script somewhere, though.

Mossaka:
okay all of these CI failures were due to a cache. After I deleted the cache, the CI succeed. I am going to rollback to the previous commit.

cpuguy83 (approved):
LGTM

PR #187, opened 2023-07-19
Link against wasmedge static library

jprendes:
This PR links the wasmedge shims agains the wasmedge static library:
* This simplifies distribution of the shims.

It uses the (new) combination of `standalone` and `static` features of the `wasmedge-sys` crate, which automatically downloads the static library as part of the build process:
* This removes the need to manually install wasmedge.
* Simplifies dependency management: the wasmedge version is now set by the upstream `wasmedge-sdk` crate.
* This requires switching to a git commit for `wasmedge-sdk` until tag `0.10.0` is released.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,22 +9,14 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
 
 FROM --platform=$BUILDPLATFORM rust:${RUST_VERSION} AS base
 COPY --from=xx / /
-RUN apt-get update -y && apt-get install --no-install-recommends -y clang jq
+RUN apt-get update -y && apt-get install --no-install-recommends -y clang jq wget
 
 FROM base AS build
 SHELL ["/bin/bash", "-c"]
 ARG BUILD_TAGS TARGETPLATFORM
-ENV WASMEDGE_INCLUDE_DIR=/root/.wasmedge/include
-ENV WASMEDGE_LIB_DIR=/root/.wasmedge/lib
-ENV LD_LIBRARY_PATH=/root/.wasmedge/lib
 RUN xx-apt-get install -y gcc g++ libc++6-dev zlib1g
 RUN xx-apt-get install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev
 RUN rustup target add $(xx-info march)-unknown-$(xx-info os)-$(xx-info libc)
-RUN <<EOT
-    set -ex
-    os=$(xx-info os)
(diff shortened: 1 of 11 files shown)


rumpl (approved):
Beautiful

CaptainVincent:
Great work! This greatly improves the user's installation experience,
and everything in my tests turned out fine, too. 👍

Dockerfile
```diff
@@ -9,22 +9,14 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
 
 FROM --platform=$BUILDPLATFORM rust:${RUST_VERSION} AS base
 COPY --from=xx / /
-RUN apt-get update -y && apt-get install --no-install-recommends -y clang jq
+RUN apt-get update -y && apt-get install --no-install-recommends -y clang jq wget
```
cpuguy83, line 12: Doesn't seem like we should need wget?

cpuguy83 (changes requested):

Dockerfile
jprendes: `wasmedge-sys` needs `wget` to download the library.
We could fix that upstream for the next release of `wasmedge-sdk` and remove it then.
https://github.com/WasmEdge/wasmedge-rust-sdk/issues/30

cpuguy83 (approved):

cpuguy83:
This looks great, if there's a change that's needed for the containerd config.toml stuff we can follow-up.

ipuustin:
I'm bit on the fence with this one... I really like that we now automatically have the right version of the WasmEdge library. However I think downloading things with custom methods during the build has some issues, such as the possibility of having difficult-to-debug issues when building in a proxy prison or other not-fully-networked environments. It also makes it more difficult to have SBOMs and reproducible builds, because we have no way of knowing what exactly has been downloaded (for example -- is a WasmEdge library security fix part of a built shim or not?).

An optimal case might be to have an option to pre-download the library, or at least specify which library checksum we are expecting during the build?

jprendes:
@ipuustin , would you mind creating an issue in the [wasmedge-rust-stk](https://github.com/WasmEdge/wasmedge-rust-sdk/issues) repo to add this functionality?

I was looking [here](https://github.com/diwic/dbus-rs/blob/master/libdbus-sys/build_vendored.rs) at how the `dbus` crate does it. But I'm not sure that's directly applicable to this case.

cpuguy83:
Apologies I clearly made a mistake here with regard to automatic downloading stuff.
Ideally this would not download anything at all and use the system libs.
The lib should be a build-time dep that needs to be handled by the builder not our makefile, build.rs, or whatever from dependencies.

We can certainly create a script to install deps that someone who's building can choose to run or not.

Also probably static compilation should be behind a crate feature.

PR #188, opened 2023-07-19
feat: auto merge dependabot PRs for patch updates

### LINKED ISSUE #150: Patch version updates automated by Dependabot
How about making patch version updates automatically merged when CI passes?

Mossaka:
This PR adds a new action called "Dependabot automation" which will automatically merge dependabot PRs for patch updates. According to this [doc](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#enable-auto-merge-on-a-pull-request), 

> If you enable auto-merge for a pull request, the pull request will merge automatically when all required reviews are met and all required status checks have passed.

which means the PRs won't be auto-merged if the tests gate fails or it doesn't have the required number of reviews. 

Hence I've also added `gh pr review --approve "$PR_URL` to automatically approve the dependabot PRs. 

Since this automates the process of merging dependabot PRs, I've set the schedule interval for depdnabot to "daily". 

Close #150 

Could you please take a look? @utam0k

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    open-pull-requests-limit: 5
     ignore:
       - dependency-name: "wasmtime"
       - dependency-name: "wasi-common"
--- a/.github/workflows/dependabot_auto_merge.yml
+++ b/.github/workflows/dependabot_auto_merge.yml
@@ -1,0 +1,36 @@
+# This action will automatically merge dependabot PRs on patch updates
+#
+# See more: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#enable-auto-merge-on-a-pull-request
+name: Dependabot automation
+
+on:
+  pull_request:
+    types:
+      - opened
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  dependabot-automation:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Enable auto-merge for Dependabot PRs on patch updates
+        # Only enable auto-merge for patch updates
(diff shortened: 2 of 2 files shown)


.github/dependabot.yml
```diff
@@ -3,7 +3,8 @@ updates:
   - package-ecosystem: cargo
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
```
cpuguy83, line 6: I think weekly is too much.
Why do we need daily?

Mossaka: I thought having this action to auto-merge dependabot PRs would grealty reduce the churn and thus shouldn't put more burden on maintainers even if we change the frequency to daily.

Mossaka: I am happy to revert it back to weekly

utam0k: If you do it daily, you will have the added burden of writing release notes. Consideration should be given to that.
Youki prevents this by generating release notes with labels.

jprendes:
> which means the PRs won't be auto-merged if the tests gate fails or it doesn't have the required number of reviews.

Does this mean that any failing CI test would prevent the auto merge?
IIUC, the `main` branch should be protected from failing CI tests. Is this currently the case?

cpuguy83:
> Is this currently the case

Just reviewed branch protection.
It is now.
PR's also require an approver.

Honestly I'm not particularly on board with auto-merging dependabot, especially knowing our CI is missing some important checks (like validating protos).

Mossaka:
It seems like this PR won't be merged in. I will convert it to a draft and leave it there.

cpuguy83:
I still don't like auto-merging dependabot, but I wouldn't want to block it if others find it useful.

Mossaka:
@cpuguy83 would you please state the reasons why you don't like auto-merging dependantabot? This will help me understand the tradeoffs better.

jsturtevant:
I am not a big fan of auto-merging because it doesn't give awareness to maintainers.  I like the manual step reviewing and understanding what changed even if it takes an extra step. Even if tests pass there might be a reason not to merge (either something like missing [tests](https://github.com/containerd/runwasi/pull/113#issuecomment-1548639736) or some other external reason (one example could be a library gets compromised, and we need it pinned to a particular version).  

In summary, I love having dependa-bot do the manually labor but like the human check before merging.

PR #189, opened 2023-07-20
fix: remove redundent cache

Mossaka:
PR #185 changes action from [actions-rs/toolchain@v1](https://github.com/actions-rs/toolchain) to [actions-rust-lang/setup-rust-toolchain@v1](https://github.com/actions-rust-lang/setup-rust-toolchain)

There is a key difference between these two actions and that is cache. [actions-rust-lang/setup-rust-toolchain@v1](https://github.com/actions-rust-lang/setup-rust-toolchain) defaults to use "Swatinem/rust-cache", but we already have used this in the CI. 

So this PR removes the redundent cache steps, and disables the rust-cache action in the release CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,6 @@ jobs:
   fmt:
     runs-on: "ubuntu-latest"
     steps:
-      - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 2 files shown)


jprendes (approved):
LGTM

cpuguy83 (approved):

PR #197, opened 2023-07-22
[CI] Make Ubuntu 20.04 and 22.04 use different rust cache

### LINKED ISSUE #193: CI might be broken after #187
seems like CI might be broken after #187 

```
   Dirty wasmedge-sys v0.15.0: the file `/usr/lib/llvm-14/lib/clang/14.0.0/include/stdbool.h` is missing
   Compiling wasmedge-sys v0.15.0
     Running `/home/runner/work/runwasi/runwasi/target/debug/build/wasmedge-sys-24c4a5d29523e79f/build-script-build`
error: failed to run custom build command for `wasmedge-sys v0.15.0`
note: To improve backtraces for build dependencies, set the CARGO_PROFILE_DEV_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.

Caused by:
  process didn't exit successfully: `/home/runner/work/runwasi/runwasi/target/debug/build/wasmedge-sys-24c4a5d29523e79f/build-script-build` (exit status: 1)
  --- stderr
  /home/runner/work/runwasi/runwasi/target/debug/build/wasmedge-sys-24c4a5d29523e79f/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /home/runner/work/runwasi/runwasi/target/debug/build/wasmedge-sys-24c4a5d29523e79f/build-script-build)
  /home/runner/work/runwasi/runwasi/target/debug/build/wasmedge-sys-24c4a5d29523e79f/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /home/runner/work/runwasi/runwasi/target/debug/build/wasmedge-sys-24c4a5d29523e79f/build-script-build)
  /home/runner/work/runwasi/runwasi/target/debug/build/wasmedge-sys-24c4a5d29523e79f/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /home/runner/work/runwasi/runwasi/target/debug/build/wasmedge-sys-24c4a5d29523e79f/build-script-build)
warning: build failed, waiting for other jobs to finish...
````

_Originally posted by @jsturtevant in https://github.com/containerd/runwasi/issues/192#issuecomment-1646270928_

### PR DESCRIPTION
**Pull request by jprendes**
Currently Ubuntu 20.04 and 22.04 are sharing the same cache.
This is a problem as 20.04 and 22.04 ship differente glibc versions.
This prevents the buil scripts compiled and cached in 22.04 from running in 20.04.

`actions-rust-lang/setup-rust-toolchain@v1` uses `Swatinem/rust-cache@v2`, which uses all environment variables matching `RUST*` as hash for the cache key. See the description of `env-vars` [here](https://github.com/marketplace/actions/rust-cache).

This PR adds an environment variable `RUST_CACHE_KEY_OS` so that different OS use different caches.

This fixes #193.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,8 @@ jobs:
         run: "mount | grep cgroup"
       - uses: actions/checkout@v3
       - uses: actions-rust-lang/setup-rust-toolchain@v1
+        env:
+          RUST_CACHE_KEY_OS: ${{ matrix.os }}
       - name: Setup OCI runtime build env
         run: ${GITHUB_WORKSPACE}/.github/scripts/build.sh
         shell: bash

### CODE REVIEW

**Comment by jprendes** [comment]
@Mossaka PTAL 🙂

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -32,6 +32,8 @@ jobs:
         run: "mount | grep cgroup"
       - uses: actions/checkout@v3
       - uses: actions-rust-lang/setup-rust-toolchain@v1
+        env:
```
Mossaka, line 35: Shouldn't this be `env-vars` according to this [doc](https://github.com/marketplace/actions/rust-cache)

Mossaka (commented):
Wow nice catch! I was wondering why a cache messed up the glibc library on the build and this is exactly why!! 🤯

Just made one comment on the use of `env`

Mossaka:
I am confused how previous builds didn't hit this problem before #189 removed the cache as we weren't using any env vars to prefix the cache key to a verison of ubuntu before.

Mossaka:
It looks like it's still using the old cache key at this [run](https://github.com/containerd/runwasi/pull/196/checks#step:3:210)

jprendes:
IIUC, by default `Swatinem/rust-cache` does use the os as part of the hash, but `actions-rust-lang/setup-rust-toolchain` specifies a key that doesn't. But I may be misinterpreting the doc.

This was not a problem for #187 because it was bumping the crate with a build script, so it wasn't hitting the cache. #189 did hit the issue.

jprendes:
> It looks like it's still using the old cache key at this run

The job is part of the cache key, so the fmt job doesn't share the cache with the builds. I didn't add any env-var there as it doesn't run in multiple OSes.

Mossaka:
I don't understand how `setup-rust-toolchain` specifies a key that doesn't use the os as part of the hash... Could you show me some clues?

I checked the action source code of `setup-rust-toolchain` and found that it just uses `rust-cache` https://github.com/actions-rust-lang/setup-rust-toolchain/blob/main/action.yml#L159-L161

jprendes:
I'm not sure how `setup-rust-toolchain` sets the cache key for `rust-cache`. I believe that's done [here](https://github.com/actions-rust-lang/setup-rust-toolchain/blob/f3c84ee10bf5a86e7a5d607d487bf17d57670965/action.yml#L144), and uses just `rustc`'s commit and build date. But that way of setting the key is not documented in `rust-cache` as far as I can tell.

`setup-rust-toolchain` doesn't provide options to configure `rust-cache` other than an on/off switch. But it does leaves most options with their default. In particular, the `env-vars` option is a list of `env-var` **prefixes** that will be used for the hash. One of those prefixes is `RUST` and that's how `RUST_CACHE_KEY_OS` ends up being used for the hash.

Maybe another cleaner approach would be to disable cache in `setup-rust-toolchain` and configure `rust-cache` manually like you did in `release.yml`

Mossaka:
Thanks for the explaination! That makes a lot of sense. 

> The env-vars are matched by *prefix*, so the default `RUST` var will
> match all of `RUSTC`, `RUSTUP_*`, `RUSTFLAGS`, `RUSTDOC_*`, etc.
  

This is the comment that explains it, which I missed! Thanks again. 

I can merge it right now if you want

Mossaka (approved):

jsturtevant:
@jprendes  good catch and thanks for helping track that one down 💯

PR #211, opened 2023-07-29
fix(release): fix the Dockerfile

Mossaka:
This commit fixes the broken Dockerfile that is used in release CI. It adds wasmedge installation step in the build, uses `xx-cargo` instead of cargo, and copies the `sed` command from Docker's build script that allows xx-cargo to be used with pkg-config in build scripts.

Reference
1. Docker's [build script](https://github.com/rumpl/containerd-wasi-shims/blob/main/Dockerfile)
2. xx [issue](https://github.com/tonistiigi/xx/issues/108)

FYI @cpuguy83 @jprendes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,6 @@
 ARG RUST_VERSION=1.71
 
 ARG XX_VERSION=1.2.1
-ARG WASMEDGE_VERSION=0.13.1
 
 FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
 
@@ -18,7 +17,7 @@ RUN sed -i -E 's/xx-clang --setup-target-triple/XX_VENDOR=\$vendor xx-clang --se
 
 FROM base AS build
 SHELL ["/bin/bash", "-c"]
-ARG BUILD_TAGS TARGETPLATFORM WASMEDGE_VERSION
+ARG BUILD_TAGS TARGETPLATFORM
 RUN xx-apt-get install -y gcc g++ libc++6-dev zlib1g
 RUN xx-apt-get install -y libsystemd-dev libdbus-1-dev libseccomp-dev
 RUN rustup target add $(xx-info march)-unknown-$(xx-info os)-$(xx-info libc)
@@ -30,11 +29,6 @@ COPY --link Cargo.toml ./
 COPY --link Cargo.lock ./
 ARG CRATE=""
 ARG TARGETOS TARGETARCH TARGETVARIANT
-RUN curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash -s -- \
-        --version ${WASMEDGE_VERSION} \
-        --platform $(xx-info os | sed -e 's/\b\(.\)/\u\1/g') \
-        --machine $(xx-info march) \
-        --path /usr/local
 
 RUN --mount=type=cache,target=/usr/local/cargo/git/db \
     --mount=type=cache,target=/usr/local/cargo/registry/cache \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -25,18 +30,25 @@ COPY --link Cargo.toml ./
 COPY --link Cargo.lock ./
 ARG CRATE=""
 ARG TARGETOS TARGETARCH TARGETVARIANT
+RUN curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash -s -- \
```
jprendes, line 33: We don't need to install WasmEdge manually anymore. That's take care automatically during the build process.

Dockerfile
```diff
@@ -25,18 +30,25 @@ COPY --link Cargo.toml ./
...
+        --machine $(xx-info march) \
+        --path /usr/local
+
 RUN --mount=type=cache,target=/usr/local/cargo/git/db \
     --mount=type=cache,target=/usr/local/cargo/registry/cache \
     --mount=type=cache,target=/usr/local/cargo/registry/index \
     --mount=type=cache,target=/build/src/target,id=runwasi-cargo-build-cache-${TARGETOS}-${TARGETARCH}${TARGETVARIANT} <<EOT
     set -e
     export "CARGO_NET_GIT_FETCH_WITH_CLI=true"
+    export RUSTFLAGS='-Clink-arg=-Wl,-rpath,$ORIGIN'
```
jprendes, line 45: We are by default statically linking against wasmedge now. You can remove this.

jprendes (changes requested):
Thanks for working on this!
Just a few comments.

cpuguy83 (dismissed):

cpuguy83:
Looks like just one comment from @jprendes but otherwise LGTM.
Thanks all.

Dockerfile
Mossaka: done.

Dockerfile
Mossaka: done.

jprendes (approved):
LGTM.
Thanks again for fixing the Dockerfile!

**nit:** [line 56](https://github.com/containerd/runwasi/pull/211/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R56) replace all the `xx-info ...` with a single `xx-cargo --print-target-triple` to compute the output directory:
```bash
cp target/$(xx-cargo --print-target-triple)/release/${bin} /build/bin/${bin}
```
I wished cargo's [`--out-dir`](https://github.com/rust-lang/cargo/issues/6790) was stable.

Dockerfile
```diff
@@ -36,7 +41,7 @@ RUN --mount=type=cache,target=/usr/local/cargo/git/db \
     if [ -n "${CRATE}" ]; then
         package="--package=${CRATE}"
     fi
-    cargo build --release --target=$(xx-info march)-unknown-$(xx-info os)-$(xx-info libc) ${package}
+    xx-cargo build --release ${package}
```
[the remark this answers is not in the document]
Mossaka: Done,

Mossaka:
> nit: [line 56](https://github.com/containerd/runwasi/pull/211/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R56) replace all the xx-info ... with a single xx-cargo --print-target-triple to compute the output directory:

Done.

Thank you @jprendes

jprendes (approved):
LGTM.
Thanks @Mossaka !

cpuguy83 (approved):

PR #215, opened 2023-07-31
fix: add a stat condition check for uploading release binaries

Mossaka:
This commits adds a check to confirm the existance of files in the release subdirectories before attempting to upload to GitHub release. This change helps avoid unnecessary errors in the case where the release subdirectories are empty.

An instance of the failed run this commits tries to resolve is found [here](https://github.com/containerd/runwasi/actions/runs/5718307257).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,9 +99,15 @@ jobs:
       - name: Create release
         run: |
           gh release create ${{ github.ref }} --generate-notes --prerelease
-          for i in release/*/*; do
-            gh release upload ${RELEASE_NAME} $i
-          done
+          # skip upload if there are no files
+          if stat release/*/* >/dev/null 2>&1; then
+            for i in release/*/*; do
+              gh release upload ${RELEASE_NAME} $i
+            done
+          else
+            echo "No files to upload"
+            exit 0
+          fi
         env:
           GH_TOKEN: ${{ github.token }}
           RELEASE_NAME: ${{ needs.generate.outputs.crate }}/${{ needs.generate.outputs.version }}


cpuguy83 (approved):

jprendes:
Did this still created a source only release?

.github/workflows/release.yml
```diff
@@ -99,9 +99,15 @@ jobs:
...
-            gh release upload ${RELEASE_NAME} $i
-          done
+          # skip upload if there are no files
+          if stat release/*/* >/dev/null 2>&1; then
+            for i in release/*/*; do
+              gh release upload ${RELEASE_NAME} $i
+            done
+          else
+            echo "No files to upload"
+            exit 0
```
jprendes, line 109: Shouldn't this be `exit 1`?

Mossaka:
Yes, for `containerd-shim-wasm`. I also triggered release for `containerd-shim-wasmtime` and `containerd-shim-wasmedge` and they are expected to have binaries.

.github/workflows/release.yml
cpuguy83: No we have crates that don't have binaries.
Each tag and release is per crate.

.github/workflows/release.yml
```diff
@@ -99,9 +99,15 @@ jobs:
       - name: Create release
         run: |
           gh release create ${{ github.ref }} --generate-notes --prerelease
-          for i in release/*/*; do
-            gh release upload ${RELEASE_NAME} $i
-          done
+          # skip upload if there are no files
+          if stat release/*/* >/dev/null 2>&1; then
+            for i in release/*/*; do
+              gh release upload ${RELEASE_NAME} $i
```
ipuustin, line 105: Should have quotes here around the variables:

    gh release upload "${RELEASE_NAME}" "$i"

jprendes: It would be ideal factor this out as a script and have [shellcheck](https://www.shellcheck.net/) run on all our scripts.

Issue #194, opened 2023-07-21
Ability to call specific modules in wasmtime no longer works


jsturtevant:
After switching to youki in #131, the ability to pass module names via the entry point no longer works. See a repo in https://github.com/containerd/runwasi/compare/main...jsturtevant:runwasi:specific-modules?expand=1

```
running 1 test
[2023-07-21T23:24:52Z ERROR libcontainer::process::container_init_process] executable ./hello.wat#foo for container process not found in PATH
[2023-07-21T23:24:52Z ERROR libcontainer::process::container_intermediate_process] failed to initialize container process: invalid executable: ./hello.wat#foo
```

Do we want to re-enable this?  It is due to youki doing runtime checks when starting the container: https://github.com/containers/youki/blob/c3559e4a0aff56d63bcda80f35e7cf0b58ccacbe/crates/libcontainer/src/process/container_init_process.rs#L106-L117

jsturtevant:
pr that originally introduce this: https://github.com/containerd/runwasi/pull/102/files

Mossaka:
I went to the [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/b74f37875da0de3a68497790c0c74a996c8a6211/config.md#L157) to checkout the specification of the entrypoints and it says they are

```Entrypoint: array of strings, OPTIONAL
A list of arguments to use as the command to execute when the container starts. These values act as defaults and may be 
replaced by an entrypoint specified when creating a container.
```

My interpretation is that entrypoints are not strictly defined as executable binaries. In the case with runwasi, they could be functions in a Wasm module like `./hello.wat#foo`. Thus, could we loose the requirement in libcontainer for verifying entrypoints? Specifically, could we make verify_binary optional?

PR #219, opened 2023-08-01
Refactor entrypoint logic to one location

jsturtevant:
This was part of the code that I was working on when I found https://github.com/containerd/runwasi/issues/194.  I saw it in https://github.com/containerd/runwasi/pull/156/files#diff-94b2ab211987ee0bd5d798ac422f3c3e169b69da929d6684874f72c347671bfeR49 as well so thought we could use it in one location.  

There is a test that is marked as `[skip]` included that could be enabled once #194 is resolved or I can remove it for now.

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/oci.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/oci.rs
@@ -35,6 +35,17 @@ pub fn get_args(spec: &Spec) -> &[String] {
     }
 }
 
+// get_module returns the module name and exported function name to be called
+// from the arguments on the runtime spec process field. The first argument will
+// be the module name and the default function name is "_start".
+//
+// If there is a '#' in the argument it will split the string
+// returning the first part as the module name and the second part as the function
+// name.
+//
+// example: "module.wasm#function" will return (Some("module.wasm"), "function")
+//
+// If there are no arguments then it will return (None, "_start")
 pub fn get_module(spec: &Spec) -> (Option<String>, String) {
     let args = get_args(spec);
 
--- a/crates/containerd-shim-wasm/src/sandbox/oci.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/oci.rs
@@ -35,6 +35,25 @@ pub fn get_args(spec: &Spec) -> &[String] {
     }
 }
 
+pub fn get_module(spec: &Spec) -> (Option<String>, String) {
+    let args = get_args(spec);
+
+    if !args.is_empty() {
+        let start = args[0].clone();
+        let mut iterator = start.split('#');
+        let mut cmd = iterator.next().unwrap().to_string();
+
+        let stripped = cmd.strip_prefix(std::path::MAIN_SEPARATOR);
+        if let Some(strpd) = stripped {
+            cmd = strpd.to_string();
+        }
+        let method = iterator.next().unwrap_or("_start");
+        return (Some(cmd), method.to_string());
+    }
+
+    (None, "_start".to_string())
+}
+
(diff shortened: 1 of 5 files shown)


crates/containerd-shim-wasm/src/sandbox/oci.rs
```diff
@@ -160,3 +179,125 @@ pub fn setup_prestart_hooks(hooks: &Option<oci_spec::runtime::Hooks>) -> Result<
     }
     Ok(())
 }
+
+#[cfg(test)]
+mod oci_tests {
```
Mossaka, line 184: Nice to see more tests were added!

crates/containerd-shim-wasm/src/sandbox/oci.rs
```diff
@@ -36,6 +36,25 @@ pub fn get_args(spec: &Spec) -> &[String] {
     }
 }
 
+pub fn get_module(spec: &Spec) -> (Option<String>, String) {
```
Mossaka, line 39: Could you please write docs for the return types of this method?

Mossaka (approved):

crates/containerd-shim-wasm/src/sandbox/oci.rs
jsturtevant: done

jsturtevant:
I think there could be some refactoring here to improve the way the spec is processed (this function is called a few times) but that can be done in a follow up if that works 👍

crates/containerd-shim-wasm/src/sandbox/oci.rs
```diff
@@ -35,6 +35,36 @@ pub fn get_args(spec: &Spec) -> &[String] {
...
+// example: "module.wasm#function" will return (Some("module.wasm"), "function")
+//
+// If there are no arguments then it will return (None, "_start")
+pub fn get_module(spec: &Spec) -> (Option<String>, String) {
+    let args = get_args(spec);
+
+    if !args.is_empty() {
+        let start = args[0].clone();
+        let mut iterator = start.split('#');
+        let mut cmd = iterator.next().unwrap().to_string();
```
ipuustin, line 55: I guess `#_start` is invalid module name?

jsturtevant: There is a little bit of a discussion in https://github.com/containerd/runwasi/pull/102 where this was originally introduced. 

My understanding is that this is really a runwasi specific seperator and could be anything (`::` was suggested at one point).  I scanned through https://github.com/WebAssembly/WASI but couldn't find anything that said `#` is valid or not

ipuustin: I think that if this is a runwasi-specific thing, we could actually write down the semantics in how this should work and especially which entrypoint strings are valid and which not. Then we could validate the input string based on that. The benefit would be that other projects could use the same notation and semantics, and we could document the notation also for the users.

For example, should the following strings work (or not):

1. `foo#bar`
2. `#bar`
3. `foo#bar#bar`
4. `foo`
5. `#foo#bar`
6. `foo#`
7. `foo#bar; -- %`

jsturtevant: this is a good idea, are you ok with opening an issue to track this?  This particular PR isn't introducing anything new, just moving some code around so we don't have the same logic in several different places.

ipuustin: Ok, added the issue.

Mossaka (approved):
LGTM!

PR #220, opened 2023-08-01
Centralize running wasi test

jsturtevant:
When working on #219, I noticed the testing logic is all the same and I was adding functionality in two places.  

This is just an example of an idea to re-use some of it.  I would like to get others thoughts on if this makes sense of not.   I am on the fence... but thought I would share as it similar to https://github.com/containerd/runwasi/pull/111#pullrequestreview-1496058596

If so we could clean it up, move it somewhere else and maybe improve the function `run_wasi_test<I, E>` as it needs instance and the associated item.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -558,6 +558,7 @@ dependencies = [
  "pretty_assertions",
  "serde",
  "serde_json",
+ "serial_test",
  "tempfile",
  "thiserror",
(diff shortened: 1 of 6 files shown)


jsturtevant:
close in favor of #111.  It has similar functionality but isn't in the main shim library which is better long term I think

PR #226, opened 2023-08-03
refactor: adding `LibcontainerInstance` for default implementation of `Instance` trait

### LINKED ISSUE #131: Use youki libcontainer crate for all shims

Mossaka:
This commits adds a new `LibcontainerInstance` trait in the containerd-shim-wasm crate. It refactors the common youki Instance implementation into this trait and as a consequence, all the Instance implementation from both the wasmtime and wasmedge shims moved to a common place.

This requires the shim author to implement the newly introduced `LibcontainerInstance` and then uses its default methods to implement Instance. The reason doing that is there are no changes to the existing `Instance` trait. The downside is that any new shim instance needs to implement both `LibcontainerInstance` and the `Instance` traits. 

This closes #131


crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -142,6 +152,155 @@ pub trait Instance {
...
+    fn wait(&self, waiter: &Wait) -> Result<(), Error> {
+        let id = self.get_id();
+        let exit_code = self.get_exit_code();
+        log::info!("waiting for instance: {}", id);
+        let code = exit_code;
+        waiter.set_up_exit_code_wait(code)
+    }
+
+    fn new(id: String, cfg: Option<&InstanceConfig<Self::E>>) -> Self {
+        Self::new_youki(id, cfg)
```
jsturtevant, line 300: clever!

crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -142,6 +152,155 @@ pub trait Instance {
     fn wait(&self, waiter: &Wait) -> Result<(), Error>;
 }
 
+/// YoukiInstance is a trait that gets implemented by a WASI runtime that
+/// uses youki's libcontainer library as the container runtime.
+/// It provides default implementations for some of the Instance trait methods.
+/// The only method that needs to be implemented is `fn new()`.
```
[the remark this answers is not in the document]
Mossaka: I will update the docs to reflect this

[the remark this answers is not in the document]
Mossaka: Done.

jsturtevant:
I like the way this shares all the work and makes it easier for the implementors and think this is a great direction eventually.

The concern I have is I've gotten back to work on #49 and found that since our move to youki this PR could cause an issue since youki doesn't support Windows (and it appears would need some major work to get it going).  If a shim such as wasmtime implements `YoukiInstance` then we won't be able to add the windows support at all since we take a hard dependecy on `libcontainer` , where as right now the youki work can be hidden behind some compile time flags.

Mossaka:
@jsturtevant I can make `libcontainer` dependency completely optional in `containerd-shim-wasm`. At least this will unblock you from the Windows support work, but you still need to implement non-youki version for the wasmtime shim that does not dep on libcontainer.

jsturtevant:
> but you still need to implement non-youki version for the wasmtime shim that does not dep on libcontainer.

was hoping to share more but with the direction towards youki, in the short term this is probably the best option.

jsturtevant:
> I can make `libcontainer` dependency completely optional in `containerd-shim-wasm`.

👍

crates/containerd-shim-wasmedge/Cargo.toml
```diff
@@ -18,7 +18,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 nix = { workspace = true }
 libc = { workspace = true }
-libcontainer = { workspace = true }
+libcontainer = { workspace = true, features = ["default"] }
```
jprendes, line 21: **question:** Shouldn't this go through `containerd-shim-wasm = { workspace = true, features = ["libcontainer", ...] }` to make sure that `containerd-shim-wasm` has `libcontainer` enabled ?

crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -142,6 +157,170 @@ pub trait Instance {
...
+/// YoukiInstance is a trait that gets implemented by a WASI runtime that
+/// uses youki's libcontainer library as the container runtime.
+/// It provides default implementations for some of the Instance trait methods.
+/// The implementor of this trait is expected to implement the
+///     new_youki()
+///     get_exit_code()
+///     get_id()
+///     get_root_dir()
+///     build_container()
+/// methods.
```
jprendes, line 169: **nit or nits:** use markdown:
```suggestion
/// The implementor of this trait is expected to implement the following methods:
/// * `new_youki(id, cfg)`
/// * `get_exit_code()`
/// * `get_id()`
/// * `get_root_dir()`
/// * `build_container()`
```

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -45,11 +37,11 @@ static mut STDERR_FD: Option<RawFd> = None;
 
 static DEFAULT_CONTAINER_ROOT_DIR: &str = "/run/containerd/wasmedge";
 
-type ExitCode = (Mutex<Option<(u32, DateTime<Utc>)>>, Condvar);
+type ExitCode = Arc<(Mutex<Option<(u32, DateTime<Utc>)>>, Condvar)>;
```
jprendes, line 40:
```suggestion
use containerd_shim_wasm::sandbox::instance::ExitCode;
```
(and merge with line 11)

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -114,137 +107,45 @@ impl Instance for Wasi {
...
 
-impl Wasi {
-    fn build_container(
-        &self,
-        stdin: Option<i32>,
-        stdout: Option<i32>,
-        stderr: Option<i32>,
-    ) -> anyhow::Result<Container> {
         let syscall = create_syscall();
+        let err_msg = |err| format!("failed to create container: {}", err);
```
jprendes, line 129:
```suggestion
        let err_other = |err| Error::Others(format!("failed to create container: {}", err));
```

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -114,137 +107,45 @@ impl Instance for Wasi {
...
         let container = ContainerBuilder::new(self.id.clone(), syscall.as_ref())
             .with_executor(vec![Box::new(WasmEdgeExecutor {
                 stdin,
                 stdout,
                 stderr,
-            })])?
-            .with_root_path(self.rootdir.clone())?
+            })])
+            .map_err(|err| Error::Others(err_msg(err)))?
```
jprendes, line 136:
```suggestion
            .map_err(err_other)?
```

jprendes (approved):
This looks great, thanks for working on it!

Some minor comments. The comments in `wasmedge` also apply to `wasmtime`.

In the PR description you mention
> The downside is that any new shim instance needs to implement both YoukiInstance and the Instance traits.

But there's a blanket implementation of `Instance` for `YoukiInstance`, so there's no need to implement both traits, or am I missing something?

Makefile
```diff
@@ -18,6 +18,8 @@ KIND_CLUSTER_NAME ?= containerd-wasm
 build:
 	cargo build -p containerd-shim-wasm --features generate_bindings $(RELEASE_FLAG)
 	cargo build $(RELEASE_FLAG)
+	# no libcontainer feature on containerd-shim-wasm
+	cargo build $(RELEASE_FLAG) --no-default-features -p containerd-shim-wasm
```
jsturtevant, line 22: This feels backwards.  The shims that want to use the `youkiInstance` should be opting into the feature.

crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -142,6 +157,170 @@ pub trait Instance {
...
+#[cfg(feature = "libcontainer")]
+pub trait YoukiInstance {
```
jsturtevant, line 171: thoughts on putting this (and related imports) into its own file?  Then you only need the one `cfg(feature=libcontainer)` and makes it easier to navigate and find all the things.

crates/containerd-shim-wasmedge/Cargo.toml
Mossaka: containerd-shim-wasm, by default, depends on libcontainer. I am not entirely if I understand your question.

crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -40,6 +41,12 @@ env_logger = { workspace = true }
...
+default = ["libcontainer", "libcontainer/default"]
+libcontainer = ["dep:libcontainer"]
 generate_bindings = ["ttrpc-codegen"]
-generate_doc = []
\ No newline at end of file
+generate_doc = []
+libseccomp = ["libcontainer", "libcontainer/libseccomp"]
+systemd = ["libcontainer", "libcontainer/systemd"]
+v2 = ["libcontainer", "libcontainer/v2"]
+v1 = ["libcontainer", "libcontainer/v1"]
```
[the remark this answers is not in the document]
Mossaka: Done.

crates/containerd-shim-wasm/src/sandbox/instance.rs
Mossaka: Done.

crates/containerd-shim-wasmedge/src/instance.rs
Mossaka: Done.

crates/containerd-shim-wasmedge/src/instance.rs
Mossaka: Done.

crates/containerd-shim-wasmedge/src/instance.rs
Mossaka: Done.

crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -40,6 +41,12 @@ env_logger = { workspace = true }
 rand = "0.8"
 
 [features]
-default = []
+default = ["libcontainer", "libcontainer/default"]
```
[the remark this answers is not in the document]
Mossaka: This does make `libcontainer` a default dep. Do you see any problems with making it a default?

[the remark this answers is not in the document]
jsturtevant: If we make it default, then that means shim authors would need to remove it if they don't want to use youki.  Youki is a nice to have (makes things like containers and wasm running side easier, container cgroups and other things) but not necessary so I think authors should be opting into it.

crates/containerd-shim-wasm/src/sandbox/instance.rs
Mossaka: Done.

crates/containerd-shim-wasm/Cargo.toml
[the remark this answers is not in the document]
Mossaka: Done.

Makefile
Mossaka: Done.

crates/containerd-shim-wasmedge/Cargo.toml
jprendes: my concern was that something like this
```toml
[package]
name = "my_crate"
version = "0.1.0"

[dependencies]
containerd-shim-wasm = { path = "../runwasi/crates/containerd-shim-wasm", default-features = false}
containerd-shim-wasmedge = { path = "../runwasi/crates/containerd-shim-wasmedge", default-features = false}
```

would not enable `libcontainer` in `containerd-shim-wasm`, but it does :-)

crates/containerd-shim-wasm/src/sandbox/mod.rs
```diff
@@ -9,6 +9,8 @@ pub mod instance;
 pub mod instance_utils;
 pub mod manager;
 pub mod shim;
+#[cfg(feature = "libcontainer")]
+pub mod youki_instance;
```
jprendes, line 13: **suggestion:** also re-export `YoukiInstance`
```suggestion
#[cfg(feature = "libcontainer")]
pub mod youki_instance;
#[cfg(feature = "libcontainer")]
pub use youki_instance::YoukiInstance;
```

crates/containerd-shim-wasm/src/sandbox/youki_instance.rs
```diff
@@ -0,0 +1,188 @@
...
+/// It provides default implementations for some of the Instance trait methods.
+/// The implementor of this trait is expected to implement the
+/// * `new_youki()`
+/// * `get_exit_code()`
+/// * `get_id()`
+/// * `get_root_dir()`
+/// * `build_container()`
+/// methods.
+#[cfg(feature = "libcontainer")]
+pub trait YoukiInstance {
```
jprendes, line 37: **question:** should this be `LibcontainerInstance`? This is the first mention of `youki` AFAIK

crates/containerd-shim-wasm/src/sandbox/youki_instance.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// The WASI engine type
+    type E: Send + Sync + Clone;
+
+    /// Create a new instance
+    fn new_youki(id: String, cfg: Option<&InstanceConfig<Self::E>>) -> Self;
```
jprendes, line 42: **suggestion / preference:** rename this to plain `new`:
* In L.65 do
```rust
<Self as YoukiInstance>::new(...)
```
* In the implementations, instead of doing
```rust
use containerd_shim_wasm::sandbox:YoukiInstance;
impl YoukiInstance for Wasi { ... }
```
do
```rust
use containerd_shim_wasm::sandbox;
impl sandbox::YoukiInstance for Wasi { ... }
```

jprendes (approved):
I really like how this is looking, Thanks @Mossaka !

devigned (commented):
Looks great! I've added a couple comments.

Cargo.toml
```diff
@@ -32,7 +32,7 @@ thiserror = "1.0"
 libc = "0.2.147"
 oci-spec = { version = "0.6.1", features = ["runtime"] }
 sha256 = "1.3.0"
-libcontainer = "0.1"
+libcontainer = { version = "0.1", default-features = false }
```
jsturtevant, line 35: is `default-features=false` needed?

jsturtevant (commented):
looking really good!  I think once the last few open comments are addressed this is in good shape

crates/containerd-shim-wasm/src/sandbox/youki_instance.rs
```diff
@@ -0,0 +1,188 @@
...
+    fn get_root_dir(&self) -> Result<PathBuf, Error>;
+
+    /// Build the container
+    fn build_container(&self) -> Result<Container, Error>;
+}
+
+/// Default implementation of the Instance trait for YoukiInstance
+/// This implementation uses the libcontainer library to create and start
+/// the container.
+#[cfg(feature = "libcontainer")]
```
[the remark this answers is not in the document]
Mossaka: Done.

crates/containerd-shim-wasm/src/sandbox/youki_instance.rs
```diff
@@ -0,0 +1,188 @@
...
+/// uses youki's libcontainer library as the container runtime.
+/// It provides default implementations for some of the Instance trait methods.
+/// The implementor of this trait is expected to implement the
+/// * `new_youki()`
+/// * `get_exit_code()`
+/// * `get_id()`
+/// * `get_root_dir()`
+/// * `build_container()`
+/// methods.
+#[cfg(feature = "libcontainer")]
```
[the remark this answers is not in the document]
Mossaka: Done.

crates/containerd-shim-wasm/src/sandbox/mod.rs
Mossaka: Done.

crates/containerd-shim-wasm/src/sandbox/youki_instance.rs
Mossaka: Done.

crates/containerd-shim-wasm/src/sandbox/youki_instance.rs
Mossaka: Done.

crates/containerd-shim-wasm/src/sandbox/youki_instance.rs
```diff
@@ -0,0 +1,188 @@
...
+        if signal as i32 != SIGKILL && signal as i32 != SIGINT {
+            return Err(Error::InvalidArgument(
+                "only SIGKILL and SIGINT are supported".to_string(),
+            ));
+        }
+        let container_root = get_instance_root(root_dir, id.as_str())?;
+        let mut container = Container::load(container_root).with_context(|| {
+            format!("could not load state for container {id}", id = id.as_str())
+        })?;
+        let signal = Signal::try_from(signal as i32)
```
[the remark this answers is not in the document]
Mossaka: Done.

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -114,137 +106,45 @@ impl Instance for Wasi {
...
+            .with_root_path(self.rootdir.clone())
+            .map_err(err_others)?
             .as_init(&self.bundle)
             .with_systemd(false)
-            .build()?;
+            .build()
+            .map_err(err_others)?;
+        // Close the fds now that they have been passed to the container process
+        // so that we don't leak them.
+        stdin.map(close);
```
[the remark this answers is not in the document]
Mossaka: I think we only care about `build` calling with an Ok result, but you remind me to close file descriptors on the wasmtime shim.

Cargo.toml
Mossaka: Actually not, let me remove it.

Mossaka: Actually it is needed because otherwise you will get this error

```
warning: /home/mossaka/developer/runwasi/crates/containerd-shim-wasm/Cargo.toml: `default-features` is ignored for libcontainer, since `default-features` was not specified for `workspace.dependencies.libcontainer`, this could become a hard error in the future
```

crates/containerd-shim-wasmedge/Cargo.toml
Mossaka: I think my last fix has resolved this.

Mossaka:
> But there's a blanket implementation of Instance for YoukiInstance, so there's no need to implement both traits, or am I missing something?

Ah that was a mistaken. I used a different implementation so that shim author only needs to implement `LibcontainerInstance` trait now.

jsturtevant:
not sure what our current set up is in regards to CI and commits, but maybe squash the commits?

crates/containerd-shim-wasm/src/sandbox/youki_instance.rs
jprendes: That's why you would need
```rust
<Self as LibcontainerInstance>::new(...)
```

But I'm not sure that's better than the method being called `new_libcontainer`.
I guess it's just odd that the method `new` is part of a trait, and not part of the struct's impl.

I'm happy to keep this as `new_libcontainer`.

crates/containerd-shim-wasmedge/src/instance.rs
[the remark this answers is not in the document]
jprendes: I woiuld be nice if `maybe_open_stdio` returned an RAII wrapper.
A simple approach would be to return the `File` instead of the `RawFd`, and using `as_raw_fd()` when needed instead of using `into_raw_fd()` early on.

jprendes (approved):
LGTM!

crates/containerd-shim-wasm/src/sandbox/libcontainer_instance.rs
```diff
@@ -0,0 +1,187 @@
...
+                Err(e) => {
+                    if e == Errno::ECHILD {
+                        log::info!("no child process");
+                        0
+                    } else {
+                        panic!("waitpid failed: {}", e);
+                    }
+                }
+            } as u32;
+            let mut ec = lock.lock().unwrap();
```
utam0k, line 97:
```suggestion
            let mut ec = lock.lock().map_err(|err| Error::Any(anyhow::anyhow!("failed get a lockr: {}", err)))?;
```

crates/containerd-shim-wasm/src/sandbox/libcontainer_instance.rs
```diff
@@ -0,0 +1,187 @@
...
+use crate::sandbox::InstanceConfig;
+use std::{path::PathBuf, thread};
+
+use super::{error::Error, instance::Wait, Instance};
+
+/// LibcontainerInstance is a trait that gets implemented by a WASI runtime that
+/// uses youki's libcontainer library as the container runtime.
+/// It provides default implementations for some of the Instance trait methods.
+/// The implementor of this trait is expected to implement the
+/// * `new_youki()`
```
utam0k, line 30:
```suggestion
/// * `new_libecontainer()`
```

crates/containerd-shim-wasmedge/src/instance.rs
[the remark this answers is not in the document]
jsturtevant: > I woiuld be nice if `maybe_open_stdio` returned an RAII wrapper. A simple approach would be to return the `File` instead of the `RawFd`, and using `as_raw_fd()` when needed instead of using `into_raw_fd()` early on.

https://github.com/containerd/runwasi/pull/240 is similar to what you reference here, thought I am not sure what RAII is

Mossaka:
Going to merge this in. Thanks everyone for reviewing it!! Really appreciate it.

crates/containerd-shim-wasm/src/sandbox/libcontainer_instance.rs
```diff
@@ -0,0 +1,187 @@
...
+    fn get_exit_code(&self) -> ExitCode;
+
+    /// Get the ID of the instance
+    fn get_id(&self) -> String;
+
+    /// Get the root directory of the instance
+    fn get_root_dir(&self) -> Result<PathBuf, Error>;
+
+    /// Build the container
+    fn build_container(&self) -> Result<Container, Error>;
```
cpuguy83, line 53: `build_container` seems like a fairly heavy trait function.

I think if we add something like `open_stdio` to the trait we can implement common `build_container`?

crates/containerd-shim-wasm/src/sandbox/libcontainer_instance.rs
cpuguy83: I think its good to let this panic anyway.
If we can't lock something is very wrong (dirty mutex state) and this should crash the program.

cpuguy83 (changes requested):
I haven't given this a thorough review yet, but I like where this is going at lot.

crates/containerd-shim-wasmedge/src/instance.rs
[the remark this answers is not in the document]
jprendes: Sorry, that's my inner C++ betraying me.
It's basically when you have a struct implementing `Drop` to cleanup after itself.
When you `drop` a `File`, it automatically closes the file for you.
It is kind of what #240 does, but instead of using `into_raw_fd` the executors would hold a `Option<File>` instead of an `Option<RawFd>`, and the conversion to `RawFd` would be done at le last possible moment.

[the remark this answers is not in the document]
jsturtevant: I tried this but ran into some lifetime issues with the File and opted for a smaller change but if you have any suggestions be happy to fix that up 👍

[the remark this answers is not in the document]
jprendes: I was thinking something like this https://github.com/jsturtevant/runwasi/compare/use-cross-platfrom-file...jprendes:runwasi:use-cross-platfrom-file

crates/containerd-shim-wasm/src/sandbox/libcontainer_instance.rs
jprendes: IIUC, that would move the stdio redirection work from the executor to the instance.
Is there any instance / executor that would not want to do the stdio redirection trick (slight/spin/wws)?
Can that be handled by default in `Instance` instead of `LibcontainerInstance`?

PR #210, opened 2023-07-27
Add crate feature to disable libwasmedge static linking

jprendes:
Fixes #206

PR #227, opened 2023-08-03
[wasmedge] fix static linking unconditionally enabled

jprendes:
See [this comment](https://github.com/containerd/runwasi/pull/210#issuecomment-1663729113)

This PR fixes the static feature being unconditionally enabled, and adds tests to verify static / dynamic linking.
```
$ make test 2>/dev/null | grep linking
test test::check_static_linking ... ok
test test::check_dynamic_linking ... ok
```

@Mossaka @cpuguy83

### DIFF
--- a/crates/containerd-shim-wasmedge/Cargo.toml
+++ b/crates/containerd-shim-wasmedge/Cargo.toml
@@ -8,7 +8,7 @@ containerd-shim = { workspace = true }
 containerd-shim-wasm = { workspace = true }
 log = { workspace = true }
 ttrpc = { workspace = true }
-wasmedge-sdk = { version = "0.11.2", features = [ "standalone", "static" ] }
+wasmedge-sdk = { version = "0.11.2" }
 chrono = { workspace = true }
 anyhow = { workspace = true }
 cap-std = { workspace = true }
--- a/crates/containerd-shim-wasmedge/Cargo.toml
+++ b/crates/containerd-shim-wasmedge/Cargo.toml
@@ -9,6 +9,7 @@ containerd-shim-wasm = { workspace = true }
 log = { workspace = true }
 ttrpc = { workspace = true }
 wasmedge-sdk = { version = "0.11.2" }
+wasmedge-sys = "*"
 chrono = { workspace = true }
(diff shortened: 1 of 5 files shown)


Mossaka (approved):
LGTM!

crates/containerd-shim-wasmedge/src/lib.rs
```diff
@@ -2,3 +2,30 @@ pub mod error;
 pub mod executor;
 pub mod instance;
 pub mod oci_utils;
+
+#[cfg(test)]
+mod test {
+    #[cfg(feature = "static")]
+    #[test]
+    fn check_static_linking() {
+        let output = std::process::Command::new("ldd")
```
ipuustin, line 11: Can we assume that `ldd` is installed? Maybe skip the test if the binary is not present.

jprendes: We can avoid using `lld` by using `libc::dladdr`.
I've pushed a new version using that instead :-)

crates/containerd-shim-wasmedge/Cargo.toml
```diff
@@ -8,7 +8,8 @@ containerd-shim = { workspace = true }
 containerd-shim-wasm = { workspace = true }
 log = { workspace = true }
 ttrpc = { workspace = true }
-wasmedge-sdk = { version = "0.10.1", features = [ "standalone", "static" ] }
+wasmedge-sdk = { version = "0.11.0" }
```
ipuustin, line 11: `0.11.2` is already available, can we use it?

ipuustin (changes requested):
One small nit, otherwise looks good to me!

crates/containerd-shim-wasmedge/Cargo.toml
jprendes: released a few hours ago :-)

ipuustin (approved):
LGTM!

jprendes:
Rebased to `main`.

cpuguy83 (approved):

cpuguy83 (approved):

PR #231, opened 2023-08-04
refactor: remove engine getter and the generic type `E`

Mossaka:
There is no practical use of `get_engine()` function from the `EngineGetter` trait, so I decided to remove it entirely. 

It looks like youki's Executor is all we need

### DIFF
--- a/crates/containerd-shim-wasmtime/src/instance.rs
+++ b/crates/containerd-shim-wasmtime/src/instance.rs
@@ -27,6 +27,7 @@ use nix::sys::wait::{Id as WaitID, WaitPidFlag, WaitStatus};
 
 use wasmtime::Engine;
 
+use crate::engine::WasmtimeEngine;
 use crate::executor::WasmtimeExecutor;
 use libcontainer::signal::Signal;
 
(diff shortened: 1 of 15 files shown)


0xE282B0:
I like it, that makes bootstrapping a new shim easier.

ipuustin (approved):
I like this change a lot, because it removes the need for direct runtime dependency (for `wasmtime::Engine´ and similar. I've been having trouble with the need for wasmtime/wasmedge engine in integration testing.

jsturtevant (commented):
LGTM

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -93,23 +93,23 @@ fn determine_rootdir<P: AsRef<Path>>(bundle: P, namespace: String) -> Result<Pat
...
         Wasi {
             id,
             exit_code: Arc::new((Mutex::new(None), Condvar::new())),
-            engine: cfg.get_engine(),
             stdin: cfg.get_stdin().unwrap_or_default(),
             stdout: cfg.get_stdout().unwrap_or_default(),
             stderr: cfg.get_stderr().unwrap_or_default(),
             bundle,
             rootdir,
+            engine,
```
jsturtevant, line 112: is this needed here? The wasm-edge shim does the engine instance at the time of the execution and it looks like that could be done here as well.  Is there a perf benefit to doing it up front before execution starts?

cpuguy83:
The reason `EngineGetter` exists is because we wanted to be able to pass a single wasmtime engine around instead of creating a new wasmtime engine for every container AND for library consumers (e.g. slight and spin) to be able to pass along a custom engine (specifically I recall Spin itself had some specific requirements not part of `::Default()`).
Then people wanted to be able to use a runtime other than wasmtime and so it was turned into a generic.

This change looks like it is running afoul of the original desire to not create a new engine for every container.

Mossaka:
> Maybe there is a way to achieve both?

I will think about this...

jsturtevant:
Another thought I had, was do we know how expensive the creation of an engine is?  I guess this might depend on the runtime but it might help to quantify how expensive the operation is.

Mossaka:
I don't really have a way to reduce the amount of complexity that the `EngineGetter` trait introduces to the `containerd-shim-wasm` crate, so going to close this PR.

Mossaka:
This is an alternative implementation. https://github.com/containerd/runwasi/pull/231/commits/04ebc3e63e05f65481cb4338aaaf81fd74685fa3

I don't think it reduces that much of complexity.

jprendes:
We can remove `EngineGetter` and use `Default` instead.
Implementations where `Default` is not enough can use a wrapper:
```rust
struct MyEngineWrapperType(OriginalEngineType)
impl Default for MyEngineWrapperType {
  fn default() -> Self { ... }
}
```

That has the improved ergonomics for most cases, and the customization for other cases, unless I'm missing something.

jprendes:
I'll create a separate PR in case that change turns out to be more controversial.

PR #238, opened 2023-08-09
Refactor use statements to enable building Windows

jsturtevant:
This PR does **_not_** enable windows features but it does make it so the shim will **_build_** on Windows.  This will make the following PR's smaller and easier to digest to do it in steps.  It also makes it easier while other changes go in without having to rebase complex changes.

one small step towards #49 

For now this disables `dead-code` and `unused imports` but should be turned on before in following PR's as functionality is added.

Todo in the PR:
Add a CI 

- [x] job that validates the build on windows.


jsturtevant:
The CI is passing 🥳 

I've opened #240, #239 to make this PR smaller and easier to review.  Since #226 is close to merging I would suggest we get these merged and that would also make this PR much smaller and simpler.

.github/scripts/build-windows.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+choco install -y wasmedge
```
jprendes, line 2: Is there a way to specify the version?
Could you create an issue in [wasmedge-rust-sdk](https://github.com/WasmEdge/wasmedge-rust-sdk/issues) to support standalone Windows?

jprendes (commented):
Thanks for working in this :-).
I think native windows support would add a lot of value.

I haven't finished, but here are my comments so far.

Also, I'm not a big fan of the new `cfg_window` / `cgf_unix` macros.
I think they add a lot of [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) to the code.
I would much prefer to create separate os-specific modules, and factor out the code that differs between each os.

.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[target.'cfg(windows)']
```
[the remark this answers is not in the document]
jsturtevant: I added this to the checklist here: https://github.com/containerd/runwasi/issues/49

.github/scripts/build-windows.sh
```diff
@@ -0,0 +1,3 @@
...
+choco install -y llvm
```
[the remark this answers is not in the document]
jsturtevant: yes needed for wasmedge

.github/scripts/build-linux.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
```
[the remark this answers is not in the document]
jsturtevant: agreed, not sure why they were named this and also why they aren't in `scripts` folder as that would make it usable/discoverable for others to consume too.

.github/scripts/build-windows.sh
jsturtevant: https://github.com/WasmEdge/wasmedge-rust-sdk/issues/54

.github/scripts/build-windows.sh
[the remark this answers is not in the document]
jprendes: then it's odd that wasmedge doesn't install llvm as a dependency

.github/scripts/build-linux.sh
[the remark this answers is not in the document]
jprendes: I think they are in `.github/scripts` instead of plain `scripts` because they were intended to be used in github workflows.
If we don't plan to use them outside of workflows, I think the current folder is fine.

.github/scripts/build-windows.sh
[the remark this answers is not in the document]
jsturtevant: I will double check but it wouldn't build when I was running locally

.github/scripts/build-windows.sh
jsturtevant: pinned the version

.github/scripts/build-linux.sh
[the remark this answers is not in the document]
jsturtevant: we give this set of instructions in the [readme](https://github.com/containerd/runwasi#building), we could put it behind a make file target and use it locally and in ci which would keep them in sync

.github/scripts/build-windows.sh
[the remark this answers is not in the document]
jsturtevant: It is the clang dep, When I have llvm uninstalled and build the wasmedge project locally I get:

```
cargo build -p containerd-shim-wasmedge --no-default-features

thread 'main' panicked at 'Unable to find libclang: "couldn't find any valid shared libraries matching: 
['clang.dll', 'libclang.dll'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can
 be found (invalid: [])"', C:\Users\jstur\.cargo\registry\src\index.crates.io-6f17d22bba15001f\bindgen-
0.65.1\lib.rs:603:31       
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace           
```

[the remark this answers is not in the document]
jprendes: Ah, ok, that's because of bindgen, which is used in the build script to generate the rust bindings to the c codebase. But I believe its not needed by the final binary (although I might be wrong)
IIRC any clang >= clang 6 is enough for wasmedge bindgen.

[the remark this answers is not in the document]
jsturtevant: I've made a note in the script as to why this is needed, maybe can revisit when sorting out https://github.com/WasmEdge/wasmedge-rust-sdk/issues/54

crates/containerd-shim-wasm/src/sandbox/oci.rs
```diff
@@ -103,7 +119,16 @@ pub fn setup_prestart_hooks(hooks: &Option<oci_spec::runtime::Hooks>) -> Result<
...
+                        #[cfg(unix)]
+                        {
+                            let hook_process_pid = Pid::from_raw(hook_process.id() as i32);
+                            let _ = signal::kill(hook_process_pid, signal::Signal::SIGKILL);
+                        }
+
+                        #[cfg(windows)]
+                        {
+                            let _ = hook_process.kill();
+                        }
```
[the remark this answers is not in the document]
jsturtevant: Took a look at the underlying source for `process.kill()` and it calls `libc::kill(self.pid, libc::SIGKILL)` https://doc.rust-lang.org/src/std/sys/unix/process/process_unix.rs.html#681 which is the same as nix kill https://docs.rs/nix/latest/src/nix/sys/signal.rs.html#929

So I've combined this to a single call

jsturtevant:
latest failure was the outofspace we are seeing on other PRs:

```
 Warning  FailedCreatePodSandBox  43s   kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd container: failed to create prepare snapshot dir: mkdir /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/new-3941906082/fs: no space left on device
```

jsturtevant:
> Also, I'm not a big fan of the new cfg_window / cgf_unix macros.
I think they add a lot of [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) to the code.
I would much prefer to create separate os-specific modules, and factor out the code that differs between each os.

@jprendes I cleaned this up where it made sense. In some cases, I believe its easier to read if it is in-lined but let me know what you think.   The one last place I would like to do this is around the mounting code but I don't know the abstraction to put in place just yet. I've started this on the Windows side but haven't full flushed it out but will follow up on moving it to an abstraction in the next PR.

jsturtevant:
test failure is #252

cpuguy83 (approved):
LGTM!

crates/containerd-shim-wasm/src/lib.rs
```diff
@@ -5,7 +5,9 @@
 
 pub mod sandbox;
 
+mod macros;
 pub mod services;
+pub mod sys;
```
jprendes, line 10: I think using the [`path` attribute](https://doc.rust-lang.org/reference/items/modules.html#the-path-attribute) combined with [`cfg_atr`](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute) would be cleaner:
```rust
#[cfg_attr(unix, path = "sys/unix.rs")]
#[cfg_attr(windows, path = "sys/windows.rs")]
mod sys;
```
With that I think you can get rid of `cfg_windows` / `cfg_unix` macros.

jprendes (changes requested):
Thanks, this is looking great!
Looking forward to having this merged :-)

scripts/setup-windows.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+choco install -y wasmedge --version 0.13.1
```
Mossaka, line 2: Is choco built-in to Windows?

crates/containerd-shim-wasm/src/sys/mod.rs
```diff
@@ -0,0 +1,18 @@
+#[cfg(unix)]
+mod unix;
+#[cfg(windows)]
+mod windows;
+
+pub mod metrics {
```
Mossaka, line 6: I like the naming of these modules!

crates/containerd-shim-wasm/src/sandbox/error.rs
```diff
@@ -39,6 +39,7 @@ pub enum Error {
     #[error("{0}")]
     Json(#[from] serde_json::Error),
     /// Error from the system
+    #[cfg(unix)]
```
Mossaka, line 42: What's the equivalent of errno on the Windows side?

Mossaka (approved):
LGTM!!

crates/containerd-shim-wasm/src/sandbox/error.rs
jsturtevant: windows error codes: https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-

I will add section here for windows when/if needed in the next pr that adds the Windows implementation.

crates/containerd-shim-wasm/src/lib.rs
jsturtevant: I've updated and I think it does simplify things! I did something similar on the instance as well.

We did loose the granular modules (but could get them back if wanted with a few tweaks) https://github.com/containerd/runwasi/pull/238#discussion_r1300726884

crates/containerd-shim-wasm/src/sys/mod.rs
jsturtevant: could you take a look again? I change the module structure in https://github.com/containerd/runwasi/pull/238#discussion_r1300764238.  Do you feel we need those for organization or does the new structure work

crates/containerd-shim-wasmtime/src/instance/instance_windows.rs
```diff
@@ -1,8 +1,8 @@
 use std::path::PathBuf;
 
-use containerd_shim_wasm::sandbox::{Instance, InstanceConfig};
-use containerd_shim_wasm::sandbox::instance::{Wait, ExitCode};
 use containerd_shim_wasm::sandbox::error::Error;
+use containerd_shim_wasm::sandbox::instance::{ExitCode, Wait};
```
Mossaka, line 4: is this sorted by rustfmt?

Mossaka (approved):
lgtm!

crates/containerd-shim-wasm/src/sys/mod.rs
Mossaka: The new structure works

jsturtevant:
looks like a different issue with ubuntu 22.04, now that storage has been solved.

Mossaka:
Cool, shall I merge this in? @jsturtevant

crates/containerd-shim-wasm/src/sys/mod.rs
jprendes: You could still keep the module structure.
You coud create a folder for each system, and place files for each module.

PR #239, opened 2023-08-09
Move code that is only used in the tests

jsturtevant:
This is some cleanup for functions that are only used in the tests for the shims.

### DIFF
--- a/crates/containerd-shim-wasmedge/src/instance.rs
+++ b/crates/containerd-shim-wasmedge/src/instance.rs
@@ -1,7 +1,6 @@
 use std::fs::File;
 use std::io::prelude::*;
 use std::io::ErrorKind;
-use std::os::unix::io::RawFd;
 use std::sync::{Arc, Condvar, Mutex};
 use std::thread;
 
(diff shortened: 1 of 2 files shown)


crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -293,6 +276,24 @@ mod wasitest {
...
 
+    static mut STDIN_FD: Option<RawFd> = None;
+    static mut STDOUT_FD: Option<RawFd> = None;
+    static mut STDERR_FD: Option<RawFd> = None;
+
+    fn reset_stdio() {
+        unsafe {
+            if STDIN_FD.is_some() {
+                dup2(STDIN_FD.unwrap(), STDIN_FILENO);
+            }
```
utam0k, line 287:
```suggestion
            if let Some(fd) = STDIN_FD {
                dup2(fd, STDIN_FILENO);
            }
```

Mossaka (approved):
LGTM!

crates/containerd-shim-wasmedge/src/instance.rs
Mossaka: Done.

Mossaka:
Hey, I went ahead and merged main branch. After the CI passes, I will merge this in

cpuguy83 (approved):

PR #240, opened 2023-08-10
Make the utility maybe_open_stdio cross platform

jsturtevant:
The removes an unused function in wasmtime and also updates `maybe_open_stdio` to cross-platform helper (part of #49 and #238)


utam0k (approved):

Mossaka (approved):
LGTM!

cpuguy83:
Needs a rebase.

jsturtevant:
@jprendes the second commit here has your suggestion, Since the merge of #156 stdio is passed to both so I need an additional step

jsturtevant:
this is ready for review again
/assign @Mossaka @utam0k @jprendes

jprendes (approved):
I'm not a big fan of `try_clone`-ing the file, since that opens yet another `fd` for the same file.
But it works, and would unblock you to continue working on windows support.
I am happy for this to be merged as is.

@cpuguy83: I agree with you in https://github.com/containerd/runwasi/pull/226#discussion_r1293814105 and think that would also improve the file handling here.

crates/containerd-shim-wasm/src/sandbox/instance_utils.rs
```diff
@@ -1,10 +1,11 @@
 //! Common utilities for the containerd shims.
 use crate::sandbox::error::Error;
 use anyhow::{bail, Context, Result};
+use nix::unistd::{dup, dup2};
 use std::{
-    fs::{self, OpenOptions},
+    fs::{self, File, OpenOptions},
     io::ErrorKind,
-    os::fd::{IntoRawFd, RawFd},
+    os::fd::AsRawFd,
```
jprendes, line 8: **question:** Does this build on Windows? I thought `AsRawFd` was only available on unix.

jsturtevant: It is, will clean it up in #238

jsturtevant:
> I'm not a big fan of `try_clone`-ing the file, since that opens yet another `fd` for the same file.

I wasn't too happy with this either but wasn't sure how else to handle the ownership here (open to suggestions). Initially, reading the [docs](https://doc.rust-lang.org/std/fs/struct.File.html#method.try_clone) I didn't get that it will create a second FD since it says it would use "same underlying file handle" but I see it uses [fcntl](https://doc.rust-lang.org/src/std/os/fd/owned.rs.html#109) which will give a new FD.  

Will open an issue for https://github.com/containerd/runwasi/pull/226#discussion_r1293814105 to follow up.

jprendes:
Feel free to drop b1b2a7fc9f8080b12329cef98270f78e266a60e4 from the PR as that would be part of the new issue :-)

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -13,6 +8,11 @@ use containerd_shim_wasm::sandbox::instance_utils::maybe_open_stdio;
 use containerd_shim_wasm::sandbox::{EngineGetter, InstanceConfig};
 use nix::unistd::close;
 use serde::{Deserialize, Serialize};
+use std::fs::File;
```
Mossaka, line 11: Why are these imports moved to here? Did you use a formatter to do it?

Mossaka (approved):
LGTM!

crates/containerd-shim-wasmedge/src/instance.rs
jsturtevant: It think so, I didn't notice but I did run `make fix` which runs fmt and clippy and looks like it re-ordered them alphabetically

Mossaka: umm rustfmt and clippy won't re-order imports alphabetically IIRC.

PR #244, opened 2023-08-16
Consolidate setup steps to keep in sync

jsturtevant:
This consolidates the scripts and came out of https://github.com/containerd/runwasi/pull/238#discussion_r1293544759.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+sudo apt -y update
+sudo apt install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -1,3 +1,0 @@
-#!/bin/bash
-sudo apt -y update
-sudo apt install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,8 @@ jobs:
       - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           components: rustfmt, clippy
-      - name: Setup OCI runtime build env
-        run: ${GITHUB_WORKSPACE}/.github/scripts/build.sh
+      - name: Setup build env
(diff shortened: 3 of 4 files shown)


scripts/setup.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 sudo apt -y update
-sudo apt install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev
\ No newline at end of file
+sudo apt install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev libssl-dev
```
jprendes, line 3: Do we need libssl-dev? Or was the README out of sync?

jprendes (approved):
LGTM! Thanks @jsturtevant !

Makefile
```diff
@@ -36,6 +36,10 @@ test:
 	# run wasmedge test without the default `static` feature
 	RUST_LOG=trace cargo test --package containerd-shim-wasmedge --verbose --no-default-features --features standalone -- --nocapture
 
+.PHONY: setup-env
+setup-env:
+	sudo ./scripts/setup.sh
```
Mossaka, line 41: I remember that @cpuguy83 had this comment https://github.com/containerd/runwasi/pull/186#pullrequestreview-1538019005.

jsturtevant: Sounds good, I don't have a strong opinion here other than we could provide a script for the readme so the deps stay in sync.

scripts/setup.sh
jsturtevant: these are the requirements from youki, but it doesn't look like we needed it since CI was running with out it.  @Mossaka any ideas?

Mossaka: If it's not needed in the CI, then I don't it's needed anymore.

Mossaka (dismissed):

Mossaka:
It looks like the CI is failing

jprendes:
> It looks like the CI is failing

I can look into it in a few hours.

scripts/setup.sh
jsturtevant: removed it

Makefile
jsturtevant: switched to using a single script and no make file

Mossaka (approved):

PR #246, opened 2023-08-17
feat: resolve a name clash of the libcontainer features

Mossaka:
This commit resolves a name clash of the `libcontainer` feature from the containerd-shim-wasm, which undesirably forces all other features to bring in libseccomp and systemd dependencies. By renaming `libcontainer` to `libcontainer_default`, the consumer of the crate can specify `cgroupsv2` to directly access libcontainer's v2 feature without bringing in libseccomp and systemd dependencies. 

This will resolve this concerns @0xE282B0 raised https://github.com/deislabs/containerd-wasm-shims/pull/122#issuecomment-1682388913

FYI @jsturtevant

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,8 @@ KIND_CLUSTER_NAME ?= containerd-wasm
 .PHONY: build
 build:
 	cargo build -p containerd-shim-wasm --features generate_bindings $(RELEASE_FLAG)
-	cargo build -p containerd-shim-wasm --features libcontainer $(RELEASE_FLAG)
+	# compiling against libcontainer's default features (dependency on libseccomp)
+	cargo build -p containerd-shim-wasm --features libcontainer_default $(RELEASE_FLAG)
 	cargo build $(RELEASE_FLAG)
 
 .PHONY: check
--- a/crates/containerd-shim-wasm/Cargo.toml
+++ b/crates/containerd-shim-wasm/Cargo.toml
@@ -43,7 +43,7 @@ rand = "0.8"
 
 [features]
 default = []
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -17,7 +17,10 @@ KIND_CLUSTER_NAME ?= containerd-wasm
 .PHONY: build
 build:
 	cargo build -p containerd-shim-wasm --features generate_bindings $(RELEASE_FLAG)
-	cargo build -p containerd-shim-wasm --features libcontainer $(RELEASE_FLAG)
+	# compiling against libcontainer's default features (dependency on libseccomp)
+	cargo build -p containerd-shim-wasm --features libcontainer_default $(RELEASE_FLAG)
+	# compiling against libcontainer's cgroups v2 feature (no dependency on libseccomp)
+	cargo build -p containerd-shim-wasm --features cgroupsv2 $(RELEASE_FLAG)
```
jsturtevant, line 23: do we need to build each Indvidual feature or can we use something like `--all-features`?  I've seen that in other projects and I am not sure what the best thing to do here. Otherwise it seems like we need to do every combination that is exposed.

Mossaka: --all-features are just additively adding all the features, and that's good to build. But we also want to make sure each individual feature is buildable. 

Unrelated: I am thinking about writing an integration tests to `grep` the output of `cargo update` on different features to make sure `libseccomp` is adding/deleting.

jprendes (approved):
LGTM!

jsturtevant:
Is the space failure due to extra build caches for each configuration taking up space?

PR #247, opened 2023-08-17
Remove EngineGetter

jprendes:
This PR removes the `EngineGetter` trait and replaces it with the use of `Default::default()`.
In the case where an instance would have implemented a non-trivial `EngineGetter`, there are two options:
* If the associated `Engine` type is local to the crate, implement a suitable `Default`.
* Otherwise, use a wrapper struct for the associated `Engine` type and implement a suitable `Default` for the wrapper (and optionally also `Deref`)

### DIFF
--- a/crates/containerd-shim-wasmedge/src/instance.rs
+++ b/crates/containerd-shim-wasmedge/src/instance.rs
@@ -5,7 +5,7 @@ use containerd_shim_wasm::libcontainer_instance::LinuxContainerExecutor;
 use containerd_shim_wasm::sandbox::error::Error;
 use containerd_shim_wasm::sandbox::instance::ExitCode;
 use containerd_shim_wasm::sandbox::instance_utils::maybe_open_stdio;
-use containerd_shim_wasm::sandbox::{EngineGetter, InstanceConfig};
+use containerd_shim_wasm::sandbox::InstanceConfig;
 use nix::unistd::close;
 use serde::{Deserialize, Serialize};
 use std::fs::File;
@@ -13,11 +13,6 @@ use std::io::prelude::*;
 use std::io::ErrorKind;
 use std::os::fd::IntoRawFd;
 use std::sync::{Arc, Condvar, Mutex};
-use wasmedge_sdk::{
-    config::{CommonConfigOptions, ConfigBuilder, HostRegistrationConfigOptions},
-    plugin::PluginManager,
-    Vm, VmBuilder,
-};
 
 use std::{
     fs,
@@ -70,7 +65,7 @@ fn determine_rootdir<P: AsRef<Path>>(bundle: P, namespace: String) -> Result<Pat
 }
 
 impl LibcontainerInstance for Wasi {
-    type Engine = Vm;
+    type Engine = ();
 
     fn new_libcontainer(id: String, cfg: Option<&InstanceConfig<Self::Engine>>) -> Self {
         let cfg = cfg.unwrap(); // TODO: handle error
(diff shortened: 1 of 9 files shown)


jprendes:
@cpuguy83 , I know you had concerns with removing `EngingeGetter` in https://github.com/containerd/runwasi/pull/231#issuecomment-1668164559. Would this be acceptable?

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -136,29 +131,6 @@ impl LibcontainerInstance for Wasi {
...
-            .build()
-            .map_err(anyhow::Error::msg)?;
-        let vm = VmBuilder::new()
-            .with_config(config)
-            .build()
-            .map_err(anyhow::Error::msg)?;
-
-        Ok(vm)
-    }
-}
```
jprendes, line 160: This is a red herring.
It's actually never used, the `WasmEdgeExecutor` builds its own `Vm` from scratch.

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -70,7 +65,7 @@ fn determine_rootdir<P: AsRef<Path>>(bundle: P, namespace: String) -> Result<Pat
 }
 
 impl LibcontainerInstance for Wasi {
-    type Engine = Vm;
+    type Engine = ();
```
Mossaka, line 68: Why is this () instead of Vm?

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -136,29 +131,6 @@ impl LibcontainerInstance for Wasi {
     }
 }
 
-impl EngineGetter for Wasi {
```
Mossaka, line 139: Where is this code moved to?

Mossaka (commented):
LGTM! Had some questions on the WasmEdge shim.

crates/containerd-shim-wasmedge/src/instance.rs
Mossaka: Oh I guess this answers my questions

Mossaka: Should we build `Vm` in the instance and then passes it to Executor so that we enable sharing.

crates/containerd-shim-wasmedge/src/instance.rs
jprendes: This wasn't actually used, so it's just removed.

crates/containerd-shim-wasmedge/src/instance.rs
jprendes: It wasn't actually being used.

crates/containerd-shim-wasmedge/src/instance.rs
jprendes: Maybe.
I didn't share in this PR because the code here (removed) and the one in the executor were different, and wasn't sure the vm in the executor was shareable (it does things like installing plugins).
@CaptainVincent , do you have any input?

Mossaka (approved):

cpuguy83 (approved):

crates/containerd-shim-wasmedge/src/instance.rs
CaptainVincent: From my memory, it was used before the refactor into the container/executor architecture, but the current structure indeed always creates a new one. The aspects of sharing and reusing also involve the implementation's support for wasm runtime context. I thought enabling instance sharing only at the shim level might be questionable. I think there shouldn't be an issue removing it at this point.

PR #248, opened 2023-08-17
Moved determine_rootdir to containerd-shim-wasm

Mossaka:
This PR moves determine_rootdir to containerd-shim-wasm for sharing. It also removes unused dependencies in the repo.

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/instance_utils.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/instance_utils.rs
@@ -4,8 +4,9 @@ use std::io::ErrorKind;
 use std::path::{Path, PathBuf};
 
 use anyhow::{bail, Context, Result};
+use serde::{Deserialize, Serialize};
 
-use crate::sandbox::error::Error;
+use super::Error;
 
 /// Return the root path for the instance.
 ///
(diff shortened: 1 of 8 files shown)


crates/containerd-shim-wasm/src/sandbox/instance_utils.rs
```diff
@@ -45,6 +46,38 @@ pub fn maybe_open_stdio(path: &str) -> Result<Option<File>, Error> {
...
+        },
+    };
+    let mut data = String::new();
+    file.read_to_string(&mut data)?;
+    let options: Options = serde_json::from_str(&data)?;
+    let path = options
+        .root
+        .unwrap_or(container_root_dir.as_ref().to_path_buf())
+        .join(namespace);
+    log::info!("youki root path is: {}", path.display());
```
jsturtevant, line 77: since this is specific to youki, could it be moved into the libcontainer_instance folder?

Mossaka: Good point

Mossaka: Actually this might not be specific to youki...

Mossaka: Essentially, it uses options.json to find the root path of the container runtime (e.g. in wasmtime case, the default is `/run/containerd/wasmtime`).

utam0k: Yes, I think this is not youki's spec.

jsturtevant:
LGTM

Still seeing the out of space on ubuntu 20.04: `Error: failed to create containerd container: error unpacking image: failed to extract layer sha256:a2cb83ead200443389c4237e67b65d20202ca73be5efff08d305224cd61f3313: write /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/55/fs/bin/kindnetd: no space left on device: unknown`

jprendes (approved):
LGTM :-)

crates/containerd-shim-wasm/src/sandbox/instance_utils.rs
```diff
@@ -45,6 +46,34 @@ pub fn maybe_open_stdio(path: &str) -> Result<Option<File>, Error> {
...
+    let mut data = String::new();
+    file.read_to_string(&mut data)?;
+    let options: Options = serde_json::from_str(&data)?;
+    let path = options
+        .root
+        .unwrap_or(containerd_rootdir.as_ref().to_path_buf())
+        .join(namespace);
+    log::info!("container runtime root path is: {}", path.display());
+    Ok(path)
+}
```
jprendes, line 75:
```suggestion
use ErrorKind::NotFound;
pub fn determine_rootdir(
    bundle: impl AsRef<Path>,
    namespace: String,
    rootdir: impl AsRef<Path>,
) -> std::io::Result<PathBuf> {
    let file = match File::open(bundle.as_ref().join("options.json")) {
        Ok(f) => f,
        Err(e) if e.kind() == NotFound => return Ok(rootdir.as_ref().join(namespace)),
        Err(e) => return Err(e),
    };
    let path = serde_json::from_reader::<_, Options>(file)?
        .root
        .unwrap_or_else(|| rootdir.as_ref().to_owned())
        .join(namespace);
    log::info!("container runtime root path is {path:?}");
    Ok(path)
}
```

Mossaka: Done. This is so much cleaner. Thanks!

PR #250, opened 2023-08-18
[rfc] Add simplified libcontainer instance API

jprendes:
DO NOT MERGE

The intention of this PR is to receive feedback on a proposed simplified API when using `libcontainer`.

This API introduces an `Engine` trait fulfilling a similar role to the `libcontainer`'s `Executor` trait.
To implement a new runtime, just implement the `Engine` trait. See the changes to the wasmedge shim.


crates/containerd-shim-wasm/src/libcontainer_instance/mod.rs
```diff
@@ -1,4 +1,5 @@
 pub mod container_executor;
+pub mod easy;
```
Mossaka, line 2: I am assuming that the `easy` mod is conditional compiled with `libcontainer` feature on, right?

jprendes: Yes, it is.

Mossaka: If that's the case, shall we remove LibcontainerInstance trait in favor of this one?

jprendes: At the moment this API builds upon `LibcontainerInstance`.
As such both APIs can coexist.
If we decide we want the simplified API and there's no use case for `LibcontainerInstance` we can do that too.

crates/containerd-shim-wasmedge/src/lib.rs
```diff
@@ -1,7 +1,80 @@
...
+        Self { vm: Arc::new(vm) }
+    }
+}
+
+impl Engine for WasmEdgeEngine {
+    fn name() -> &'static str {
+        ENGINE_NAME
+    }
+
+    fn run(&self, ctx: RuntimeContext, _spec: &Spec) -> anyhow::Result<i32> {
```
jsturtevant, line 45: I like how this new interface makes the shim implementors only need to know about their particular runtime and how to configure it.

crates/containerd-shim-wasm/src/libcontainer_instance/mod.rs
jsturtevant: I think we need some kind of feedback from shim creators on that type of decision (even if its just us as we add support for other runtimes). It's a little extra work to maintain both but as it is here but having the ability to fall back on something that provides more flexibility seems like a good idea at least initially.  It seems like it would be pretty easy in the future to fold `LibContainerInstace` into `Engine `.

If we moved forward with this we would have three ways to implement shims, which all build on each other.

- `Instance` - you do everything
- `LibContainerInstance` - Provides ability to run container/wasm workloads side by side. You can configure Libcontainer.  note it is Linux Only.
- `Engine` - similar to LibContainerInstance but you don't have control over the  LibContainer.

Mossaka:
Need a rebase

Mossaka:
Would you like to talk about this RFC on tomorrow's CG meeting? @jprendes

jprendes:
Yep, that's a good plan :-)

Mossaka:
Cool, I've added it to the agenda

crates/containerd-shim-wasm/src/libcontainer_instance/executor.rs
```diff
@@ -0,0 +1,78 @@
...
+use crate::sandbox::Stdio;
+
+#[derive(Clone)]
+pub(crate) struct Executor<E: Engine> {
+    pub engine: E,
+    pub stdio: Stdio,
+}
+
+impl<E: Engine> LibcontainerExecutor for Executor<E> {
+    fn validate(&self, spec: &Spec) -> Result<(), ExecutorValidationError> {
```
Mossaka, line 24: ummm it feels strange to me that `validate` is already checking either Linux binary or wasm bianry but `execute` has to check the same thing again. This time, it delegates to different runtime to exec. 

I guess this makes sense and preserves what we had before. It's just a bit efficient.

jprendes: See https://github.com/containerd/runwasi/pull/290#discussion_r1309404870
We could use internal mutability, and store the result of the check so that we don't have to repeat it.

Mossaka: Could we mutate a state in `Executor` to save the `validate` result and reuse in `execute`? 

Something like `is_linux_executable: RefCell<bool>`

Mossaka: Ah you commented a few seconds before me!

jprendes: Done :-)

crates/containerd-shim-wasmedge/src/lib.rs
jprendes: Do you think we could use a similar API for windows?
i.e., do the normal setup for a windows container, and then let the implementor run the runtime inside the container.

jprendes:
Closing in favor of #293

PR #251, opened 2023-08-18
refactor: group imports 

Mossaka:
This commits adds a rustfmt configuration to create three import groups:
1. std, core
2. external
3. local crates

and group all the imports based on these three groups. 

Changed `cargo fmt --all` to `cargo +nightly fmt --all` as this feature isn't stable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,13 +14,15 @@ jobs:
     runs-on: "ubuntu-latest"
     steps:
       - uses: actions/checkout@v3
-      - uses: Swatinem/rust-cache@v2
-      - uses: dtolnay/rust-toolchain@nightly
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           components: rustfmt, clippy
       - name: Setup build env
         run: ./scripts/setup.sh
         shell: bash
+      - run: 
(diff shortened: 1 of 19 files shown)


jprendes (approved):
I love this change! I didn't know you could configure `rustfmt` like this. 
Thanks @Mossaka !

devigned (dismissed):
So much cleaner!

Mossaka:
Question: do you all prefer imports granularity to be "crate" or "module", or something else?

See https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#imports_granularity

jprendes:
> Question: do you all prefer imports granularity to be "crate" or "module", or something else?

Module is fine for me, but either is ok. To me consistency is more important than what particular option we choose.

Mossaka:
I am going to merge this in. @jsturtevant any concerns on merging? For rebasing, I think the easier way is to copy the `rustfmt.toml` file and then ran `cargo +nightly fmt --all` before rebasing this PR.

jsturtevant:
we will see how rebasing #238 will go, might make me 😢 

>  To me consistency is more important than what particular option we choose.

💯

Mossaka:
Merging it

PR #260, opened 2023-08-21
Use an RAII wrapper for stdio redirection

jprendes:
This PR introduces an RAII wrapper to hande stdio redirection.
The wrapper handles the lifetime of the file descriptors involved.
It also introduces Windows support for stdio redirection.

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/stdio.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/stdio.rs
@@ -1,0 +1,66 @@
+use std::fs::{File, OpenOptions};
+use std::io::ErrorKind::NotFound;
+use std::io::{Error, Result};
+use std::path::Path;
+use std::sync::{Arc, Mutex};
+
+use crate::sys::stdio::*;
+
+#[derive(Default, Clone)]
+pub struct Stdio {
+    pub stdin: Stdin,
+    pub stdout: Stdout,
+    pub stderr: Stderr,
+}
+
+impl Stdio {
+    pub fn redirect(&self) -> Result<()> {
+        self.stdin.redirect()?;
+        self.stdout.redirect()?;
+        self.stderr.redirect()?;
+        Ok(())
+    }
+}
+
+macro_rules! stdio_impl {
+    ( $stdio_type:ident, $fd:expr ) => {
+        #[derive(Default, Clone)]
+        pub struct $stdio_type(Arc<Mutex<Option<File>>>);
+
+        impl<P: AsRef<Path>> TryFrom<Option<P>> for $stdio_type {
+            type Error = std::io::Error;
+            fn try_from(path: Option<P>) -> Result<Self> {
+                path.and_then(|path| match path.as_ref() {
+                    path if path.as_os_str().is_empty() => None,
+                    path => Some(path.to_owned()),
+                })
+                .map(
+                    |path| match OpenOptions::new().read(true).write(true).open(path) {
+                        Err(err) if err.kind() == NotFound => Ok(None),
+                        Ok(f) => Ok(Some(f)),
+                        Err(err) => Err(err),
+                    },
+                )
+                .transpose()
+                .map(|opt| Self(Arc::new(Mutex::new(opt.flatten()))))
+            }
+        }
+
+        impl $stdio_type {
+            pub fn redirect(&self) -> Result<()> {
+                if let Some(f) = self.0.try_lock().ok().and_then(|mut f| f.take()) {
+                    let f = try_into_fd(f)?;
+                    let _ = unsafe { libc::dup($fd) };
+                    if unsafe { libc::dup2(f.as_raw_fd(), $fd) } == -1 {
+                        return Err(Error::last_os_error());
+                    }
+                }
+                Ok(())
+            }
+        }
+    };
+}
+
+stdio_impl!(Stdin, STDIN_FILENO);
+stdio_impl!(Stdout, STDOUT_FILENO);
+stdio_impl!(Stderr, STDERR_FILENO);
--- a/crates/containerd-shim-wasm/src/lib.rs
+++ b/crates/containerd-shim-wasm/src/lib.rs
@@ -5,9 +5,9 @@
 
 pub mod sandbox;
 pub mod services;
-#[cfg_attr(unix, path = "sys/unix.rs")]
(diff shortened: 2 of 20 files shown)


crates/containerd-shim-wasm/src/sandbox/sys/unix/stdio.rs
```diff
@@ -0,0 +1,9 @@
+use std::io::Result;
+pub use std::os::fd::AsRawFd;
+use std::os::fd::OwnedFd;
+
+pub use libc::{STDERR_FILENO, STDIN_FILENO, STDOUT_FILENO};
+
+pub fn try_into_fd(f: impl Into<OwnedFd>) -> Result<impl AsRawFd> {
+    Ok(f.into())
```
Mossaka, line 8: It looks like this function will never return Error. Why don't we remove the `try` prefix?

crates/containerd-shim-wasm/src/sandbox/stdio.rs
```diff
@@ -0,0 +1,73 @@
...
+        self.stdout.redirect()?;
+        self.stderr.redirect()?;
+        Ok(())
+    }
+}
+
+macro_rules! stdio_impl {
+    ( $stdio_type:ident, $fd:expr ) => {
+        #[derive(Default, Clone)]
+        pub struct $stdio_type(Arc<Mutex<Option<File>>>);
```
Mossaka, line 32: question: why do we want a Arc on the stdio?

Mossaka (approved):
LGTM! Asked a few questions

crates/containerd-shim-wasm/src/sandbox/sys/unix/stdio.rs
jprendes: On windows this operation can fail.
It returns a result here so that we can use the same interface in both unix and windows.

crates/containerd-shim-wasm/src/sandbox/stdio.rs
jsturtevant: I believe this is because it's not clone/copy so won't work across both of the executors (https://github.com/containerd/runwasi/pull/240#issuecomment-1680836865)

jprendes: The `Stdio` struct has to be passed to the wasm executor and the linux / default executor. Both will have ownership of the files, but only one of them will end up using them.
In any case, `Arc` allows it to be `clone`d cheaply.

Mossaka: Ahh I see! Is `Rc<RefCell<T>>` good enough since we aren't dealing with multithreading?

crates/containerd-shim-wasm/src/sandbox/sys/windows/stdio.rs
```diff
@@ -0,0 +1,42 @@
...
+        return Err(Error::new(
+            ErrorKind::Other,
+            "Failed to open file descriptor",
+        ));
+    }
+    let _ = handle.into_raw_handle(); // drop ownership of the handle, it's managed by fd now
+    Ok(OwnedFd { fd })
+}
+
+pub trait AsRawFd {
```
jsturtevant, line 28: I am not sure how I feel about implementing `AsRawFd` and related for windows.  I haven't really seen this pattern in other projects.  (though my experience is limited tbh). 

This seems fairly contained to use inside the stdio package so I think this should be ok

crates/containerd-shim-wasm/src/sandbox/stdio.rs
jprendes: That might work if it doesn't break some `Sync` requirements. I'll try it tomorrow.

crates/containerd-shim-wasm/src/sandbox/sys/unix/stdio.rs
jsturtevant, line 1: I put the sys folder up at the `/src/` folder level in #238. thoughts on which level it should be on?

crates/containerd-shim-wasm/src/sandbox/sys/windows/stdio.rs
```diff
@@ -0,0 +1,42 @@
+use std::io::{Error, ErrorKind, Result};
+use std::os::windows::prelude::{AsRawHandle, IntoRawHandle, OwnedHandle};
+
+type RawFd = libc::c_int;
+
+pub static STDIN_FILENO: RawFd = 0;
+pub static STDOUT_FILENO: RawFd = 1;
+pub static STDERR_FILENO: RawFd = 2;
```
jsturtevant, line 8: for reference https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/dup-dup2?view=msvc-170

crates/containerd-shim-wasm/src/sandbox/sys/windows/stdio.rs
jprendes: I know, I thought it would be fine since it's contained to the module.
I was being lazy, so I would save writing a trait in the unix case.
But we could use a different name as well :-)

jprendes: Ah, as long as the method name `as_raw_fd` doesn't change, it should be fine.

crates/containerd-shim-wasm/src/sandbox/sys/windows/stdio.rs
jsturtevant: yes, that's what I took from it 👍  I was verifying and left a note here for future reference

Mossaka (dismissed):

Mossaka:
Need a rebase

Mossaka (approved):
LGTM! Thanks

jsturtevant:
is https://github.com/containerd/runwasi/pull/260#discussion_r1300744582 the location we want?  otherwise lgtm

jprendes:
> is [#260 (comment)](https://github.com/containerd/runwasi/pull/260#discussion_r1300744582) the location we want? otherwise lgtm

I'll fix that tomorrow. I'm AFK.

jprendes:
I moved the `stio` within the `sys` module from #238 . I also took the opportunity to split up the `sys` modules into smaller submodules.

Mossaka (approved):

PR #262, opened 2023-08-22
DO NOT MERGE: another inspect disk usage on CI

jprendes:


Mossaka:
pre-inspect failed so the `test` didn't actually run

image

jprendes:
What's interesting is that the root already starts almost full.

jprendes:
Thanks for pushing the change. All the space is being taken by `/var/lib/docker` 😮

PR #265, opened 2023-08-23
Make wasmedge preopen the container root

jprendes:
With #78 we inadvertedly stopped preopening the container root.
This means that the wasmedge shim doesn't have access to the container filesystem.
This PR reintroduces preopening the container root with wasmedge.

### DIFF
--- a/crates/containerd-shim-wasmedge/src/executor.rs
+++ b/crates/containerd-shim-wasmedge/src/executor.rs
@@ -91,7 +91,7 @@ impl WasmEdgeExecutor {
         wasi_module.initialize(
             Some(args.iter().map(|s| s as &str).collect()),
             Some(envs.iter().map(|s| s as &str).collect()),
-            None,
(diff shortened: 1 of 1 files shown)


rumpl (approved):

devigned:
In the long run, pre-opening anything may be a bad (default) choice. We need to think about how pre-opens and other access / capabilities should be described.

jprendes:
This is preopening the container root, not the host root.
@devigned what use case(s) did you have in mind for configuring preopening?

jprendes:
It's a valid point that we should consider.
I'm just struggling to understand the use case. I woud imagine that if you don't want fs access you just don't include any file in your image root (except your wasm file, until we move that to artifacts). Am I missing something?

devigned:
It's application of least privileges (capabilities) for Wasm. For example, you might run some untrusted, 3rd party Wasm app, and want the runtime sandbox to not provide access to the fs.

PR #268, opened 2023-08-24
feat: add version flag to all the shims

Mossaka:
This follows the PR https://github.com/containerd/rust-extensions/pull/167

### DIFF
--- a/crates/containerd-shim-wasmedge/src/lib.rs
+++ b/crates/containerd-shim-wasmedge/src/lib.rs
@@ -1,3 +1,7 @@
+use std::env;
+
+use containerd_shim::parse;
+
 pub mod error;
 
 #[cfg_attr(unix, path = "instance/instance_linux.rs")]
@@ -7,6 +11,19 @@ pub mod oci_utils;
 
 #[cfg(unix)]
 pub mod executor;
+
+pub fn parse_version() {
+    let os_args: Vec<_> = env::args_os().collect();
+    let flags = parse(&os_args[1..]).unwrap();
+    if flags.version {
+        println!("{}:", os_args[0].to_string_lossy());
+        println!("  Version: {}", env!("CARGO_PKG_VERSION"));
+        println!("  Revision: {}", env!("CARGO_GIT_HASH"));
+        println!();
+
+        std::process::exit(0);
+    }
+}
 
 #[cfg(unix)]
 #[cfg(test)]
--- a/crates/containerd-shim-wasmedge/build.rs
+++ b/crates/containerd-shim-wasmedge/build.rs
@@ -1,0 +1,24 @@
+use std::process::Command;
+use std::str::from_utf8;
+
+fn main() {
+    let output = match Command::new("git").arg("rev-parse").arg("HEAD").output() {
+        Ok(output) => output,
+        Err(_) => {
+            return;
+        }
+    };
+    let mut hash = from_utf8(&output.stdout).unwrap().trim().to_string();
+
+    let output_dirty = match Command::new("git").arg("diff").arg("--exit-code").output() {
+        Ok(output) => output,
+        Err(_) => {
+            return;
+        }
+    };
+
+    if !output_dirty.status.success() {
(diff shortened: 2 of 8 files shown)


jprendes:
**suggestion:** use https://crates.io/crates/git-version and remove the build scripts.

Mossaka:
https://crates.io/crates/git-version uses `git describe`, which fetches the tag information for a commit hash. It is a bit awkward to have the following revision

```bash
./target/debug/containerd-shim-wasmedge-v1:
  Version: 0.1.1
  Revision: containerd-shim-wasm/v0.1.2-278-ged85c04-modified
```

jprendes:
Could we use `compile_time_run` instead? With something like:
```rust
const fn get_rev() -> &'static str {
    compile_time_run::run_command_str!("sh", "-c", "echo $(git rev-parse --short=10 HEAD)$(git diff --quiet || echo .m)")
}
```

crates/containerd-shim-wasmedge/src/lib.rs
```diff
@@ -8,6 +12,19 @@ pub mod oci_utils;
...
+    let flags = parse(&os_args[1..]).unwrap();
+    if flags.version {
+        println!("{}:", os_args[0].to_string_lossy());
+        println!("  Version: {}", env!("CARGO_PKG_VERSION"));
+        println!("  Revision: {}", env!("CARGO_GIT_HASH"));
+        println!();
+
+        std::process::exit(0);
+    }
+}
```
jprendes, line 26: **suggestion:** create a `containerd_shim_wasm::run` function that takes the version / revision, and handles this logic.

jprendes (changes requested):

Mossaka:
> Could we use [compile_time_run](https://crates.io/crates/compile-time-run) instead? With something like:

crates/containerd-shim-wasmedge/src/lib.rs
Mossaka: Sure

Mossaka: I don't think there is a good way to handle this in containerd-shim-wasm crate, as each shim implementation still needs to read os args, parse flags and get the version / revision... Could not save much logic here.

jprendes: I meant something like:
```rust
// somewhere in containerd-shim-wasm
pub fn run<Shim: ...>(runtime_id: &str, opts: Option<Config>, version: &str, revision: &str) {
    let os_args: Vec<_> = env::args_os().collect();
    let flags = parse(&os_args[1..]).unwrap();
    if flags.version {
        println!("{}:", os_args[0].to_string_lossy());
        println!("  Version: {}", env!("CARGO_PKG_VERSION"));
        println!("  Revision: {}", env!("CARGO_GIT_HASH"));
        println!();

        std::process::exit(0);
    }
    shim::run::<Shim>(runtime_id, opts)
}
```

and
```rust
// in main.rs
use containerd_shim_wasm::run;
fn main() {
    run::<ShimCli<WasiInstance>>(
        "io.containerd.wasmtime.v1",
        None,
        env!("CARGO_PKG_VERSION"),
        env!("CARGO_GIT_HASH"),
    );
}
```

then you can remove `parse_version` from all the `lib.rs`

jprendes:
> I am a bit hesitate because it introduces an external library basically doing the same thing we want in `build.rs`, but less explictly.

Haha, the opposite for me. I prefer the extra dep than adding a `build.rs` where it's not needed.
Also I think the extra dep is clearer and is less code to maintain (it has infrequent version bumps).
On the other hand, I'm not sure if that would work on Windows due to using `sh` / `echo`.
In any case, it's just a preference, and I'm happy for you to merge the build scripts.

jprendes (approved):
It's looking good. Some comments on ongoing discussions, but nothing major.

Mossaka:
I will merge now! Thank you @jprendes

PR #267, opened 2023-08-23
Share artifacts between jobs

jprendes:
This PR shares the artifacts created during the build jobs with the e2e test jobs.
It cuts checks times from 20\~25 min to 13\~16 min.

Further improvements can be made by splitting the Ubuntu based jobs and Windows base jobs, since currently Windows builds are in the critical path.

PR #270, opened 2023-08-24
Use reusable workflows to split linux and windows builds

jprendes:
This is a continuation of #267 and further reduces the PR runtime to \~8min

### DIFF
--- a/.github/workflows/action-fmt.yml
+++ b/.github/workflows/action-fmt.yml
@@ -11,17 +11,6 @@ jobs:
   fmt:
     runs-on: ${{ inputs.os }}
     steps:
-      - name: Free Disk Space (Ubuntu)
-        uses: insightsengineering/disk-space-reclaimer@v1
-        if: runner.os == 'Linux'
-        with:
-          tools-cache: true
-          android: true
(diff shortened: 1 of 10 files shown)


.github/workflows/action-fmt.yml
```diff
@@ -0,0 +1,45 @@
...
+    inputs:
+      os:
+        required: true
+        type: string
+
+jobs:
+  fmt:
+    runs-on: ${{ inputs.os }}
+    steps:
+      - name: Free Disk Space (Ubuntu)
```
jsturtevant, line 14: do the linting and building jobs need the clean up?  might make them faster without?

jprendes: I believe that by sharing artifacts we might not need the cleanup anywhere

jprendes: By the end, the job with the highest disk usage still had 13Gb free (started with 19 Gb free), so I think it's ok to skip the space reclaiming step.

Mossaka (approved):
LGTM! Thanks for reducing the CI time significantly!

PR #273, opened 2023-08-24
chore: bump rust to 1.72

Mossaka:
There is an addition for `resolver=2` applies to the workspace, and the reason for that is to resolve this build warning message

```bash
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "crates/containerd-shim-wasmedge",
     "crates/containerd-shim-wasmtime",
 ]
+resolver = "2"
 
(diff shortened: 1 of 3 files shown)


jprendes (approved):
LGTM

jsturtevant:
LGTM

PR #279, opened 2023-08-25
Merge both existing `run_wasi_test` implementations

jprendes:
Small cleanup deduplicating code.

### DIFF
--- a/crates/containerd-shim-wasm/src/libcontainer_instance/instance.rs
+++ b/crates/containerd-shim-wasm/src/libcontainer_instance/instance.rs
@@ -5,7 +5,6 @@ use std::thread;
 
 use anyhow::Context;
 use chrono::Utc;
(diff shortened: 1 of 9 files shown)


Mossaka (dismissed):
LGTM! Love the use of generics

jsturtevant:
nice, this is similar to what I tried in https://github.com/containerd/runwasi/pull/220 but cleaner after the clean up to the instance and engine traits.

This function could be used to help with https://github.com/containerd/runwasi/pull/111 as well.

jprendes:
> Love the use of generics

Otherwise I wold have had to move `tempfile` to a full `dependency` instead of a `dev-dependency` :-)

jprendes:
green CI now :-)

Mossaka (approved):

PR #280, opened 2023-08-27
cleaning up fd to avoid thread hangup

CaptainVincent:
Adding the missing section. 
The original behavior was inside instance_linux.rs, uncertain whether it applies to Windows.


jprendes:
Why do we need to do this?
IIUC, `File` should close the fd when dropped already.
Before this change I would have expected the `File` to be dropped at either of the following:
* after a `redirect()`, as it `take()`s the value out of the `Option`.
* when the last clone of `Stdio` is dropped, as the `Arc` count reaches 0 ans drops its content.

I would say in general it's when calling `redirect`.

I don't see where the fd is leaked outside of these cases. Do you you when that happens?

CaptainVincent:
I could reproduce hang up issue by run 
`sudo ctr run --rm --runtime=io.containerd.wasmedge.v1 ghcr.io/containerd/runwasi/wasi-demo-app:latest testwasm /wasi-demo-app.wasm echo 'hello'` (wasmtime shim has same behavior)

In fact, this issue doesn't have much output information. 
Currently, I only know that the status switches to TaskStateWrapper::Exited when the task ends. 
And I'm trying to locate the first commit where this issue first appeared. 11d867a 

So I'm attempting to fill in what I thought was missing to address this issue. However, you might be right; this might not be the true root cause. I'll first fix the CI and then give more thought to your feedback.

jprendes:
I can totally see how https://github.com/containerd/runwasi/commit/11d867aee41a565e24155dfded1c7e69bf224c5f could have introduced the issue. But I don't know why.

Maybe something related to the process being cloned?

CaptainVincent:
yup, me too. 🤔
I will keep try conduct further testing to give more logical explanation for this PR.

jprendes:
> I could reproduce hang up issue by run basic demo
sudo ctr run --rm --runtime=io.containerd.wasmedge.v1 ghcr.io/containerd/runwasi/wasi-demo-app:latest testwasm /wasi-demo-app.wasm echo 'hello' (wasmtime shim has same behavior)

Does it happen every time or sporadically? I can't reproduce that locally.

CaptainVincent:
> Does it happen every time or sporadically? I can't reproduce it locally, even after 1500 iterations. What is your test environment?

What ~ really?! I ran it on our development server and it hangs every time.
Did I miss any relevant information that might be available beyond what's presented below?

- ctr version
ctr containerd.io 1.6.22

- os version
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.3 LTS
Release:        22.04
Codename:       jammy

jprendes:
I just tried in a Ubuntu 22.04 VM, but I can't reproduce.
Ubuntu 22.04 did install containerd 1.7.2.
I will try with 20.04

CaptainVincent:
In [here](https://github.com/CaptainVincent/runwasi/actions/runs/6003350103/job/16281618200), I upgrade my github runner to containerd v1.7.5 and check it inside workflow again
```
Run ctr --version
ctr github.com/containerd/containerd v1.7.5
```
But no luck 🥲

jprendes:
Ok, I see what's happening.
The `redirect()` method gets called in a different process.
This means that the `Stdio` copy in original process' `Instance` never gets consumed.

PR #287, opened 2023-08-29
Fix dangling fd with stdio

jprendes:
With #260 the stdio `fd` are dangling in the parent shim process.
Normally redirecting the `stdio` would clear the `fd`s.
However, the redirection happens in the `clone3`d child process.
This means that the parent process maintains a dangling cope of the ´fd´s, keps alive by the shim `Instance`s.

The key changes in this PR are:
* `Stdio` now offers a `take` method.
* The `Instance` impls now use `stdio.take()` for the last executor, giving up ownership of the `fd`s.
* Adds a smoke test that used to hang without this change.

The rest of the changes are general fixes:
* Change the `Muxtex` in `Stdio` to an `AtomicCell` ~~(we need two nested ones to be able to implement ´take´)~~
* Add timeouts to the tests in CI
* Update the release workflow, which was using the old path of a moved script
* Replace the macro in `Stdio` with a generic const.
* implement `TryFrom<&InstanceConfig<Engine>>` for `Stdio`

### DIFF
--- a/.github/workflows/action-test-smoke.yml
+++ b/.github/workflows/action-test-smoke.yml
@@ -1,4 +1,4 @@
-name: Run end to end tests on k3s
+name: Run smoke tests
 
 on:
   workflow_call:
@@ -11,8 +11,8 @@ on:
         type: string
 
 jobs:
-  e2e-k3s:
-    name: e2e k3s test on ${{ inputs.os }}
+  smoke-test:
+    name: smoke test on ${{ inputs.os }}
     runs-on: ${{ inputs.os }}
     steps:
       - uses: actions/checkout@v3
@@ -36,7 +36,9 @@ jobs:
           path: dist
       - name: run
         timeout-minutes: 5
-        run: make test/k3s
-      - name: cleanup
-        if: always()
-        run: make test/k3s/clean
+        run: |
+          ls -alh dist
+          ls -alh dist/bin
+          make load
+          sudo cp -f dist/bin/* /usr/local/bin
+          sudo ctr run --rm --runtime=io.containerd.${{ inputs.runtime }}.v1 ghcr.io/containerd/runwasi/wasi-demo-app:latest testwasm /wasi-demo-app.wasm echo 'hello'
--- a/crates/containerd-shim-wasm/src/sandbox/stdio.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/stdio.rs
@@ -2,8 +2,11 @@ use std::fs::File;
 use std::io::ErrorKind::NotFound;
 use std::io::{Error, Result};
 use std::path::Path;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
+use crossbeam::atomic::AtomicCell;
+
+use super::InstanceConfig;
 use crate::sys::stdio::*;
 
 #[derive(Default, Clone)]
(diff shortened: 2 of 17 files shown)


jprendes:
@CaptainVincent , this is an alternative fix to #280.

.github/workflows/action-test-smoke.yml
```diff
@@ -0,0 +1,44 @@
...
+    inputs:
+      os:
+        required: true
+        type: string
+      runtime:
+        required: true
+        type: string
+
+jobs:
+  e2e-k3s:
```
Mossaka, line 14: Why is this called `e2e-k3s`?

jprendes: copy pasta :-)

Mossaka (dismissed):
lgtm!

jprendes:
Rebased to account for #290.
The fact that we now have only one executor simplifies things and we don't need the nested `Arc<AtomicCell<...>>`.

jprendes:
@Mossaka @jsturtevant PTAL.
If possible I would like to have this merged before the end of the week.

crates/containerd-shim-wasmedge/src/instance/instance_linux.rs
```diff
@@ -39,11 +39,7 @@ impl LibcontainerInstance for Wasi {
             )
             .unwrap(),
             exit_code: Arc::new((Mutex::new(None), Condvar::new())),
-            stdio: Stdio {
-                stdin: cfg.get_stdin().try_into().unwrap(),
-                stdout: cfg.get_stdout().try_into().unwrap(),
-                stderr: cfg.get_stderr().try_into().unwrap(),
-            },
+            stdio: cfg.try_into().expect("failed to open stdio"),
```
jsturtevant, line 42: This is mostly just readability but this looks a bit strange to me since config is a lot more than just stdio.

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -83,23 +83,17 @@ impl WasmtimeExecutor {
...
-        let module_name = match module_name {
-            Some(m) => m,
-            None => {
-                return Err(anyhow::format_err!(
-                    "no module provided, cannot load module from file within container"
-                ))
-            }
-        };
+        let module_name = module_name
+            .context("no module provided, cannot load module from file within container")?;
```
jsturtevant, line 94: ah this makes it so much cleaner!

crates/containerd-shim-wasm/src/sandbox/stdio.rs
```diff
@@ -14,51 +17,73 @@ pub struct Stdio {
...
 impl Stdio {
-    pub fn redirect(&self) -> Result<()> {
+    pub fn redirect(self) -> Result<()> {
         self.stdin.redirect()?;
         self.stdout.redirect()?;
         self.stderr.redirect()?;
         Ok(())
     }
+
+    pub fn take(&self) -> Self {
```
jsturtevant, line 27: do I understand properly that it is this `take()` that transfers the ownership?

jsturtevant (commented):
LGTM, few comments to help my understanding

crates/containerd-shim-wasmedge/src/instance/instance_linux.rs
jprendes: This does not consume the `cfg` since it's taken by reference.
Do you prefer as it was before?
Maybe something like `Stdio::init_from_config(cfg)` would be better?

crates/containerd-shim-wasm/src/sandbox/stdio.rs
jprendes: Yes, this is the `take` called that transfers the ownership.
It retuns a new `Stdio` and leaves behing a... `None`-ish `Stdio`.
Also now doing `redirect()` consumes the object, forcing you to `take()` it first (`clone()` would also work).

Mossaka (approved):

crates/containerd-shim-wasmedge/src/instance/instance_linux.rs
Mossaka: I think `Stdio::init_from_config(cfg)` would look better

Mossaka (approved):

PR #289, opened 2023-08-29
bump to libcontainer latest commit

yihuaf:
- Update to the latest executor interface
- Implement our own binary validation logic. libcontainer no longer has a default validation logic.


jprendes:
Needs rebase

PR #290, opened 2023-08-29
bump libcontainer to tip of tree

yihuaf:
- fixed all the api breakage
- implemented new executor validation logic

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ thiserror = "1.0"
 libc = "0.2.147"
 oci-spec = { version = "0.6.1", features = ["runtime"] }
 sha256 = "1.4.0"
-libcontainer = { version = "0.1", default-features = false }
+libcontainer = { git = "https://github.com/containers/youki", rev = "09e67372a892f22a89eeef62ff429c3cbcac6d41", default-features = false }
 windows-sys = { version = "0.48" }
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ thiserror = "1.0"
 libc = "0.2.147"
 oci-spec = { version = "0.6.1", features = ["runtime"] }
 sha256 = "1.4.0"
+# TODO: once lincontainer releases 0.2, switch to released version. The current commit is the tip of the tree from `youki` and a release candidate.
 libcontainer = { git = "https://github.com/containers/youki", rev = "09e67372a892f22a89eeef62ff429c3cbcac6d41", default-features = false }
 windows-sys = { version = "0.48" }
 
--- a/crates/containerd-shim-wasm/src/libcontainer_instance/container_executor.rs
+++ b/crates/containerd-shim-wasm/src/libcontainer_instance/container_executor.rs
@@ -3,12 +3,12 @@ use std::io::Read;
 use std::path::PathBuf;
 
 use libcontainer::workload::default::DefaultExecutor;
-use libcontainer::workload::{Executor, ExecutorError};
+use libcontainer::workload::{Executor, ExecutorError, ExecutorValidationError};
 use oci_spec::runtime::Spec;
 
 use crate::sandbox::{oci, Stdio};
 
-#[derive(Default)]
+#[derive(Clone)]
 pub struct LinuxContainerExecutor {
     stdio: Stdio,
     default_executor: DefaultExecutor,
@@ -18,7 +18,7 @@ impl LinuxContainerExecutor {
     pub fn new(stdio: Stdio) -> Self {
         Self {
             stdio,
-            ..Default::default()
+            default_executor: DefaultExecutor {},
         }
     }
 }
@@ -32,7 +32,19 @@ impl Executor for LinuxContainerExecutor {
         self.default_executor.exec(spec)
     }
 
+    fn validate(&self, spec: &Spec) -> Result<(), ExecutorValidationError> {
+        self.default_executor.validate(spec)?;
+
+        self.can_handle(spec)
+            .then_some(())
+            .ok_or(ExecutorValidationError::InvalidArg)
+    }
+}
+
+impl LinuxContainerExecutor {
     fn can_handle(&self, spec: &Spec) -> bool {
+        // TODO: some of these logic is now implemented inside the default
+        // executor. Should we de-dup?
         let args = oci::get_args(spec);
 
         if args.is_empty() {
@@ -105,9 +117,5 @@ impl Executor for LinuxContainerExecutor {
                 false
             }
         }
-    }
-
-    fn name(&self) -> &'static str {
-        self.default_executor.name()
     }
 }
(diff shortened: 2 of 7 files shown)


yihuaf:
@jprendes PTAL. This turned out to be a little more code change (wider impact) than I anticipated. Let me know if you would like me to make any change.

crates/containerd-shim-wasm/src/libcontainer_instance/container_executor.rs
```diff
@@ -32,7 +32,19 @@ impl Executor for LinuxContainerExecutor {
...
+
+        self.can_handle(spec)
+            .then_some(())
+            .ok_or(ExecutorValidationError::InvalidArg)
+    }
+}
+
+impl LinuxContainerExecutor {
     fn can_handle(&self, spec: &Spec) -> bool {
+        // TODO: some of these logic is now implemented inside the default
```
yihuaf, line 46: We should address this TODO before merging.

yihuaf:
One more question, what is the expected behavior of wasm executor when a regular OCI container is passed in? The current implementation will try to check if the container is a wasm container. If not wasm, then we use the default executor to validate or exec. Let me know if this assumption doesn't hold.

Cargo.toml
```diff
@@ -33,7 +33,7 @@ thiserror = "1.0"
 libc = "0.2.147"
 oci-spec = { version = "0.6.1", features = ["runtime"] }
 sha256 = "1.4.0"
-libcontainer = { version = "0.1", default-features = false }
+libcontainer = { git = "https://github.com/containers/youki", rev = "09e67372a892f22a89eeef62ff429c3cbcac6d41", default-features = false }
```
Mossaka, line 36: Could you please add a comment saying 

"the tip of the main branch. TODO: once lincontainer releases 0.2, switch to released version"

crates/containerd-shim-wasmedge/src/executor.rs
```diff
@@ -98,6 +100,27 @@ impl WasmEdgeExecutor {
...
+                return Err(ExecutorValidationError::CantHandle(EXECUTOR_NAME));
+            }
+        };
+        let path = PathBuf::from(module_name);
+
+        path.extension()
+            .map(|ext| ext.to_ascii_lowercase())
+            .is_some_and(|ext| ext == "wasm" || ext == "wat")
+            .then_some(())
+            .ok_or(ExecutorValidationError::InvalidArg)?;
```
jprendes, line 120: IIUC, this will make it so that if the entrypoint doesn't end in `wasm` or `wat` then the container fails, rather than attempt to run a native container.
I think the logic was inverted before. We first tried to execute as a native container, and if that failed, we tried with as a wasm container.

crates/containerd-shim-wasm/src/libcontainer_instance/container_executor.rs
jprendes: I think it's ok to completely remove this function now.

crates/containerd-shim-wasmedge/src/executor.rs
```diff
@@ -98,6 +100,27 @@ impl WasmEdgeExecutor {
 
         Ok(vm)
     }
+
+    fn can_handle(&self, spec: &Spec) -> Result<(), ExecutorValidationError> {
```
jprendes, line 104: This doens't use self. It coud be a free function.

jprendes: Same applies to prepare if we move `self.stdio.redirect()?` outside the function

crates/containerd-shim-wasmedge/src/executor.rs
```diff
@@ -22,45 +24,45 @@ impl WasmEdgeExecutor {
...
+                Ok(())
+            }
+            Err(_) => Err(ExecutorError::InvalidArg),
+        }
     }
 
-    fn name(&self) -> &'static str {
-        EXECUTOR_NAME
+    fn validate(&self, spec: &Spec) -> std::result::Result<(), ExecutorValidationError> {
+        match self.can_handle(spec) {
```
jprendes, line 57: **nit:** I'm not a huge fan of doing this check twice, here and on `exec`. But it works.
I'll try to think if we can avoid it or not.

jprendes (approved):

crates/containerd-shim-wasmedge/src/executor.rs
yihuaf: This is one of the downside/trade off of this interface. In both valid and exec, we have to run the `can_handle` logic to figure out the correct exec. Also due to rust safety features, in `exec` we have to run a lot of the similar checks to return `InvalidArg` error. For example, if there is no process block in the spec, we have to check it again inside `exec` even though we already checked in the `valid`. It doesn't make sense for us to `panic`.

crates/containerd-shim-wasmedge/src/executor.rs
yihuaf: I will move the `can_handle` to free function. I will leave the `prepare` for future PR.

Mossaka (dismissed):
lgtm!

yihuaf:
> Shall I merge or are you working on pushing more commits? @yihuaf

I have one more to address some of the nits and suggestion. I will let you know when this is ready.

yihuaf:
@Mossaka Once CI is green, this is ready for merge.

jprendes (approved):
LGTM! Thanks @yihuaf !

Mossaka (approved):

crates/containerd-shim-wasmtime/src/executor.rs
```diff
@@ -25,47 +27,42 @@ impl WasmtimeExecutor {
...
+                log::info!("calling start function");
+                match f.call(&mut store, &[], &mut []) {
+                    Ok(_) => std::process::exit(0),
+                    Err(_) => std::process::exit(137),
+                };
             }
-        };
-        let path = PathBuf::from(module_name);
+            Err(ExecutorValidationError::CantHandle(_)) => {
+                LinuxContainerExecutor::new(self.stdio.clone()).exec(spec)?;
```
Mossaka, line 48: I realized this changes the behavior of previous runwasi's mechanism. Previously, wasmtime shim first checks if container is a Linux binary, and then check if it's a wasm container. 

This logic seems to first check if container is a wasm binary, and then a Linux container.

jprendes: That's correct.
That's fine in the wasmedge and wasmtime cases where entrypoint should point to wasm files.
For spin/slight/wws the implementation of the executor should be somewhat different.
If you want take a look at the executor in #250, which keeps the old behaviour we has before this PR.

jprendes: The executor validation of #250 is [here](https://github.com/containerd/runwasi/pull/250/files#diff-25efb300a63bf9e9ac46c7a4f5ed76d477d9ab007003c2d0f808523c3f870085R24-R34)

Mossaka: I had a discussion with @cpuguy83 before, and we reached a conclusion that there isn't a robust way to check a wasm text file (wasm binary file could be checkd by verifying its magic number).

jprendes: I agree that the current check (i.e., the module has `.wasm` or `.wat` extension is not robust, but I expect it to work in most cases with wasmedge / wasmtime. But I do think we should find a better way.
For wasmedge / wasmtime, once the module is an OCI artifact the discussion will be different.

PR #281, opened 2023-08-28
add wasmer

dierbei:
Add support for wasmer.

Fix #275

PR #293, opened 2023-09-04
Provide a simplified API when using `libcontainer`

jprendes:
When using `libcontainer` many of the functionalities from the `sandbox::Instance` trait can be delegated to the actual OS container. This can be seen in the current implementation of the shims that build on top of `libcontainer`, where the same code pattern is seen in all of them.

This PR attempts to capture those patterns and provide a simplified API when wasm container is running inside of an OS container.

This PR introduces an `Engine` trait. This trait is very similar to `libcontainer`'s `Executor` trait, but intends to be OS agnostic.
See the `WasmEdgeEngine` and `WasmTimeEngine` for reference implementations.

The name `Engine` is chosen because this type eventually becomes the `sandbox::Instance::Engine` type in the `Instance` implementation.

For reference, these are the changes to adopt this API in `containerd-wasm-shim`: https://github.com/deislabs/containerd-wasm-shims/compare/main...jprendes:containerd-wasm-shims:runwasi-simple-api

This PR superseeds #250


README.md
```diff
@@ -48,6 +57,23 @@ pub trait Instance {
...
 ```
 
+The `container::Engine` trait provides a simplified API:
+
+```rust
+pub trait Engine: Clone + Send + Sync + 'static {
+    /// The name to use for this engine
+    fn name() -> &'static str;
+    /// Run a WebAssembly container
+    fn run(&self, ctx: impl RuntimeContext, stdio: Stdio) -> Result<i32>;
```
Mossaka, line 67: Just an opinion: I like to rename this function to `run_wasi(...)` to better describe its use cases.

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -0,0 +1,236 @@
...
+    // is "_start".
+    //
+    // If there is a '#' in the argument it will split the string returning the first part
+    // as the module name and the second part as the function name.
+    //
+    // example: "/app/module.wasm#function" will return the tuple
+    // (Path::new("/app/module.wasm"), "function")
+    //
+    // If there are no arguments then it will return (Path::new(""), "_start")
+    fn module(&self) -> (&Path, &str);
```
Mossaka, line 38: When we support [wasm components](https://github.com/WebAssembly/component-model), we probably need to reexamine this function. Perhaps we will add a `fn component(&self) -> (&Path, &str)` function.

Mossaka, line 38: Instead of `(&Path, &str)`, could you please create a struct that captures these two fields? Something like

```
struct WasiModule { path: Path, function: String }
```

crates/containerd-shim-wasm/src/container/executor.rs
```diff
@@ -0,0 +1,126 @@
...
+        match self.inner(spec) {
+            InnerExecutor::CantHandle => Err(ExecutorValidationError::CantHandle(E::name())),
+            _ => Ok(()),
+        }
+    }
+
+    fn exec(&self, spec: &Spec) -> Result<(), LibcontainerExecutorError> {
+        // If it looks like a linux container, run it as a linux container.
+        // Otherwise, run it as a wasm container
+        match self.inner(spec) {
```
Mossaka, line 45: Love this a lot!

Mossaka (commented):
Did my first pass of the review and mostly LGTM. Left some minor comments. 

Might need some coordination with PR #281

README.md
jprendes: Done

crates/containerd-shim-wasm/src/container/context.rs
jprendes: I like the suggestions :-)
Now we need to wait for that to land

jprendes: Done.
```rust
pub struct WasiEntrypoint {
    pub path: PathBuf,
    pub func: String,
}
```

jprendes:
I did some small refactor to the `RuntimeContext`.

The `RuntimeContext` has a lot of functionalities to deal with env-vars and `PATH`, but by the time either `can_handle` or `run_wasi` get called, `libcontianer` already populated the env-vars with the correct values, pivoted the rootfs, and did the chdir. That means that we can use the functionalities in `std` instead of the `RuntimeContext` implementations.

Something missing from `std` is resolving things w.r.t. `PATH`, so I pulled tha tinto an utility `PathResolve` trait.

Mossaka:
Need a rebase

jprendes:
Rebased. E2E failures seem to be sporadic, as re-running produces failures on a different set of nodes.

Mossaka (approved):
lgtm!

README.md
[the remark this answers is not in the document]
jprendes: Done

Mossaka (approved):

PR #299, opened 2023-09-07
Reset stdio in tests

jprendes:
At some point we had this functionality, but at some other point it broke with lots of refactoring around it. I couldn't pin point exacly where.
This PR adds that functionality to the `Stdio` mod and uses it in tests.


Mossaka (approved):
lgtm!

Mossaka:
Need a rebase

jprendes:
this commit was merged as part of #250 .

PR #304, opened 2023-09-09
Fix `Makefile` to avoid duplicated work in CI

### LINKED ISSUE #305: CI failures on runtimes are misleading
CI is slip in separate jobs for each runtime, including the build step, which invokes `make build RUNTIMES=xxx` and `make test RUNTIMES=xxx`.
However the `make build` and `make test` ignore the `RUNTIMES` argument.
This means that CI is currently building each runtime 3 times, and running their (non e2e) tests 3 times.
This also means that failures during the build or (non e2e) tests are misleading, since they may originate from a different runtime. To make things worst, the tests all have the same names and do not include the runtime name anywhere, making parsing the logs even harder.

jprendes:
Fixes #305

### DIFF
--- a/.github/workflows/action-build.yml
+++ b/.github/workflows/action-build.yml
@@ -27,17 +27,18 @@ jobs:
         with:
           rustflags: '' #Disable.  By default this action sets environment variable is set to -D warnings.  We manage this in the Makefile
       - name: Build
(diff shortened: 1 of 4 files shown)


jsturtevant:
LGTM

PR #306, opened 2023-09-10
Shim testing framework

jprendes:
This PR intends to set the bases to build what's been proposed in #111 .
It creates a new crate that wasm shim implementors can use to test their shims.


jprendes:
It turns out that the wasmer shim rootfs preopen setup is a bit more involved than initially thought.
That made the `has_default_devices` test fail.
This PR also fixes that issue.

crates/containerd-shim-wasm-test/src/modules/has_default_devices.rs
```diff
@@ -0,0 +1,21 @@
+use std::path::Path;
```
Mossaka, line 1: Why are some of the modules "wat" and some of are "rs"? For these Rust source files, do you have a step to compile them to wasm modules for testing?

crates/containerd-shim-wasm-test/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    pub fn with_start_fn(self, start_fn: impl AsRef<str>) -> Result<Self> {
+        let dir = self.tempdir.path();
+        let start_fn = start_fn.as_ref();
+
+        log::info!("setting wasi test start_fn to {start_fn:?}");
+
+        let entrypoint = match start_fn {
+            "" => "/hello.wasm".to_string(),
+            s => "/hello.wasm#".to_string().add(s),
```
Mossaka, line 83:
```suggestion
            s => format!("/hello.wasm#{s}"),
```

crates/containerd-shim-wasm-test/build.rs
```diff
@@ -0,0 +1,108 @@
...
+    let mut writer = std::fs::File::create(modules_file)?;
+
+    let paths = std::fs::read_dir(modules_dir)?;
+    for entry in paths.flatten() {
+        let src = entry.path();
+        let name = path_to_ident(&src)?.to_ascii_uppercase();
+
+        println!("cargo:rerun-if-changed={}", src.to_string_lossy());
+
+        let dst = match src
```
Mossaka, line 32: Ah I see that the build step is going to compile `wat` and `rs` files.

crates/containerd-shim-wasm-test/build.rs
```diff
@@ -0,0 +1,108 @@
...
+use lazy_static::lazy_static;
+
+fn env_path(key: impl AsRef<str>) -> Result<PathBuf> {
+    std::env::var_os(key.as_ref())
+        .map(Into::into)
+        .with_context(|| format!("failed to read env-var {}", key.as_ref()))
+}
+
+lazy_static! {
+    static ref OUT_DIR: PathBuf = env_path("OUT_DIR").unwrap();
```
Mossaka, line 15: What is the OUT_DIR directory? Is it `/target`?

scripts/test-runner.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+sudo -E env LD_LIBRARY_PATH=${LD_LIBRARY_PATH} "$@"
```
Mossaka, line 2: What is this for?

Mossaka (approved):
LGTM! I love how this testing framework simplies a lot the repeated code in all three shims. Thank you for this amazing work! 

I had a few comments / questions. Not a blocker.

Mossaka:
Need a rebase

crates/containerd-shim-wasm-test/src/modules/has_default_devices.rs
jprendes: the crate's buildscript compiles the files based on the extension

crates/containerd-shim-wasm-test/build.rs
jprendes: it's an env-var with a folder where buildscripts can store artifacts.
it's also accesible from the crate itself.

scripts/test-runner.sh
jprendes: In `.cargo/config.toml` you can specify a test runner.
Normally you could specify `sudo -E`, and all the tests would be run with `sudo`.

In our case, it's a bit more complicated because of wasmedge.

When we run the wasmedge tests without static linking, the binary uses `LD_LIBRARY_PATH` to find the shared library.
However `sudo` likes to block `LD_LIBRARY_PATH` even whith `-E` argument, so we need to set it manually.

So this is just a script that we can specify as the test runner that does `sudo` and forwards `LD_LIBRARY_PATH`.

If you look [here](https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sandbox/testutil.rs#L37-L48) we are already doing this in `run_test_with_sudo`.

jprendes:
Rebased

Mossaka (approved):
lgtm! Thanks

Issue #266, opened 2023-08-23
Pre-open directories and other capabilities


devigned:
In this [Slack conversation](https://cloud-native.slack.com/archives/C04LTPB6Z0V/p1692799401937929) and #265, we have started to discuss the fact that we are pre-opening the container root filesystem (fs) to the Wasm guest application. Pre-opening the root fs for use by the guest Wasm app is intended to enable users to build Wasm apps that feel more like the apps they know and love running in containers.

The default for many Wasm runtimes is to offer the guest no capabilities. This least privileged approach is one key features of Wasm.

We may want to reconsider pre-opening the container root fs in the future. Perhaps, we can replace the default behavior with a user specified behavior, opt'ing in to the pre-open. Additionally, we may also want to consider how other capabilities could be expressed by the user.

jsturtevant:
discussion with some historical context in https://cloud-native.slack.com/archives/C04LTPB6Z0V/p1692799401937929

Is this another area where having specific wasi runtime configuration might be helpful?

Mossaka:
related: https://github.com/spinkube/containerd-shim-spin/issues/108

see @jsturtevant comment at the bottom

Mossaka:
This might be shim specific and we want to provide some guidance on how to do it

PR #307, opened 2023-09-11
feat: Support wasmedge plugins

CaptainVincent:
It's a feature PR

By the way, please note that this feature currently only functions when dynamically linking the WasmEdge library to the WasmEdge shim.

For building this, you can use the following command:
`cargo build --package containerd-shim-wasmedge --no-default-features`

In addition, I have provided all the WasmEdge runtime demo cases via containerd ([here](https://github.com/CaptainVincent/runwasi/actions/runs/6150982952)), including the `wasinn plugin` demo related to this feature. Considering that these demos may only be supported with the WasmEdge shim, I plan to relocate and maintain them under the WasmEdge or Second State Org's repository. At the same time, I will ensure that they run demos daily in CI and guarantee that these demos are always executed based on the latest upstream code.

### DIFF
--- a/crates/containerd-shim-wasmedge/src/instance.rs
+++ b/crates/containerd-shim-wasmedge/src/instance.rs
@@ -15,13 +15,8 @@ pub struct WasmEdgeEngine {
 
 impl Default for WasmEdgeEngine {
     fn default() -> Self {
-        PluginManager::load(None).unwrap();
-
         let host_options = HostRegistrationConfigOptions::default();
         let host_options = host_options.wasi(true);
-        #[cfg(all(target_os = "linux", feature = "wasi_nn", target_arch = "x86_64"))]
-        let host_options = host_options.wasi_nn(true);
-
         let config = ConfigBuilder::default()
             .with_host_registration_config(host_options)
             .build()
@@ -59,6 +54,8 @@ impl Engine for WasmEdgeEngine {
             None => "main".to_string(),
         };
 
+        PluginManager::load(None)?;
+        let vm = vm.auto_detect_plugins()?;
         let vm = vm
             .register_module_from_file(&mod_name, &path)
             .context("registering module")?;


crates/containerd-shim-wasmedge/src/executor.rs
```diff
@@ -118,6 +125,28 @@ fn env_to_wasi(spec: &Spec) -> Vec<String> {
     env.to_vec()
 }
 
+fn genereate_preopen(spec: &Spec) -> Vec<String> {
+    let mut preopens: Vec<String> = vec![];
+    preopens.push("/:.".to_string());
```
jsturtevant, line 130: Why is this moving back to CWD? 
 https://github.com/containerd/runwasi/commit/20900fc26c9f572d35c4368d50a0122ad4b413b6.

CaptainVincent: You are correct. This feature has been delayed for a while, so I have rebased to the older version of the code. 
And I have noticed that the architecture seems to have changed again. I will fix it when I rebase it done.

CaptainVincent: recover done.

crates/containerd-shim-wasmedge/src/instance_linux.rs
```diff
@@ -51,14 +50,17 @@ impl Engine for WasmEdgeEngine {
             .initialize(
                 Some(args.iter().map(String::as_str).collect()),
                 Some(envs.iter().map(String::as_str).collect()),
-                Some(vec!["/:/"]),
+                Some(preopens.iter().map(|s| s as &str).collect()),
```
Mossaka, line 53: I am a bit concerned about enabling preopens by default for wasm modules, see discussions in #266. Is this required to support WasmEdge plugins?

Mossaka (commented):
Thanks! Added a few comments.

crates/containerd-shim-wasmedge/src/instance_linux.rs
jprendes: we are already preopening the whole root.
I don't think this change is doing what it intended to do, as any path in the array will be taked as a path **inside** the container (as by this point libcontainer will have pivoted the container root).

jprendes:
Where are the plugins supposed to be found?
In the host? Or in the container image?

If inside the container image, I don't see why we need to fiddle with the mount points, as libcontainer will have already handled that for us.
If it's in the host, by this point it's too late and the process doesn't have access to any path ouside the container root.

crates/containerd-shim-wasmedge/src/instance_linux.rs
CaptainVincent: The changes were not essential for supporting the plugin; they were made solely to pass the model file and image file to WASM for reading during the execution of our example use wasinn plugin.

I believe @jprendes 's  opinion is correct; if we are already directly binding the root to WASI, it should meet my requirements. 
> The initial version was implemented as a legacy when preopen was different from the current implementation. At that time, the legacy version of preopen used the root path in oci spec, not the container's root. Later, the readonly flag was added, so all of preopen code was included in this PR. However, I agree that the current architecture meets my requirements.

CaptainVincent:
> Where are the plugins supposed to be found? In the host? Or in the container image?
> 
> If inside the container image, I don't see why we need to fiddle with the mount points, as libcontainer will have already handled that for us. If it's in the host, by this point it's too late and the process doesn't have access to any path ouside the container root.

The plugin is a .so file loaded through `dlopen`, and it is not specifically restricted to running within either the host or the container. As mentioned earlier, this `mount` aspect belongs to legacy code to handle filesystem for wasm, and I will remove it and modify the way my example is used.

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -20,6 +20,11 @@ pub trait RuntimeContext {
     //   "my_module.wat" -> { path: "my_module.wat", func: "_start" }
     //   "#init" -> { path: "", func: "init" }
     fn wasi_entrypoint(&self) -> WasiEntrypoint;
+
+    // ctx.mounts() is a wrapper function
+    // Used to keep the ability for the runtime to determine how to handle the folder
+    // that being bound into the container.
+    fn mounts(&self) -> &Option<Vec<Mount>>;
```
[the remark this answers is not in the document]
CaptainVincent: Remove done.

crates/containerd-shim-wasmedge/src/instance_linux.rs
CaptainVincent: Remove done.

CaptainVincent:
The statically linked version is also poised to support plugin functionality. However, the changes made rely solely on the Wasmedge library. Therefore, after merging this PR, upstream will only need to await the release of the new SDK version to enable it in both ways.

jprendes (approved):
LGTM!

jsturtevant:
LGTM

devigned (approved):
lgtm! Thank you, @CaptainVincent.

PR #312, opened 2023-09-13
Fix ci parallel test 

### LINKED ISSUE #311: Fix `ci` parallel test issue
The current test looks like the following:
https://github.com/containerd/runwasi/blob/7b2adf09dfc9d1ed99988bc83c71a4c58f236344/Makefile#L72-L82
https://github.com/containerd/runwasi/blob/7b2adf09dfc9d1ed99988bc83c71a4c58f236344/crates/containerd-shim-wasmedge/src/tests.rs#L71-L73

The use of `--test-threads=1` and `#[serial] ` ensures that tests do not interfere with each other.

We would like to remove `--test-threads=1`, `#[serial] `, and still have ci pass successfully.

dierbei:
Fix #311

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,32 +162,32 @@ Shared mode requires precise control over real threads and as such should not be
 
 #### Components
 
-- **containerd-shim-[ wasmedge | wasmtime ]-v1**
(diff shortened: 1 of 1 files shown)


Mossaka:
Need a rebase

PR #313, opened 2023-09-13
rfc: add proc-macro to implement shims

jprendes:
This PR introduces two proc-macros for writing shims based on the simplified container API.
* The `main` annotation macro is similar to `tokio::main`. If uses the annotated main function as body of the `run_wasi` method in the `Engine` trait and generates all the surrounding boilerplate around it, including a suitable `main` function.
* The `validate` annotation lets you override the validation step, i.e., the `can_handle` method in the `Engine` trait.

The generated `main` function uses (lowercased) `argv[0]` to decide to run as a standalone shim for `containerd-shim-{id}-v1`, as a a shim daemon for `containerd-{id}d`, or as a client of the daemon for `containerd-shim-{id}d-v1`.

The `main` attribute also takes a string argument which the display name of the runtime. This (lowercased) argument is used wherever an id is required (e.g., as the `name` method in `Engine`, as the shim `id`, and in the `argv[0]` checking. This argument is also used to expose the `Engine` and `Instance` structs in the calling scope, which is useful for testing, e.g., `#[containerd_shim_wasm::main("MyRuntime")]` would expose `MyRuntimeEngine` and `MyRuntimeInstance`. The supplied string must form a valid identifier.

The functions annotated with `main` and `validate` follow the `run_wasi` and `can_handle` methods of the `Engine` trait. Their inputs are the same as their `Engine` counterparts (except the `self` argument). Their output however are a bit more permisive:
* The annotated `main` accepts `()`, `i32`, `Resutl<T, impl Into<anyhow::Error>>` (with `T` an accepted output type).
* The annotated `validate` accepts `()`, `bool`, `Result<T, impl Into<anyhow::Error>>` (in the last two cases, with `T` an accepted output type).

If the `tokio` feature is enabled, these methods can also return a `Future<Output = T>` (with `T` an accepted output type), and a `tokio` runtime will be used to execute them. This allows these methods to be `async`.

See https://github.com/deislabs/containerd-wasm-shims/compare/main...jprendes:containerd-wasm-shims:use-macro for an example on how this would be implemented in the `containerd-wasm-shims` repo.


jprendes:
Closing I  favour of #317

PR #314, opened 2023-09-14
Simple api on Windows

jprendes:
This PR adds the skeleton to make the simple API available on windows.
The actual implementation is still `todo!()`, but this change allows us to unify the implementation of the shim binaries.


jprendes:
@jsturtevant PTAL, let me know if something like this makes sense for Windows, or if there's something fundamentally wrong.

crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -42,13 +42,5 @@ ttrpc-codegen = { version = "0.4.2", optional = true }
 tempfile = "3.8"
 
 [features]
-default = []
-libcontainer_default = ["libcontainer/default"]
```
jsturtevant, line 46: This would mean that shim-implementors would be required to use `libcontainer`? Should the `Instance` api still be available for workloads that don't want to use libcontainer? Is that a scenario we should support?

jsturtevant (commented):
This looks great, This greatly simplifies the work a shim implementor will need to do to support both platforms.   I believe this will work for Windows even with our discussion around the differences between launching new process on Windows/Linux.

crates/containerd-shim-wasm/Cargo.toml
jprendes: It would be a required dependency in Linux, only.
You can still choose not to use it by using the `Instance` API instead of the `Engine` API.
The main downside of this is that we are forcing downstream dependencies to build libcontianer even if they don't use it. The linker should be smart enough to remove almost every trace of libcontainer from the final binary (except for the error definitions).

The main motivation for this is that the features system is not great with system dependent branching, and the number of combinations we have to handle grows exponentially.
But I'm open to suggestions

Cargo.toml
```diff
@@ -20,26 +20,26 @@ homepage = "https://github.com/containerd/runwasi"
 
 [workspace.dependencies]
 anyhow = "1.0"
+cap-std = "1.0"
```
Mossaka, line 23: Just curious - are you using some script to sort the deps?

PR #316, opened 2023-09-15
Fix release Dockerfile and add wasmer

jprendes:
This PR fixes an issue where the release Dockerfile wouldn't build with `BASE_IMAGE=alpine` because of the selected features.
It also adds wasmer to the the default list of crates in the release Dockerfile.

Note that adding wasmer is not strictly necessary as the release workflow manually sets the `CRATE` argument.
Fixing the issue with the features is required for the release.


Mossaka (approved):
lgtm!

PR #317, opened 2023-09-15
Factor `main`'s logic into an utility function

jprendes:
The logic of `main.rs` is always the same for all shims.
This PR moves that logic into an utility function to reduce duplication.


Mossaka (approved):
lgtm! It looks like this PR is partially extract from #313 . It really simplifies a lot!

jprendes:
rebased

Mossaka (approved):

PR #323, opened 2023-09-18
Run e2e test with both kind and k3s for all runtimes

jprendes:
This PR adds k3s e2e tests to wasmtime and kind e2e tests to wasmedge and wasmer.
It does it by adding a parameter to the e2e test in the Makefile, which indicates the runtime to use.


Mossaka (dismissed):
LGTM!

Mossaka (approved):

PR #221, opened 2023-08-01
feat: remove cgroups implementation

Mossaka:
This commit removes the cgroups implementation in `containerd-shim-wasm` as the container runtimes are now using youki's `libcontainer`.

PR #327, opened 2023-09-21
fix: avoid writing to read-only fs on talos

### LINKED ISSUE #325: Shims fail to start on talos linux
Hi,
During testing with [talos linux](https://www.talos.dev/) all shims crash during startup with the following error message:
```
Error: failed to create containerd task: failed to start shim: 
  start failed: io.containerd.wasmtime.v1: 
    Other("failed to create bind mount destination file : Read-only file system (os error 30)")
```

This error is thrown here: https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sandbox/shim.rs#L1429-L1440. When I output the value of `dest` I get `/etc/hostname`.

/cc @saiyam1814

0xE282B0:
Hi all,

Together with @DMajrekar and @saiyam1814 I've tried to run the runwasi based shims on Talos Linux. It failed because the shims tried to write to a read-only folder. As described in #325.

During testing I removed the code that caused the problem as done in this PR and I expected mounts like `/etc/hosts` and `/var/run/secrets/kubernetes.io/serviceaccount/token` to be missing but they are still there. I assume it is handled by libcontainer.

So my idea was to exclude this code when libcontainer is used, but it seems that libcontainer is always used when the platform is unix.
```rust
[target.'cfg(unix)'.dependencies].
libcontainer = { workspace = true, features = ["libseccomp", "systemd", "v1", "v2"]}
...
```

This leads me to the question of whether the code is needed at all.

@jsturtevant is this code needed on Windows?
@CaptainVincent @utam0k Is this code needed when using libcontainer?

closes  #325

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/shim.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/shim.rs
@@ -4,7 +4,7 @@
 
 use std::collections::HashMap;
 use std::env::current_dir;
-use std::fs::{self, canonicalize, create_dir_all, DirBuilder, File, OpenOptions};
+use std::fs::{self, DirBuilder, File};
 use std::ops::Not;
 #[cfg(unix)]
 use std::os::unix::fs::DirBuilderExt;
@@ -30,7 +30,7 @@ use containerd_shim::{self as shim, api, warn, ExitSignal, TtrpcContext, TtrpcRe
 use log::{debug, error};
 #[cfg(unix)]
 use nix::mount::{mount, MsFlags};
-use oci_spec::runtime::{self, Spec};
+use oci_spec::runtime::Spec;
 use shim::api::{StatsRequest, StatsResponse};
 use shim::Flags;
 use ttrpc::context::Context;
@@ -38,7 +38,6 @@ use ttrpc::context::Context;
 use super::instance::{Instance, InstanceConfig, Nop, Wait};
 use super::{oci, Error, SandboxService};
 use crate::sys::metrics::get_metrics;
-use crate::sys::networking::setup_namespaces;
 
 type InstanceDataStatus = (Mutex<Option<(u32, DateTime<Utc>)>>, Condvar);
 
@@ -338,6 +337,7 @@ mod localtests {
     use std::time::Duration;
 
     use anyhow::Context;
+    use oci_spec::runtime;
     use serde_json as json;
     use tempfile::tempdir;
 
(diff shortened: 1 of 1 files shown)


jprendes:
Is `/etc/host` et.al.? specified as a mount path in the spec? If so, libcontainer would handle that for us.
IIUC, this part of the codebase pre-dates libcontainer.
@rumpl , do you have any input?

jprendes:
IIUC, this was introduced so that DNS resolution would work inside the container: https://github.com/containerd/runwasi/commit/adb03617c5529a783092c6479abc224f02c6765e

My only concern is that it's still possible to create a shim wihout using libcontainer.
But I guess they would have to handle the mounts themselves then :-)

jsturtevant:
> @jsturtevant is this code needed on Windows?

not in its current form. I will bring back in what is needed.

> My only concern is that it's still possible to create a shim wihout using libcontainer.
But I guess they would have to handle the mounts themselves then :-)

This also happened when https://github.com/containerd/runwasi/pull/221

-----

Since it seems libcontainer does the mounting is namespace configuration also not needed?  I would it is done there too. https://github.com/containerd/runwasi/blob/ab2158ffce953b51c996e516bc61e3eaa39ba3c1/crates/containerd-shim-wasm/src/sys/unix/networking.rs#L10 not needed as well?

crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -30,15 +30,14 @@ use containerd_shim::{self as shim, api, warn, ExitSignal, TtrpcContext, TtrpcRe
...
-use oci_spec::runtime::{self, Spec};
+use oci_spec::runtime::Spec;
 use shim::api::{StatsRequest, StatsResponse};
 use shim::Flags;
 use ttrpc::context::Context;
 
 use super::instance::{Instance, InstanceConfig, Nop, Wait};
 use super::{oci, Error, SandboxService};
 use crate::sys::metrics::get_metrics;
-use crate::sys::networking::setup_namespaces;
```
jprendes, line 41: I guess you can now remove that file as it's unused, right?

0xE282B0:
@jsturtevant, yes `setup_namespaces()` is not needed as well. The function it is also used in `manager.rs`, but it makes sense to me to run the manager in its own namespace.

jprendes (approved):
LGTM!

jsturtevant:
LGTM

I will note this puts the burden the implementor if they choose not to use the libcontainer implementation provided.

jprendes:
Would this also need to be removed?
https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sandbox/shim.rs#L869-L941

devigned (approved):
lgtm!

PR #236, opened 2023-08-07
Add alpine build

jprendes:
This PR adds support for building alpine binaries.

The main changes:
* Add a build argument to the Dockerfile to specify the base image.
* Update the bake file to account for the new argument
* Move some of the logic from the Dockerfile to a shell script that handles the different base images.
* The name of the tar archives now end with `-gnu`or `-musl` (for debian and alpine respectively).
* Bumped `wasmedge-sdk` to 0.11.2 (required for building with alpine https://github.com/WasmEdge/wasmedge-rust-sdk/pull/48).
* Add a feature to vendor the dbus crate (so that it links statically)

Result:
```bash
$ docker buildx bake bins-cross tar-cross
# ... 4 parallel builds, go get a coffee ...

$ ls bin
alpine-amd64  alpine-arm64  bullseye-amd64  bullseye-arm64

$ ls bin/alpine-amd64/
containerd-shim-wasmedged-v1  containerd-shim-wasmtimed-v1  containerd-wasmedged
containerd-shim-wasmedge-v1   containerd-shim-wasmtime-v1   containerd-wasmtimed
# same output for bin/alpine-arm64, bin/bullseye-amd64, and bin/bullseye-arm64

$ ls release/
containerd-shim-wasmedge-linux-amd64-gnu.tar.gz   containerd-shim-wasmtime-linux-amd64-gnu.tar.gz
containerd-shim-wasmedge-linux-amd64-musl.tar.gz  containerd-shim-wasmtime-linux-amd64-musl.tar.gz
containerd-shim-wasmedge-linux-arm64-gnu.tar.gz   containerd-shim-wasmtime-linux-arm64-gnu.tar.gz
containerd-shim-wasmedge-linux-arm64-musl.tar.gz  containerd-shim-wasmtime-linux-arm64-musl.tar.gz

$ ldd bin/alpine-amd64/containerd-shim-wasmedge-v1 
        statically linked

$ ldd bin/bullseye-amd64/containerd-shim-wasmedge-v1 
        linux-vdso.so.1 (0x00007ffe1e1b8000)
        librt.so.1 => /usr/lib/librt.so.1 (0x00007f33f077d000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007f33f0778000)
        # ... and more ...
```

PR #330, opened 2023-09-21
release: remove extra layer in create release

Mossaka:
I noticed that the release directory structure was changed in #236 . This commits fixes that. 

It should fix the issue described in #328 

CC @jprendes


jprendes (approved):
LGTM

PR #331, opened 2023-09-22
Bump wasmedge-sdk with fixed musl issue

jprendes:
The new patch release of wasmedge-sdk fixes the issue where it fails to build with musl.

PR #332, opened 2023-09-22
Use `cross-rs`

jprendes:
This PR enables the use of `cross-rs` for building and testing.
```sh
make build-wasmedge CARGO=cross TARGET=x86_64-unknown-linux-musl
```

It also works as an example on how to set up the cross containers for building `runwasi` with `libseccomp`

This PR depends on #331 as otherwise wasmedge wouldn't build on `musl` targets.

Note that the release workflows are not using `cross-rs` which can be done in a follow up PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,19 @@
 PREFIX ?= /usr/local
 INSTALL ?= install
+CARGO ?= cargo
 LN ?= ln -sf
 TEST_IMG_NAME ?= wasmtest:latest
 RUNTIMES ?= wasmedge wasmtime wasmer
-export CONTAINERD_NAMESPACE ?= default
+CONTAINERD_NAMESPACE ?= default
 
-TARGET ?= debug
+TARGET ?= 
+OPT_PROFILE ?= debug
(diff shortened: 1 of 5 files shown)


jprendes:
We've used this to build the x86_64 and aarch64 binaries we ship with Docker Desktop.
https://github.com/rumpl/containerd-wasi-shims/blob/main/.github/workflows/build-runwasi.yml#L31

cpuguy83 (approved):

Makefile
```diff
@@ -24,92 +29,94 @@ DOCKER_BUILD ?= docker buildx build
 
 KIND_CLUSTER_NAME ?= containerd-wasm
 
+export
```
jsturtevant, line 32: typo?

jprendes: No, this means "export everything".
Since this is using nested `make` calls, I thought it would be simpler to export everything.

jsturtevant:
LGTM

devigned (approved):
lgtm!

PR #336, opened 2023-09-26
Bump nix to 1.27.1

jprendes:
`nix 1.27` doesn't enable features by default, so we need to enable them manually.
Also, libcontainer uses Nix 1.26, and the `Pid` type had changed between the two versions. To work around this, we need to go through the raw `pid_t`.

This supersedes #294


jsturtevant:
LGTM

cpuguy83 (approved):

Issue #326, opened 2023-09-20
Process crashes and hangs if invalid entrypoint provided


jsturtevant:
Using the lastest commit `commit df35a387799248b3e82b52609890d02f58a411e5`

Pass an invalid entrypoint and the process hangs for ~30 seconds eventually spitting out the error `Others("failed to receive. \"waiting for init ready\". BrokenChannel")`. During which time no logs are output to containerds log stream.

```
❯ sudo ./ctr run --rm --runtime=io.containerd.wasmtime.v1 ghcr.io/containerd/runwasi/wasi-demo-app:latest testwasm /something.exe echo

~ 30sec

ctr: Others("failed to receive. \"waiting for init ready\". BrokenChannel"): unknown
```

Running it a second time and the process hangs indefinitely. 

```
❯ sudo ./ctr run --rm --runtime=io.containerd.wasmtime.v1 ghcr.io/containerd/runwasi/wasi-demo-app:latest testwasm /something.exe echo
hangs....
```

## analysis
Was able to track that down to the failed process earlier where it didn't clean up properly ending up with 

```
bind(5, {sa_family=AF_UNIX, sun_path="/run/containerd/dfe5ace9c8dbc561.sock"}, 40) = -1 EADDRINUSE (Address already in use)
```

I've tracked this error down `"failed to receive. \"waiting for init ready\". BrokenChannel"` to https://github.com/containers/youki/blob/57ffefe89318df1cd1d1487c3cbcd4576a2d4dea/crates/libcontainer/src/process/channel.rs#L166-L173

I believe reason for the first "init ready" is somewhere in 

https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sys/unix/container/executor.rs#L87C1-L106

and it not handling the errors properly.  Not sure if this is libcontainer or runwasi handling it wrong.

jprendes:
Interesting. We probe for Linux entrypoint in https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sys/unix/container/executor.rs#L87C1-L106, as you pointed out, and if that fails, we prove for a wasm entrypoint here https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/container/engine.rs#L22-L37. If that also fails, we tell libcontainer that we can't handle that container.

jprendes:
Ok, I see a slightly different behaviour.
[...]

PR #340, opened 2023-09-27
Build libcontainer's container on Instance::new to be spec compliant

### LINKED ISSUE #158: Instance::new should return a Result type
This was discussed before: https://github.com/containerd/runwasi/pull/54#issuecomment-1403269766

IMHO we should aim to remove as many `unwrap()` calls as possible from shim "main thread", because a library should not panic easily.

_Originally posted by @ipuustin in https://github.com/containerd/runwasi/pull/142#discussion_r1238411469_

jprendes:
Fixes #158 and #326 

The issue behind #326 is that we are asking youki to **create** and **start** the container on the **start** call.
We don't do any meaningful work on the **create** call.

This missmach means that containerd is assuming that extra setup has happened successfully. but youki hasn't done it yet. When youki fails to do it at **start**, we are breaking those assumptions (e.g., being able to provide a PID for the init process).
The solution is to `build` (**create**) the youki container on the create call (i.e., the `new` method of runwasi's instance).
This also requires that we change the signature of `new` to return a `Result<Self, Error>`, which was previously mentioned in #158.

~~I haven't added any test for this here as testing it is rather involved.
The sideeffect of this issue on the first invocation is that we leave a dangling process, and on a second invocation is a hang.
I'm open to suggestions on how we could test it.~~ Thanks @jsturtevant for suggesting testing approach.


jprendes:
> this is the side affect of this change?

No, this is what I observe on linux before this change. With this change I do not observe any dangling process and no hanging.

jprendes:
> Would it be something like:

Yes, that's 100% correct, thanks!. I'll add test points for all shims.

jprendes:
I gave the test a go. The right place for it would be in the `containerd-shim-wasm` crate, as there's where the logic is implemented.

However, that test uses the `containerd-shim-wasm-test` crate. Adding that crate as a dependency creates a cycle which causes some weird interaction with types being doubly defined (e.g., there's a `crate::sandbox::Instance` and a `containerd_shim_wasm::sandbox::Instance`, and they are not the same trait for the compiler)

A solution is bringing the testing to the `containerd-shim-wasm-test` functionality to the `containerd-shim-wasm` behind a feature flag. Doing that significantly increases the number of files touched by this PR, but shouldn't add much complexity.

jprendes:
I recommend reviewing the commits separately :-)

crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -16,13 +16,16 @@ anyhow = { workspace = true }
 chrono = { workspace = true }
 git-version = "0.3.5"
 containerd-shim = { workspace = true }
+containerd-shim-wasm-test-modules = { workspace = true, optional = true }
 crossbeam = { workspace = true }
+env_logger = { workspace = true, optional = true }
```
jsturtevant, line 21: since these are under dev-dependencies are they needed here?

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
```diff
@@ -109,30 +108,20 @@ impl<E: Engine> SandboxInstance for Instance<E> {
...
-            Ok(mut container) => container.delete(true).map_err(|err| {
-                SandboxError::Any(anyhow::anyhow!(
-                    "failed to delete container {}: {}",
-                    &self.id,
-                    err
-                ))
-            })?,
+        match Container::load(container_root) {
+            Ok(mut container) => {
+                let _ = container.delete(true);
```
jsturtevant, line 121: are we losing the fact that it might fail here?

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
```diff
@@ -19,52 +20,50 @@ use crate::sys::container::executor::Executor;
...
 
-    fn new(id: String, cfg: Option<&InstanceConfig<Self::Engine>>) -> Self {
+    fn new(id: String, cfg: Option<&InstanceConfig<Self::Engine>>) -> Result<Self, SandboxError> {
         let cfg = cfg.unwrap();
         let engine = cfg.get_engine();
-        let bundle = cfg.get_bundle().unwrap_or_default().into();
+        let bundle = cfg.get_bundle().unwrap_or_default();
         let namespace = cfg.get_namespace();
         let rootdir = Path::new(DEFAULT_CONTAINER_ROOT_DIR).join(E::name());
         let rootdir = determine_rootdir(&bundle, &namespace, rootdir).unwrap();
```
jsturtevant, line 38: since we return an error here, we can bubble this error up now?

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
```diff
@@ -19,52 +20,50 @@ use crate::sys::container::executor::Executor;
...
     id: String,
+    _phantom: PhantomData<E>,
 }
 
 impl<E: Engine> SandboxInstance for Instance<E> {
     type Engine = E;
 
-    fn new(id: String, cfg: Option<&InstanceConfig<Self::Engine>>) -> Self {
+    fn new(id: String, cfg: Option<&InstanceConfig<Self::Engine>>) -> Result<Self, SandboxError> {
         let cfg = cfg.unwrap();
```
jsturtevant, line 33: since this will panic as well we could clean these up while we are here.

jsturtevant (commented):
great find!  Creating the container during the new aligns with containerd's "task_create" which makes sense. left a few minor comments

crates/containerd-shim-wasm/Cargo.toml
jprendes: For consumers of the `testing` feature of the library we need to specify it as a normal dependency as well.
All 3 dev-dependencies are also specified as optional normal dependencies.
I'm open to suggestions :smile:

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
jprendes: Indeed, I missed that.

crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -746,10 +746,10 @@ impl<T: Instance + Send + Sync> Local<T> {
         );
         InstanceData {
             instance: None,
-            base: Some(Nop::new(id, None)),
+            base: Some(Nop::new(id, None).unwrap()),
```
jsturtevant, line 749: this is safe since `Nop::new()` always returns `Ok`.  might be work a comment incase things change in future?

jsturtevant (commented):
LGTM

devigned:
@jprendes this needs a rebase, but lgtm after that.

devigned (approved):
lgtm!

PR #342, opened 2023-09-29
Remove complexity from wait API

### LINKED ISSUE #326: Process crashes and hangs if invalid entrypoint provided

### LINKED ISSUE #339: Simplify the code that waits for the container exit code
Currently to capture the exit code of a container during the wait call we have:
* a `Mutex` + `Condvar` pair in [`shim.rs`](https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sandbox/shim.rs#L43)
* a `mpsc` channel bridging between [`shim.rs`](https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sandbox/shim.rs#L1013) and [`instance.rs`](https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sandbox/instance.rs#L165)
* a `Mutex` + `Condvar` pair in [`instance.rs`](https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sandbox/instance.rs#L12)

This, specially managing the Condvar, is error prone. Specially when we consider the early returns from error handling. This is for instance the root cause of https://github.com/containerd/runwasi/issues/326.

All this complexity can probably be simplified, and/or abstracted in a way that makes it more resiliant to error situations.

jprendes:
This PR simplifies the API around the `Instance::wait` method and around `TaskState` and `InstanceData`.
Fixes #339.

Note that a side effect of this PR is that it fixes #326 but it shouldn't be considered a substitute for #340.

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/instance.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/instance.rs
@@ -1,15 +1,12 @@
 //! Abstractions for running/managing a wasm/wasi instance.
 
-use std::sync::mpsc::Sender;
-use std::sync::{Arc, Condvar, Mutex};
-use std::thread;
+use std::time::Duration;
 
 use chrono::{DateTime, Utc};
 
 use super::error::Error;
+use super::sync::WaitableCell;
 use crate::sys::signals::*;
-
-pub type ExitCode = Arc<(Mutex<Option<(u32, DateTime<Utc>)>>, Condvar)>;
 
 /// Generic options builder for creating a wasm instance.
 /// This is passed to the `Instance::new` method.
@@ -107,7 +104,9 @@ impl<Engine: Send + Sync + Clone> InstanceConfig<Engine> {
 
 /// Represents a WASI module(s).
 /// Instance is a trait that gets implemented by consumers of this library.
-pub trait Instance {
+/// This trait requires that any type implementing it is `'static`, similar to `std::any::Any`.
+/// This means that the type cannot contain a non-`'static` reference.
+pub trait Instance: 'static {
     /// The WASI engine type
     type Engine: Send + Sync + Clone;
 
@@ -128,61 +127,30 @@ pub trait Instance {
     /// This is called after the instance has exited.
     fn delete(&self) -> Result<(), Error>;
 
-    /// Set up waiting for the instance to exit
-    /// The Wait struct is used to send the exit code and time back to the
-    /// caller. The recipient is expected to call function
-    /// set_up_exit_code_wait() implemented by Wait to set up exit code
-    /// processing. Note that the "wait" function doesn't block, but
-    /// it sets up the waiting channel.
-    fn wait(&self, waiter: &Wait) -> Result<(), Error>;
-}
-
-/// This is used for waiting for the container process to exit and deliver the exit code to the caller.
-/// Since the shim needs to provide the caller the process exit code, this struct wraps the required
-/// thread setup to make the shims simpler.
-pub struct Wait {
-    tx: Sender<(u32, DateTime<Utc>)>,
-}
-
-impl Wait {
-    /// Create a new Wait struct with the provided sending endpoint of a channel.
-    pub fn new(sender: Sender<(u32, DateTime<Utc>)>) -> Self {
-        Wait { tx: sender }
+    /// Waits for the instance to finish and retunrs its exit code
+    /// This is a blocking call.
+    fn wait(&self) -> (u32, DateTime<Utc>) {
+        self.wait_timeout(None).unwrap()
     }
 
-    /// This is called by the shim to create the thread to wait for the exit
-    /// code. When the child process exits, the shim will use the ExitCode
-    /// to signal the exit status to the caller. This function returns so that
-    /// the wait() function in the shim implementation API would not block.
-    pub fn set_up_exit_code_wait(&self, exit_code: ExitCode) -> Result<(), Error> {
-        let sender = self.tx.clone();
-        let code = Arc::clone(&exit_code);
-        thread::spawn(move || {
-            let (lock, cvar) = &*code;
-            let mut exit = lock.lock().unwrap();
-            while (*exit).is_none() {
-                exit = cvar.wait(exit).unwrap();
-            }
-            let ec = (*exit).unwrap();
-            sender.send(ec).unwrap();
-        });
-
-        Ok(())
-    }
+    /// Waits for the instance to finish and retunrs its exit code
+    /// Returns None if the timeout is reached before the instance has finished.
+    /// This is a blocking call.
+    fn wait_timeout(&self, t: impl Into<Option<Duration>>) -> Option<(u32, DateTime<Utc>)>;
 }
 
 /// This is used for the "pause" container with cri and is a no-op instance implementation.
 pub struct Nop {
     /// Since we are faking the container, we need to keep track of the "exit" code/time
     /// We'll just mark it as exited when kill is called.
-    exit_code: ExitCode,
+    exit_code: WaitableCell<(u32, DateTime<Utc>)>,
 }
 
 impl Instance for Nop {
     type Engine = ();
     fn new(_id: String, _cfg: Option<&InstanceConfig<Self::Engine>>) -> Result<Self, Error> {
         Ok(Nop {
-            exit_code: Arc::new((Mutex::new(None), Condvar::new())),
+            exit_code: WaitableCell::new(),
         })
     }
     fn start(&self) -> Result<u32, Error> {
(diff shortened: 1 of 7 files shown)


crates/containerd-shim-wasm/src/sandbox/sync.rs
```diff
@@ -0,0 +1,117 @@
...
+
+struct WaitableCellImpl<T> {
+    // Ideally we would just use a OnceLock, but it doesn't
+    // have the `wait` and `wait_timeout` methods, so we use
+    // a Condvar + Mutex pair instead.
+    // We can't guard the OnceCell **inside** the Mutex as
+    // that would produce ownership problems with returning
+    // `&T`. This is because the mutex doesn't know that we
+    // won't mutate the OnceCell once it's set.
+    mutex: Mutex<()>,
```
Mossaka, line 19: Initially, it seems strange to me that the `Mutex` type isn't really guarding anything. After I read the comment, I understand that we can't guard the `OnceCell` due to the use of `&T` within the OnceCell issue. 

I am wondering if we can change it to `cell: Mutex<Option<T>>` instead of `OnceCell<T>`?

crates/containerd-shim-wasm/src/sandbox/sync.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    /// If the `WaitableCell` is empty when this guard is dropped, it will set with the result of `f`.
+    pub fn set_guard_with<R: Into<T>>(&self, f: impl FnOnce() -> R) -> impl Drop {
+        let cell = (*self).clone();
+        WaitableCellSetGuard { f: Some(f), cell }
+    }
+
+    /// Wait for the WaitableCell to be set a value.
+    pub fn wait(&self) -> &T {
+        let value = self.wait_timeout(None);
```
Mossaka, line 88: When `wait_timeout` returns a `None` value, this could be dangerous to do `unwrap_unchecked()`.

Could you please add a `assert` statement to check the value is not `None` at runtime at least?

crates/containerd-shim-wasm/src/sandbox/sync.rs
```diff
@@ -0,0 +1,117 @@
...
+            Some(dur) => cvar
+                .wait_timeout_while(guard, dur, |_| self.inner.cell.get().is_none())
+                .map(|(guard, _)| guard)
+                .unwrap(),
+        };
+        self.inner.cell.get()
+    }
+}
+
+impl<T, R: Into<T>, F: FnOnce() -> R> Drop for WaitableCellSetGuard<T, R, F> {
```
Mossaka, line 113: Could you please add some comments explaining this usage?

crates/containerd-shim-wasm/src/sandbox/sync.rs
```diff
@@ -0,0 +1,117 @@
+use std::cell::OnceCell;
+use std::sync::{Arc, Condvar, Mutex};
+use std::time::Duration;
+
+/// A cell where we can wait (with timeout) for
+/// a value to be set
+pub struct WaitableCell<T> {
```
Mossaka, line 7: I'd love to see some tests to ensure `WaitableCell<T>` behaves as expected in concurrent senarios.

crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -107,7 +104,9 @@ impl<Engine: Send + Sync + Clone> InstanceConfig<Engine> {
 
 /// Represents a WASI module(s).
 /// Instance is a trait that gets implemented by consumers of this library.
-pub trait Instance {
+/// This trait requires that any type implementing it is `'static`, similar to `std::any::Any`.
+/// This means that the type cannot contain a non-`'static` reference.
+pub trait Instance: 'static {
```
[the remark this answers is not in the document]
jprendes: This is so that we can move an `Arc<Instance>` into a thread.
We call `instance.wait()` in a thread since it's now a blocking call.

The `'static` here doesn't mean that any `Instance` variable has to be a global static variable. It does mean that you *could* if you wanted, create a global static of that type.
For example, a `String` would satisfy `'static`, as would `'static str`. However `'a str` would not.
Anything we think of as an _owned_ type, would satisfy `'static`, and `'static` references too. But non `'static` references would not.

The alternative to this is having a _shared state_ object that the `Instance` can return to signal the end of the process (similar-ish to what we had before). What I mean by _shared state_ is for example, and `mpsc::channel` or even a clone of the `WaitableCell`. Before this change the `Mutex` + `CondVar` tuple was the shared state.

I personally prefer imposing `'static` (which I think is a low price to pay) and having a simpler API, but I'd love to hear your thoughts.

crates/containerd-shim-wasm/src/sandbox/sync.rs
jprendes: The problem with that is that we wouldn't be able to return a reference to it's content. The general use case for a mutex doesn't know that once set the content can't change, and thus it's safe to share a reference to it with the world.
We can work around this issue if we require `Clone` and we return a `clone` of the type. That would work for the `(u32, DateTime)` exit result. It would make the `WaitableCell` less generic, but that might be fine, it's not like we are using it for anything else.

I am also considering whether I should just remove the `WaitableCell` and replace it with a `sync::OnceCell` from the `once-cell` crate (which is the inspiration for the `OnceCell` and `OnceLock` of the rust std lib).
Unfortunately when the std lib adopted them, they didn't include the `wait` method.
The main problem with this is that while they have a `wait` method, they don't have a `wait_timeout` method. The only place where we use timeouts in the wait is on tests, and it might be fine to just spin a thread in those cases.

crates/containerd-shim-wasm/src/sandbox/sync.rs
jprendes: Will do!

crates/containerd-shim-wasm/src/sandbox/sync.rs
jprendes: > I am wondering if we can change it to cell: Mutex<Option<T>> instead of OnceCell<T>?

In short, IIUC, no, that wouldn't solve the issue unless we require `Clone`.
But I'm happy to discuss further as I might be missing something.

crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -35,286 +36,175 @@ use shim::api::{StatsRequest, StatsResponse};
...
+            Self::Instance(i) => i.wait_timeout(t),
+            Self::Nop(i) => i.wait_timeout(t),
         }
-        self.base.as_ref().unwrap().wait(waiter)
     }
 }
 
-type EventSender = Sender<(String, Box<dyn MessageDyn>)>;
+struct InstanceData<T: Instance> {
+    instance: InstanceOption<T>,
```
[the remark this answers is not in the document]
jprendes: In some cases this is a `Nop` instance, in others this is a `T` instance.
See:
* https://github.com/containerd/runwasi/blob/32b4fba42c19b5550be2184e45626eaa16ab4760/crates/containerd-shim-wasm/src/sandbox/shim.rs#L629
* https://github.com/containerd/runwasi/blob/32b4fba42c19b5550be2184e45626eaa16ab4760/crates/containerd-shim-wasm/src/sandbox/shim.rs#L704

IIUC, this is used for the _CRI pause_ container, whatever that means.

crates/containerd-shim-wasm/src/sandbox/sync.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+impl<T> WaitableCell<T> {
+    /// Creates an empty WaitableCell.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Creates a WaitableCell with a value.
+    pub fn with_value(val: T) -> Self {
```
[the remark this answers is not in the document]
jprendes: Good catch!
Leftover from a previous life of this code, will remove it.

crates/containerd-shim-wasm/src/sandbox/sync.rs
jprendes: This type is internal to the module. It is an implementation details on how `set_guard_with` works.
I've added a few comments on this type, and more detaisl comments in `set_guard_with`.

crates/containerd-shim-wasm/src/sandbox/sync.rs
jprendes: Added unit tests

Mossaka (approved):
lgtm!

PR #345, opened 2023-10-02
[refactor] split sandbox shim

jprendes:
This PR splits the `sandbox/shim.rs` module into smaller more maintainable submodules.
In the process of doing so this PR also:
* Restricts the visibility of many symbols within the `shim` module that were publicly visible but not used outside the module.
* Simplifies the how paths (`stdin`, `stdout`, `stderr`, and `bundle`) are stored in `InstanceConfig` to avoid unnecessary convertions to `String` back and forth. This also affects (hopefully simplifies) the `stdio` module.
* Tries to minimize the use of `.unwrap()` in the `shim` module outside of tests and mutex poisoning.
* Refactors `Local`'s event handling to avoid spining a thread just to forward messages.


Mossaka:
Is this ready to review?

jprendes:
This is waiting on #342 as it builds up on top of it.

crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -1,1274 +0,0 @@
...
-
-type LocalInstances<T> = Arc<RwLock<HashMap<String, Arc<InstanceData<T>>>>>;
-
-/// Local implements the Task service for a containerd shim.
-/// It defers all task operations to the `Instance` implementation.
-#[derive(Clone)]
-pub struct Local<T: Instance + Send + Sync> {
-    engine: T::Engine,
-    instances: LocalInstances<T>,
-    events: Arc<Mutex<EventSender>>,
```
Mossaka, line 218: The `Mutex` type was removed in the `RemoteEventSender` implementation, why?

crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -1,1274 +0,0 @@
...
-
-    fn instance_exists(&self, id: &str) -> bool {
-        self.instances.read().unwrap().contains_key(id)
-    }
-
-    fn is_empty(&self) -> bool {
-        self.instances.read().unwrap().is_empty()
-    }
-
-    fn task_create(&self, req: api::CreateTaskRequest) -> Result<api::CreateTaskResponse> {
```
Mossaka, line 669: It looks like there are some significant changes to this function in the new `local.rs` code. Need a closer look into it...

crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -1,1274 +0,0 @@
...
-            pid,
-            ..Default::default()
-        };
-
-        let mut resp = api::DeleteResponse {
-            pid,
-            ..Default::default()
-        };
-
-        if let Some(ec) = i.wait_timeout(Duration::ZERO) {
```
Mossaka, line 957: I don't see this block of logic in the new code. Why is it removed?

Mossaka (commented):
Thanks for the cleaning up! I really love the refactoring work here and the new addition of `RemoteEventSender`

crates/containerd-shim-wasm/src/sandbox/shim.rs
jprendes: It was never needed to start with.
The new `EventSender` trait requires `Sync`, which means it's safe to call it's methods from different threads.
The previous `EventSender` type was also `Sync`, so it didn't need the `Mutex` either.

crates/containerd-shim-wasm/src/sandbox/shim.rs
```diff
@@ -1,1274 +0,0 @@
...
-
-        let id = opts.id.clone();
-
-        let grouping = annotations
-            .get("io.kubernetes.cri.sandbox-id")
-            .unwrap_or(&id)
-            .to_string();
-
-        #[cfg(unix)]
-        mount::<str, Path, str, str>(
```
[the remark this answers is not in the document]
jprendes: Hmm, good catch.
IIRC, the `mount` is not needed because libcontainer takes care of this for us.
I think we don't need the `setup_namespace` either, will try without it.

crates/containerd-shim-wasm/src/sandbox/shim.rs
jprendes: The main difference is that we are not mounting the non-rootfs mounts, i.e., [this for loop](https://github.com/containerd/runwasi/pull/345/files#diff-8f8704e2bb5644be97cfbd99c736cb7f432da2de20292de2b0f69e4193bd0f64L750-L752) is gone.
IIUC, libcontainer does this for us.
But I might be wrong, and would appreciate the extra pair of eyes :-)

crates/containerd-shim-wasm/src/sandbox/shim.rs
jprendes: It's not removed. This logic is captured by:
* converting the timestamps is [here](https://github.com/containerd/runwasi/pull/345/files#diff-9aeebb54e7b1d3667456ff20b60bd3416f3e44f11866f6f2821282f626177606R251).
* setting the event's s`exit_status` and `exited_at` is [here](https://github.com/containerd/runwasi/pull/345/files#diff-9aeebb54e7b1d3667456ff20b60bd3416f3e44f11866f6f2821282f626177606R258-R259)
* setting the response `exit_status` and `exited_at` is [here](https://github.com/containerd/runwasi/pull/345/files#diff-9aeebb54e7b1d3667456ff20b60bd3416f3e44f11866f6f2821282f626177606R265-R266)

jprendes:
> At the same time, I think this PR is bigger than a chore PR.

Changed the title to `[refactor]` as that should better reflect the changes in the PR.

crates/containerd-shim-wasm/src/sandbox/shim.rs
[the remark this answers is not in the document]
jprendes: The `setup_namespace` is the one that was removed in #327 [here](https://github.com/containerd/runwasi/pull/327/files#diff-8f8704e2bb5644be97cfbd99c736cb7f432da2de20292de2b0f69e4193bd0f64L1374), which I didn't notice while rebasing.

The `mount` is handled in `libcontainer` [here](https://github.com/containers/youki/blob/710cb4c636bc693a429aea32f840260f0d2b03b2/crates/libcontainer/src/rootfs/rootfs.rs#L55).

crates/containerd-shim-wasm/src/sandbox/shim.rs
jprendes: `libcontainer` mounts everything in `spec.mounts()` in [here](https://github.com/containers/youki/blob/710cb4c636bc693a429aea32f840260f0d2b03b2/crates/libcontainer/src/rootfs/rootfs.rs#L90).

Mossaka (dismissed):
Need a rebase

Mossaka (dismissed):
LGTM, I am hoping to get a second pair of eyes for reviewing. @jsturtevant @cpuguy83

cpuguy83 (approved):
LGTM, just one comment about the mod declaration. Not sure if its an issue.

cpuguy83 (commented):
LGTM, just one comment about the mod declaration. Not sure if its an issue.

crates/containerd-shim-wasm/src/sandbox/mod.rs
```diff
@@ -14,7 +14,7 @@ pub mod sync;
 pub use error::{Error, Result};
 pub use instance::{Instance, InstanceConfig};
 pub use manager::{Sandbox as SandboxService, Service as ManagerService};
-pub use shim::{Cli as ShimCli, Local};
+pub use shim::Cli as ShimCli;
```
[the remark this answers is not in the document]
cpuguy83: Ok, this was just an editor thing. It disappered on its own.

Issue #347, opened 2023-10-03
Sporadic failures in E2E CI


jprendes:
CI is sporadically failing on E2E tests, across all of them: smoke tests, kind, and k3s tests
For an example, see this run:
https://github.com/containerd/runwasi/actions/runs/6391393082/job/17346723004

jprendes:
From the containerd logs from [this CI run](https://github.com/containerd/runwasi/actions/runs/6398602640/job/17369215846), which generated [these logs artifact](https://github.com/containerd/runwasi/suites/16843375484/artifacts/962013504)
```
[INFO] starting instance: 665173b26c87393303cfdd55a68427afc5025f083f2e6a56916d11005c02c5a1
[INFO] about to build container instance: 665173b26c87393303cfdd55a68427afc5025f083f2e6a56916d11005c02c5a1
[INFO] about to REALLY build the container instance: 665173b26c87393303cfdd55a68427afc5025f083f2e6a56916d11005c02c5a1
[INFO] cgroup manager V2 will be used
[WARN] Controller rdma is not yet implemented.
[WARN] Controller misc is not yet implemented.
[time="2023-10-03T21:33:53.554067718Z" level=info msg="StartContainer for \"33e8e760bae0ce1ed34b80d16e8e0cf187b51a641a82421fc1379a8a59526365\" returns successfully"
WARN] Controller rdma is not yet implemented.
[WARN] Controller misc is not yet implemented.
[INFO] starting instance: 25633cc2ec2915d7ab31407f5312d86a271498e784f25cd9268b8d300bf76f56
[INFO] about to build container instance: 25633cc2ec2915d7ab31407f5312d86a271498e784f25cd9268b8d300bf76f56
[INFO] about to REALLY build the container instance: 25633cc2ec2915d7ab31407f5312d86a271498e784f25cd9268b8d300bf76f56
[INFO] cgroup manager V2 will be used
[WARN] Controller rdma is not yet implemented.
[WARN] Controller misc is not yet implemented.
[WARN] Controller rdma is not yet implemented.
[WARN] Controller misc is not yet implemented.
[INFO] starting instance: c2f10adebbddfc541ac0eefbeb7382b428f9bd8c87af4fd675cbf49e7cd3bc16
[INFO] about to build container instance: c2f10adebbddfc541ac0eefbeb7382b428f9bd8c87af4fd675cbf49e7cd3bc16
[...]

PR #352, opened 2023-10-10
Bump oci spec package

jsturtevant:
This allows for artifact field in the spec: https://github.com/containerd/runwasi/pull/147#discussion_r1344770351


jsturtevant:
the wasmer failure is #347

jsturtevant:
> the wasmer failure is #347

same with wasmtime kind job

Issue #322, opened 2023-09-18
Pull request jobs are flaking with k3s


jsturtevant:
Wasmer has failed in e2es in several PR's recently:

https://github.com/containerd/runwasi/pull/320
https://github.com/containerd/runwasi/pull/319
https://github.com/containerd/runwasi/pull/318

Looking at logs it doesn't give a ton of info, looks like the pods are stuck in pending state

fyi @0xE282B0 @dierbei

jprendes:
I'm not sure it's related to warmer, I think it's rather flakiness from k3s.
* [#1191](https://github.com/containerd/runwasi/actions/runs/6225493086): wasmer
* [#1190](https://github.com/containerd/runwasi/actions/runs/6225491702): wasmer
* [#1189](https://github.com/containerd/runwasi/actions/runs/6225488999): wasmer
* [#1186](https://github.com/containerd/runwasi/actions/runs/6203705560): wasmedge
* [#1185](https://github.com/containerd/runwasi/actions/runs/6203644167): wasmer
* [#1184](https://github.com/containerd/runwasi/actions/runs/6203583693): wasmedge
* [#1182](https://github.com/containerd/runwasi/actions/runs/6202982553): wasmedge
* [#1181](https://github.com/containerd/runwasi/actions/runs/6202883315): wasmer + wasmedge
* [#1175](https://github.com/containerd/runwasi/actions/runs/6191500343): wasmedge

jprendes:
I'm hoping #323 can give us further insight into the issue

jsturtevant:
Failed with wasmtime on this round: https://github.com/containerd/runwasi/actions/runs/6228704779/job/16906088261?pr=323

dierbei:
@jsturtevant @jprendes I'll take a look at that.

Right now according to my thoughts it is:
1. look at the Containerd logs in Pending state
2. if all are in Pending state, are there any conflicts (e.g. as in unit tests)

Mossaka:
Anyone tried reproducing it locally?

dierbei:
> Anyone tried reproducing it locally?

Unfortunately, I tested make test-wasmer 13 times without any problems.

My OS is ubuntu 22.04.

0xE282B0:
Hi, I noticed that the first shim with sidecar that comes up after installation is stuck. When deleting it, it starts without problems.
Often it is Wasmer because it is the first one that starts but observed it with Lunatic and WasmEdge as well.
[...]

PR #353, opened 2023-10-11
Retry bootstrapping k3s on failure

jprendes:
Sporadically k3s would fail to bootstrap, see
* missing file from install: https://github.com/containerd/runwasi/issues/322#issuecomment-1751267331
* kube-sys pods never start: https://github.com/containerd/runwasi/actions/runs/6455790697/job/17524166280

This PR tries to mitigate this by re-bootstrapping k3s when that happens.
This change uses a timeout of 40s for k3s bootstrapping. If that fails, it uninstalls k3s, reinstalls it, and tries again.
This would continue until it succeeds or the CI timeout of 5min is hit.

Note that in the case where the kube-sys pods never show up, simply restarting the systemd service is not enough (I tired this in one of the many many attempts), which indicates that a complete reinstall is a better option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,17 +161,7 @@ bin/k3s/clean:
 
 .PHONY: test/k3s-%
 test/k3s-%: dist/img.tar bin/k3s dist-%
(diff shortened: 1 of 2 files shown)


devigned (approved):
lgtm! 👏 

Thank you for digging into this issue and providing a pragmatic mitigation.

PR #355, opened 2023-10-11
Use fallback if `git_version` fails

### LINKED ISSUE #341: Issue with Git-Dependent Functions in Release Tarball
Hello,

I've encountered an issue while working with the release tarball from the [containerd/runwasi ](https://github.com/containerd/runwasi) GitHub repository, specifically the release tagged as [containerd-shim-wasmer/v0.2.0](https://github.com/containerd/runwasi/releases/tag/containerd-shim-wasmer%2Fv0.2.0).

The problem I encountered is that this release tarball contains dependencies on Git-specific functions, which made it necessary for me to manually remove these functions in order to successfully build the packages.

Here are the specifics:

    Affected Files:

    I had to remove the Git-related function "revision!()", from the following files in the release tarball:
        crates/containerd-shim-wasmedge/src/main.rs
        crates/containerd-shim-wasmtime/src/main.rs
        crates/containerd-shim-wasmer/src/main.rs

    Steps to Reproduce:

    To reproduce the issue, simply download the release tarball from the provided link and attempt to build the packages. You will encounter Git-related function dependencies.

I have Git installed on my system, as confirmed by running the following commands:
```
> which git
/usr/bin/git
```
```
> git --version
git version 2.34.1
```

I believe addressing this issue will enhance the usability and accessibility of the release tarball for a wider audience.

Thank you for your attention to this matter.

Best regards,

Arthur Ferraz

jprendes:
Fix #341

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/cli.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/cli.rs
@@ -10,7 +10,7 @@ use crate::sandbox::{Instance, Local, ManagerService, ShimCli};
 use crate::services::sandbox_ttrpc::{create_manager, Manager};
 
 pub mod r#impl {
-    pub use git_version::git_describe;
+    pub use git_version::git_version;
 }
 
 pub use crate::{revision, version};
@@ -25,12 +25,13 @@ macro_rules! version {
 #[macro_export]
 macro_rules! revision {
     () => {
-        Some($crate::sandbox::cli::r#impl::git_describe!(
-            "--match=:",
-            "--always",
-            "--abbrev=15",
-            "--dirty=.m"
-        ))
+        match $crate::sandbox::cli::r#impl::git_version!(
+            args = ["--match=:", "--always", "--abbrev=15", "--dirty=.m"],
+            fallback = "",
+        ) {
+            "" => None,
+            version => Some(version),
+        }
     };
 }
 


jsturtevant:
This is probably fine as a work around. Is there any reason to ship the source code with a simple file that includes the git commit?

jsturtevant:
I was thinking something similar to https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#_exporting_your_repository

jprendes:
Hm, I was thinking of something on the line of `include_str("../LAST_COMMIT")`. However, that means that a missing `LAST_COMMIT` file would result in a compile error. This would work fine for runwasi itself but would break dependant projects without a `LAST_COMMIT` file. It would be simpler if there was a `option_include_str` macro!

Also that approach would not reflect any local modifications that could have been done, so it could be missleading if used for debugging. We could append a `.a` to indicate the id comes from an archive and can't be 100% trusted, in a similar fashion to what we do now appending `.m` to indicate a dirty repo.

I personally prefer a `<none>` or `<unknown>` tag as is the case now.

jsturtevant:
> I personally prefer a <none> or <unknown> tag as is the case now.

This make sense to me based on your feedback above

Mossaka (approved):
lgtm!

devigned (approved):
lgtm!

PR #356, opened 2023-10-11
chore: Use latest kind and k8s release

jsturtevant:
Updating to use more recent versions of kubernetes and kind

PR #359, opened 2023-10-18
Use `cross-rs` for building in CI

jprendes:
This PR switches CI to use the `cross-rs` support introduced in #332.
This should unblock #356.

This PR uses `cross-rs` for the linux builds of the shims. It generates a `linux-musl` and a `linux-gnu` artifacts using `musl` and `glibc` respectively.

It also stops building the shims on `ubuntu-20.04`. Since the `cross-rs` build happens in a container it is not necessary to build on different OS versions. However, we keep running e2e tests in `ubuntu-20.04` to maintain test coverage on an OS running `cgroups-v1`.

The e2e tests are switched to use the `musl` build of the shims, as this would unblock #356.
Since `cross-rs` uses an older `glibc` version, it might be possible to also enable e2e tests using `glibc`, but I suggest we defer that to another PR after #356 is merged.

There are some complications around running the unit tests with `cross-rs`. This PR works around those issues by building the tests with `cross-rs`, but running them in the host. See https://github.com/containerd/runwasi/pull/359#discussion_r1364078589 for more details.

### DIFF
--- a/.github/workflows/action-build.yml
+++ b/.github/workflows/action-build.yml
@@ -9,10 +9,16 @@ on:
       runtime:
         required: true
         type: string
+      target:
+        required: false
+        type: string
+      slug:
+        required: true
+        type: string
 
 jobs:
   build:
-    name: build on ${{ inputs.os }}
+    name: build for ${{ inputs.slug }}
     runs-on: ${{ inputs.os }}
     steps:
       - uses: actions/checkout@v3
@@ -23,12 +29,16 @@ jobs:
         shell: bash
       - uses: actions-rust-lang/setup-rust-toolchain@v1
         env:
-          RUST_CACHE_KEY_OS: ${{ inputs.os }}
+          RUST_CACHE_KEY_OS: rust-cache-${{ inputs.os }}-${{ inputs.slug }}
         with:
           rustflags: '' #Disable.  By default this action sets environment variable is set to -D warnings.  We manage this in the Makefile
+      - name: Setup cross-rs
+        if: runner.os == 'Linux'
+        run: ./scripts/setup-cross.sh ${{ inputs.target }}
       - name: Build
         run: make build-${{ inputs.runtime }}
       - name: Validate docs
+        if: ${{ inputs.runtime == 'common' }}
         run: ./scripts/validate-docs.sh
       - name: Run tests
         timeout-minutes: 5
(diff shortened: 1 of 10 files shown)


.github/workflows/action-build.yml
```diff
@@ -41,13 +48,13 @@ jobs:
           make dist-${{ inputs.runtime }}
           # Check if there's any files to archive as tar fails otherwise
           if stat dist/bin/* >/dev/null 2>&1; then
-            tar -czf dist/containerd-shim-${{ inputs.runtime }}-${{ inputs.os }}.tar.gz -C dist/bin .
+            tar -czf dist/containerd-shim-${{ inputs.runtime }}-${{ inputs.slug }}.tar.gz -C dist/bin .
```
jsturtevant, line 51: why can't we continue to use `os` here?

Makefile
```diff
@@ -6,7 +6,21 @@ TEST_IMG_NAME ?= wasmtest:latest
...
+TARGET ?= $(shell uname -m)-unknown-linux-musl
+# When using `cross` we need to run the tests outside the `cross` container.
+# We stop `cargo test` from running the tests with the `--no-run` flag.
+# We then need to run the generate test binary manually.
+# For that we use `--message-format=json` and `jq` to find the name of the binary, `xargs` and execute it.
+TEST_ARGS_SEP= --no-run --color=always --message-format=json | \
+	jq -R '. as $$line | try (fromjson | .executable | strings) catch ($$line+"\n" | stderr | empty)' -r | \
+	xargs -I_ sudo -E ./_
+else
+TARGET ?= $(shell rustc --version -v | sed -En 's/host: (.*)/\1/p')
```
jsturtevant, line 20: in the case of #356 if I am running ubuntu locally then run `make test/k8s-wasmtime` it will fail.

Makefile
```diff
@@ -47,23 +61,23 @@ check: check-wasm $(RUNTIMES:%=check-%);
 
 check-common: check-wasm;
 check-wasm:
-	$(CARGO) +nightly fmt $(TARGET_FLAG) -p oci-tar-builder -p wasi-demo-app -p containerd-shim-wasm -p containerd-shim-wasm-test-modules -- --check
+	CARGO= $(CARGO) +nightly fmt -p oci-tar-builder -p wasi-demo-app -p containerd-shim-wasm -p containerd-shim-wasm-test-modules -- --check
```
jsturtevant, line 64: why is CARGO being set here?

Makefile
```diff
@@ -6,7 +6,21 @@ TEST_IMG_NAME ?= wasmtest:latest
...
+ifeq ($(CARGO),cross)
+# Set the default target as defined in Cross.toml
+TARGET ?= $(shell uname -m)-unknown-linux-musl
+# When using `cross` we need to run the tests outside the `cross` container.
+# We stop `cargo test` from running the tests with the `--no-run` flag.
+# We then need to run the generate test binary manually.
+# For that we use `--message-format=json` and `jq` to find the name of the binary, `xargs` and execute it.
+TEST_ARGS_SEP= --no-run --color=always --message-format=json | \
+	jq -R '. as $$line | try (fromjson | .executable | strings) catch ($$line+"\n" | stderr | empty)' -r | \
+	xargs -I_ sudo -E ./_
```
jsturtevant, line 18: 😮

.github/workflows/action-build.yml
jprendes: I didn't want to encode `ubuntu-XX.YY` into the artifacts since we generate a `musl` static binary that can run in any distro.

I'd also like to evolve this to be `linux-musl` and `linux-gnu`. As it stands, we are loosing coverage of `glibc` builds, and I would like to avoid that.

Makefile
jprendes: It looks like `rust-fmt` uses the `CARGO` env-var for something, and it fails when I set to `cross`.

jsturtevant: maybe a note here to call that out?

Mossaka (approved):
lgtm!

Makefile
jprendes: Ah, I see, I think that's correct.
I think we could force using `cross-rs` here.

jprendes: The latest push forces using `cross-rs` for k8s testing.

jsturtevant:
LGTM

Mossaka (approved):

PR #360, opened 2023-10-19
Use `cross-rs` for release builds

jprendes:
This is a follow up to #359 using `cross-rs` to generate the release artifacts.

For an example of the pipeline run and the generated artifacts see:
* https://github.com/jprendes/runwasi/actions/runs/6579306152 (failure is from `cargo publish`)
* https://github.com/jprendes/runwasi/releases/tag/containerd-shim-wasmedge%2Fv0.2.0


.github/workflows/release.yml
```diff
@@ -85,29 +109,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Setup buildx
-        run: docker buildx create --use
-      - name: build binaries
-        run: docker buildx bake --set *.cache-from=type=gha,scope=buildkit-release-${CRATE} --set *.cache-to=type=gha,scope=buildkit-release-${CRATE} tar-cross
```
jsturtevant, line 91: dropping this means we no longer need https://github.com/containerd/runwasi/blob/main/docker-bake.hcl and the related docker file?

jprendes: Thanks, good catch!

Mossaka (approved):
LGTM! I will try to release a new version next week following this PR merged.

PR #362, opened 2023-10-20
Fix issue running k8s and cross locally

jsturtevant:
This PR addresses two issues after #359 when running `make test/k8s-wasmtime` locally.

First, the `TARGET` is not being set properly. Since `TARGET` was already set when re-invoking `$(MAKE)` it gets the prior value. Example output:

```
 -f /home/jstur/projects/runwasi/dist/bin/containerd-shim-wasmtime-v1 ] || make install-wasmtime CARGO=cross PREFIX="/home/jstur/projects/runwasi/dist" OPT_PROFILE="debug"
make[1]: Entering directory '/home/jstur/projects/runwasi'
cross build --target=x86_64-unknown-linux-gnu -p containerd-shim-wasmtime
```

Second, cross has issues when switching targets: https://github.com/cross-rs/cross/wiki/FAQ#glibc-version-error.  I ran into this only my local set up, where the first step of cross (after fixing the above) would error out with

```
=> => naming to localhost/cross-rs/cross-custom-runwasi:x86_64-unknown-linux-musl-d9964                                                              0.0s
   Compiling libc v0.2.149
   Compiling serde v1.0.189
   Compiling thiserror v1.0.49
error: failed to run custom build command for `libc v0.2.149`

Caused by:
  process didn't exit successfully: `/target/debug/build/libc-01039b2d07357d71/build-script-build` (exit status: 1)
  --- stderr
  /target/debug/build/libc-01039b2d07357d71/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /target/debug/build/libc-01039b2d07357d71/build-script-build)
```

This also adds a `clean` target which allows for testing this scenario, otherwise several of the steps get skipped since they already compiled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,28 +6,47 @@ TEST_IMG_NAME ?= wasmtest:latest
 RUNTIMES ?= wasmedge wasmtime wasmer
 CONTAINERD_NAMESPACE ?= default
 
+# We have a bit of fancy logic here to determine the target 
+# since we support building for gnu and musl
+# TARGET must evenutually match one of the values in the cross.toml
+HOST_TARGET = $(shell rustc --version -v | sed -En 's/host: (.*)/\1/p')
+
+# if TARGET is not set and we are using cross
+# default to musl to facilitate easier use shim on other distros becuase of the static build
+# otherwise use the host target
+ifeq ($(TARGET),)
 ifeq ($(CARGO),cross)
-# Set the default target as defined in Cross.toml
-TARGET ?= $(shell uname -m)-unknown-linux-musl
+override TARGET = $(shell uname -m)-unknown-linux-musl
+else
+override TARGET = $(HOST_TARGET)
+endif
+endif
+
+# always use cross when the target is not the host target
+ifneq ($(TARGET),$(HOST_TARGET))
+override CARGO = cross
+endif
+
+ifeq ($(CARGO),cross)
+override TARGET_DIR := $(or $(TARGET_DIR),./target/build/$(TARGET)/)
 # When using `cross` we need to run the tests outside the `cross` container.
 # We stop `cargo test` from running the tests with the `--no-run` flag.
 # We then need to run the generate test binary manually.
 # For that we use `--message-format=json` and `jq` to find the name of the binary, `xargs` and execute it.
 TEST_ARGS_SEP= --no-run --color=always --message-format=json | \
 	jq -R '. as $$line | try (fromjson | .executable | strings) catch ($$line+"\n" | stderr | empty)' -r | \
+	sed -E 's|^/target|$(TARGET_DIR)|' | \
 	xargs -I_ ./scripts/test-runner.sh ./_
 else
-TARGET ?= $(shell rustc --version -v | sed -En 's/host: (.*)/\1/p')
+override TARGET_DIR := $(or $(TARGET_DIR),./target/)
 TEST_ARGS_SEP= --
 endif
+TARGET_FLAG = --target=$(TARGET) --target-dir=$(TARGET_DIR)
 
 OPT_PROFILE ?= debug
 RELEASE_FLAG :=
 ifeq ($(OPT_PROFILE),release)
 RELEASE_FLAG = --release
-endif
-ifneq ($(TARGET),)
-TARGET_FLAG = --target=$(TARGET)
 endif
 
 FEATURES_wasmedge = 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -150,7 +154,7 @@ bin/kind: test/k8s/Dockerfile
 
 # Use a static build of the shims for better compatibility.
 # Using cross defaults to x86_64-unknown-linux-musl, which creates a static build.
-test/k8s/_out/img-%: CARGO=cross
+test/k8s/_out/img-%: CARGO=cross TARGET=$(shell uname -m)-unknown-linux-musl
```
jprendes, line 157: **praise:** Good catch! Thanks for fixing it!

Makefile
```diff
@@ -193,3 +197,11 @@ test/k3s-%: dist/img.tar bin/k3s dist-%
...
 .PHONY: test/k3s/clean
 test/k3s/clean: bin/k3s/clean;
+
+.PHONY: clean
+clean:
+	rm -rf dist
+	rm -rf bin
+	$(MAKE) test-image/clean
+	$(MAKE) test/k8s/clean
+	$(MAKE) test/k3s/clean
```
jprendes, line 207: can we prefix these with `-` so that we keep going if they fail (e.g., in case they were not installed)?
https://www.gnu.org/software/make/manual/html_node/Errors.html

jprendes (changes requested):
I don't think we need a wrapper script.

We are already specifying the target in the `Makefile` with `TARGET_FLAG`.
It's just a matter of adding `--target-dir=./target/build/$(TARGET)` to `TARGET_FLAG`.

You also need to use the same `target-dir` in the `install-%` rule (IIUC, that's why CI failed).

I propose we use a new makefile variable `TARGET_DIR`, and we set it to `./target/build/$(TARGET)` when using cross or to `./target` otherwise.

Makefile
jsturtevant: Don't love that it's here and above but tried a few different ways to try this but seem to be the only way I could get Make to overload that value, since it saw empty as a set value and couldn't unset it here.

jprendes: What if you set `TARGET=` here? (i.e., an empty value)
I **think** the `TARGET ?= ...` from above should override it if its empty.

jprendes: Ok, no, `?=` doesn't override empty values.
Stackoverflow suggests replacing with
```makefile
TARGET = $(or $(TARGET),$(shell uname -m)-unknown-linux-musl)
```

jsturtevant: I tried that and it same it as Empty value not an `unset` value, and so `?=` didn't take the default and used "" instead

jsturtevant:
looks like latest failures is due to cross setting. This is working locally when I run it, with same setting 👀 

```
/usr/bin/docker run --userns host -e 'PKG_CONFIG_ALLOW_CROSS=1' -e 'XARGO_HOME=/home/runner/.xargo'
 -e 'CARGO_HOME=/home/runner/.cargo' -e 'CROSS_RUST_SYSROOT=/home/runner/.rustup/toolchains/1.72.0-x86_64-unknown-linux-gnu' 
-e 'CARGO_TARGET_DIR=/target'
```

.github/workflows/action-build.yml
```diff
@@ -44,6 +44,8 @@ jobs:
         timeout-minutes: 5
         run: |
           make test-${{ inputs.runtime }}
+        env:
+          CARGO_TARGET_DIR: /target
```
jprendes, line 48: Was this meant to be `./target`?
Same in L.60

jsturtevant: No, In CI it is failing to find the test binary with it is `./target`.  Locally it works. But I don't know that this is correct either. There are some nuances around folder layout here (we `cd` to a few folders) that is causing hickups.

jsturtevant: took different approach

Makefile
jsturtevant: using different approach

Mossaka (approved):
lgtm! I didn't run the commands locally though. Will try it out next week.

jprendes (approved):
LGTM

Makefile
```diff
@@ -5,10 +5,12 @@ LN ?= ln -sf
 TEST_IMG_NAME ?= wasmtest:latest
 RUNTIMES ?= wasmedge wasmtime wasmer
 CONTAINERD_NAMESPACE ?= default
+CARGO_TARGET_DIR?=$(PWD)/target
```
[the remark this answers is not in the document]
jsturtevant: This is no longer the case latest commits

jprendes (approved):
LGTM

Mossaka:
I am trying to get `make test/k8s-wasmtime` runnning locally...

Mossaka:
Merging it in now. The progress on my side is slow and I don't want to block the PR

PR #364, opened 2023-10-26
Add the shim to to the network NS specified

jsturtevant:
When removing the mounting code we also removed the network setup: https://github.com/containerd/runwasi/pull/327#issuecomment-1730071090

Ends up this is required as youki is only setting the Pid and User namespaces: https://github.com/containers/youki/blob/ba4a10f4a09b0360eeb38ab70a95adc986b2b9da/crates/libcontainer/src/process/container_intermediate_process.rs#L70

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,6 +230,10 @@ test/k8s/cluster-oci-%: dist/img-oci.tar bin/kind test/k8s/_out/img-oci-%
 test/k8s-%: test/k8s/clean test/k8s/cluster-%
 	kubectl --context=kind-$(KIND_CLUSTER_NAME) apply -f test/k8s/deploy.yaml
 	kubectl --context=kind-$(KIND_CLUSTER_NAME) wait deployment wasi-demo --for condition=Available=True --timeout=90s
+	# verify that we are still running after some time	
+	sleep 5s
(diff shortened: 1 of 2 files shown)


jprendes (approved):
LGTM

Makefile
```diff
@@ -230,6 +230,10 @@ test/k8s/cluster-oci-%: dist/img-oci.tar bin/kind test/k8s/_out/img-oci-%
 test/k8s-%: test/k8s/clean test/k8s/cluster-%
 	kubectl --context=kind-$(KIND_CLUSTER_NAME) apply -f test/k8s/deploy.yaml
 	kubectl --context=kind-$(KIND_CLUSTER_NAME) wait deployment wasi-demo --for condition=Available=True --timeout=90s
+	# verify that we are still running after some time	
+	sleep 5s
+	kubectl --context=kind-$(KIND_CLUSTER_NAME) wait deployment wasi-demo --for condition=Available=True --timeout=90s
```
jprendes, line 235: The second time should be a small timeout, otherwise we might be caught in the same false success when the failed pods are restarted.

jprendes (approved):
Thanks @jsturtevant , great work finding the root cause!

Mossaka (approved):
Nice find! Thanks

Issue #349, opened 2023-10-06
Add support for structured logging


jsturtevant:
If https://github.com/containerd/rust-extensions/pull/212 goes in and we adopt it we can clean up our logging story so we have more metadata for logs.

For instance, write now our logs:

```
INFO] waiting for instance: df03285ac14ced6475f603924a4d2da31314bfb0f5056322d50b22a0f90e2a92
[INFO] calling start function
[INFO] redirect stdio
[INFO] Create a Store
[INFO] loading module from file "/wasi-demo-app.wasm"
[INFO] Creating `WasiEnv`...: args ["/wasi-demo-app.wasm"], envs: Vars { .. }
[INFO] Creating `WasiEnv`...: args ["/wasi-demo-app.wasm"], envs: Vars { .. }
[INFO] Running "_start"
[[INFO] Creating `WasiEnv`...: args ["/wasi-demo-app.wasm"], envs: Vars { .. }
```

This is challenging, What which instance is `Creating `WasiEnv`...: args` for? It gets harder with more replicates more things happening.

With strucutred logging we could have something like the following so we can distinguish between multiple instances running

```
time="2023-10-06T21:49:05.767259938Z" level=info instance="aslkdfasdfhasgh09234kl23j4l2k3j42lkj34" msg="Create a Store"
time="2023-10-06T21:49:05.767259938Z" level=info instance="df03285ac14ced6475f603924a4d2da31314bfb0f5056322d50b22a0f90e2a92" msg="Creating `WasiEnv`"
time="2023-10-06T21:49:05.767259938Z" level=info instance="aslkdfasdfhasgh09234kl23j4l2k3j42lkj34" msg="Creating `WasiEnv`"
```

Mossaka:
Is this issue resolved? @jsturtevant

jsturtevant:
Not fully, the code is there to handle this but a wrapper/macro needs to be added to runwasi to add fields like ` instance="aslkdfasdfhasgh09234kl23j4l2k3j42lkj34"`

Mossaka:
This issue is currently blocked by #863

PR #370, opened 2023-10-30
chore: bump versions of the containerd-shim-wasm to `0.4` and shims to `0.3`

Mossaka:


Cargo.toml
```diff
@@ -12,7 +12,7 @@ resolver = "2"
 
 [workspace.package]
 edition = "2021"
-version = "0.2.0"
+version = "0.3.0"
```
jprendes, line 15: **question:** Should we bump this to 0.4 so that it matches the lib version? Should we always keep these two the same?

I believe there's been previous discussions on this topic before, is that correct?

jprendes (approved):

Cargo.toml
Mossaka: I don't recall a previous discussions on this topic. I am okay with either ways. 

>  Should we always keep these two the same?
Maybe not as crates and binaries can evolve independently imo.

Mossaka:
Waiting for a new release of `containerd-shim` crate https://github.com/containerd/rust-extensions/pull/216#event-10811692644 which adds timestamps for logs.

Cargo.toml
jprendes: I think there was a discussion on keeping the release of each shim, as well as the lib, independent, as their versions could diverge.

jsturtevant:
> Bump containerd-shim to 0.6.0 to include structured logging. This PR will close #349

it won't close it but enable it.  The PR will include timestamps on our logs though 🥳

jsturtevant:
LGTM

PR #373, opened 2023-10-30
chore: add published version of wasm-test-modules dep

Mossaka:
In cargo, it is possible to specify a local path and a register version for one dep (see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations). 

What this means for us is that we can keep using local `containerd-shim-wasm-test-modules` for development. At the same time, when we do `cargo publish`, it will resolve to use the remote version of `containerd-shim-wasm-test-containers` from crates.io.


jsturtevant:
LGTM

PR #374, opened 2023-10-30
ci: add runwasi-committee to owner of the crate

Mossaka:
Instead of checking if the `github:containerd:runwasi-committers` is an owner of the crate, we always add it as an owner. 

This may help to resolve the error we encountered at this run: https://github.com/containerd/runwasi/actions/runs/6699464559/job/18203943248#step:7:12

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -136,8 +136,6 @@ jobs:
       - name: Check crates.io ownership
         run: |
           cargo owner --add github:containerd:runwasi-committers ${{ needs.generate.outputs.crate }}
-          cargo owner --add ${{ github.actor }} ${{ needs.generate.outputs.crate }}
-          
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -135,9 +135,9 @@ jobs:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_PUBLISH_TOKEN }}
       - name: Check crates.io ownership
         run: |
+          cargo owner --add github:containerd:runwasi-committers ${{ needs.generate.outputs.crate }}
+          cargo owner --add ${{ github.actor }} ${{ needs.generate.outputs.crate }}
```
cpuguy83, line 139: We can probably remove this?

cpuguy83 (approved):

PR #375, opened 2023-10-31
ci: add build env to release job

Mossaka:
The release of `containerd-shim-wasm` crate failed at the cargo publish step saying that `protoc` was not found. Thus I am adding build env to the release job.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -109,6 +109,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - name: Setup build env
+        run: ./scripts/setup-linux.sh
       - name: Download artifacts
         if: ${{ needs.generate.outputs.runtime != 'wasm' }}
         uses: actions/download-artifact@master


jprendes (approved):
LGTM

jsturtevant:
LGTM

Maybe as a follow up after this release we can run the release CI up to the Publishing steps so we can catch issues earlier?

Mossaka:
> LGTM
> 
> Maybe as a follow up after this release we can run the release CI up to the Publishing steps so we can catch issues earlier?

There is a nice flag for `cargo publish --dry-run` to run all the checks without uploading to crates.io. Is this what you meant?

jsturtevant:
yes, if we can run that semi regularly then we could catch issues before release time?

Mossaka:
> yes, if we can run that semi regularly then we could catch issues before release time?

Maybe we should refactor the pipeline to an reusable action and run it at our build (with `--dry-run` flag on)

Mossaka:
@jsturtevant just realized that `--dry-run` won't be able to catch errors like https://github.com/containerd/runwasi/actions/runs/6715767237/job/18251185616#step:7:794

jprendes:
It can still catch other issues, I think it's still worth adding.

PR #376, opened 2023-10-31
chore: bump patch version of shims and add containerd-shim-wasm version

Mossaka:
Add a published version to containerd-shim-wasm dep for all the shims. Bumped the patch version so that the containerd-shim-wasm-test-modules could be published again. 

I will need to publish crates in a particular order
`containerd-shim-wasm-test-modules` -> `containerd-shim-wasm` -> shims


jprendes (approved):
LGTM

PR #383, opened 2023-11-01
remove wasmedge-sys dependency

jprendes:
This PR removes the dependency on `wasmedge-sys`.

This depdency was only used to test that a symbol is statically / dinamically linked in the binary.
This PR inlines the FFI symbol instead of using the crate dependency.

The risk of depending on `wasmedge-sys` is that the version could drifts from what `wasmedge-sdk` uses.
In that case, the test would be probing the wrong symbol, and it may silenly succeed when it should fail.

The selected symbol for the probe is expected to remain unchaged across different versions of wasmedge.
If it ever changes, then compilation should fail, indicating that the symbol needs to be updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -666,7 +666,6 @@ dependencies = [
  "serial_test",
  "ttrpc",
  "wasmedge-sdk",
- "wasmedge-sys",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Mossaka (approved):
LGTM! Thanks

PR #389, opened 2023-11-07
feat: support any shim version in `shim_main`

Mossaka:
This PR adds a `shim_version` parameter to `shim_main` macro to relex the version for shim implementations.

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/cli.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/cli.rs
@@ -36,11 +36,11 @@ macro_rules! revision {
     };
 }
 
-pub fn shim_main<I>(
+pub fn shim_main<'a, I>(
     name: &str,
     version: &str,
-    revision: Option<&str>,
-    shim_version: Option<&str>,
+    revision: impl Into<Option<&'a str>>,
+    shim_version: impl Into<Option<&'a str>>,
     config: Option<Config>,
 ) where
     I: 'static + Instance + Sync + Send,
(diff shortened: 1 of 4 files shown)


jprendes (approved):
LGTM!

crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -55,12 +60,13 @@ where
 
         std::process::exit(0);
     }
+    let shim_version = shim_version.unwrap_or("v1");
```
jprendes, line 63:
```suggestion
    let shim_version = shim_version.into().unwrap_or("v1");
```

Same with `revision`.

crates/containerd-shim-wasmedge/src/main.rs
```diff
@@ -2,5 +2,5 @@ use containerd_shim_wasm::sandbox::cli::{revision, shim_main, version};
 use containerd_shim_wasmedge::WasmEdgeInstance;
 
 fn main() {
-    shim_main::<WasmEdgeInstance>("wasmedge", version!(), revision!(), None);
+    shim_main::<WasmEdgeInstance>("wasmedge", version!(), revision!(), Some("v1"), None);
```
jprendes, line 5:
```suggestion
    shim_main::<WasmEdgeInstance>("wasmedge", version!(), revision!(), "v1", None);
```

Same on the other shims.

jprendes (approved):
LGTM

jprendes (approved):
LGTM!

cpuguy83:
This is an improvement (because it unblocks things), but I don't think the original logic is very good here.

cpuguy83 (approved):
LGTM but we should probably rethink this logic for determining shim names, binaries, etc.

PR #401, opened 2023-11-16
feat(wasmtime): support wasm components

### LINKED ISSUE #384: Feature: support wasm component model and `wasi:cli/command` world
This is a feature request for enhencing the capability of the wasmtime shim this repo provides. 

## Motivation

The [Wasm Component Model](https://github.com/WebAssembly/component-model) is a Wasm proposal for building more interoperable Wasm applications. It defines a new binary format that wraps around the [Wasm core spec](https://webassembly.github.io/spec/) that specifies its import and export using richer written in a IDL known as [WIT](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md). More importantly, it serves the foundation for the upcoming [WASI preview 2](https://github.com/WebAssembly/WASI/tree/main/preview2). To support WASI preview 2, we want to be able to support Wasm component artifacts.

## Feature Request

- Enable component support by turning on the `component-model` feature in wasmtime crate. 
- Adapt the wasmtime shim to handle both the component and core module APIs, ensuring full compatibility with each.
- Extensive testing to validate that both core Wasm modules and Wasm components function correctly under the new shim.

See the list of APIs here: https://docs.rs/wasmtime/latest/wasmtime/component/index.html.

Mossaka:
This PR adds wasm components support to the wasmtime shim which enables it with the `component-model` feature on. The component loading, linking and execution has a different set of APIs than the wasm module one. I used `wasmparser` to determine if a byte array is a wasm module or a wasm component, and then dispatch to two different paths. 

To finish this work completely, I hope to add some integration tests
- [x] add a component test to wasmtime shim

As a follow up PR, I hope to add `wasmtime serve` functionality to the wasmtime shim so that we have HTTP capability for the shim.

Closes #384


crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -3,74 +3,172 @@ use containerd_shim_wasm::container::{
...
+        }
+    }
+}
+
+pub struct Data {
+    wasi: wasi_preview2::WasiCtx,
+    table: Table,
+}
+
+impl wasmtime_wasi::preview2::WasiView for Data {
```
jprendes, line 35: I'm surprised `wasmtime_wasi` doesn't provide an off the shelf way of doing this

Mossaka: Sadly, true.

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -3,74 +3,172 @@ use containerd_shim_wasm::container::{
...
+            Some(WasmBinaryType::Module) => {
+                let module = Module::from_binary(&self.engine, &data)?;
+                let mut wasi_builder = WasiCtxBuilder::new();
+                wasi_builder
+                    .args(ctx.args())?
+                    .envs(envs.as_slice())?
+                    .inherit_stdio()
+                    .preopened_dir(root_path, "/")?;
+
+                stdio.redirect()?;
```
jsturtevant, line 105: can this be moved up?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -3,74 +3,172 @@ use containerd_shim_wasm::container::{
...
+        &mut self.wasi
+    }
+}
+
+pub enum WasmBinaryType {
+    Module,
+    Component,
+}
+
+fn wasm_binary_type(bytes: &[u8]) -> Option<WasmBinaryType> {
```
jsturtevant, line 58: also suprise the don't provie this as well

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -3,74 +3,172 @@ use containerd_shim_wasm::container::{
...
+
+                log::debug!("running {path:?} with start function {func:?}");
+
+                let status = start_func.call(&mut store, &[], &mut []);
+                status
+            }
+            Some(WasmBinaryType::Component) => {
+                let component = Component::from_binary(&self.engine, &data)?;
+                let file_perms = wasi_preview2::FilePerms::all();
+                let dir_perms = wasi_preview2::DirPerms::all();
```
jsturtevant, line 132: I believe this is the way it's done today but in the case of the OCI layers I don't think we need these permissions.  Any thoughts on how we can make this more configurable?  It feels like this is where a OCI artifact with configuration that had info like this in it would come into play

jsturtevant:
Not needed as part of this but we will also need a way to compose together multiple components?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -3,74 +3,172 @@ use containerd_shim_wasm::container::{
...
-                Module::from_binary(&self.engine, &module.layer)?
+                module.layer.clone()
             }
             [..] => bail!("only a single module is supported when using images with OCI layers"),
         };
 
-        let mut linker = Linker::new(&self.engine);
+        let status = match wasm_binary_type(&data) {
+            Some(WasmBinaryType::Module) => {
+                let module = Module::from_binary(&self.engine, &data)?;
```
jsturtevant, line 97: would it make sense from a readability/maintaince to move these longer sections calls to functions?

crates/containerd-shim-wasmtime/src/instance.rs
jprendes: I guess we need to figure out the story for this.
I can see the argument where te runtime provides an fs component with access to the container root, but you can choose to use your own fs component instead, and ignore/proxy the one provided by the runtime.

If we want to introduce config files as custom layers, it would be good to have a discussion with a wider group (wasi-wg, or even wasi-cg)

jprendes: > If we want to introduce config files as custom layers, it would be good to have a discussion with a wider group (wasi-wg, or even wasi-cg)

They may have already thought of possible solutions, or many people even outside container (e.g., wasmtime itself) could have a similar pain point.

jsturtevant: > If we want to introduce config files as custom layers, it would be good to have a discussion with a wider group (wasi-wg, or even wasi-cg)

+1 to wider discussion, wasn't intending to figure it out on this PR just calling it out.   I don't think it should be a custom layer either, that feels like its too open. I think this is more evidence that we really want a "real" "OCI artifact" with its own custom config media type that contains WASI specific meta data

Mossaka: Raised an issue here for further discussion: https://github.com/containerd/runwasi/issues/413

Mossaka:
This PR is ready to review. Could you PTAL? @jsturtevant ?

crates/containerd-shim-wasm/src/container/wasm.rs
```diff
@@ -0,0 +1,49 @@
...
+    WASM_MAGIC_NUMBER[3],
+    WASM_MODULE_VERSION.to_le_bytes()[0],
+    WASM_MODULE_VERSION.to_le_bytes()[1],
+    KIND_MODULE.to_le_bytes()[0],
+    KIND_MODULE.to_le_bytes()[1],
+];
+
+impl WasmBinaryType {
+    /// Returns the type of the wasm binary.
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
```
jsturtevant, line 40: This feels like tsomething that should be avaliable upstream, Do we have an issue opened that we could reference here to remove this one upstreamed?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -1,83 +1,89 @@
...
+        log::info!("building wasi context");
+        let wasi_ctx = prepare_wasi_ctx(ctx, envs)?;
+        let store = Store::new(&self.engine, wasi_ctx);
 
-        log::debug!("running with start function {func:?}");
+        let status = match WasmBinaryType::from_bytes(&wasm_binary) {
+            Some(WasmBinaryType::Module) => self.execute_module(&wasm_binary, store, &func)?,
+            Some(WasmBinaryType::Component) => self.execute_component(wasm_binary, store, func)?,
+            None => bail!("not a valid wasm binary format"),
+        };
```
jsturtevant, line 85: Nice, I think this reads better!

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -88,7 +94,107 @@ impl Engine for WasmtimeEngine {
...
+            Dir::from_std_file(File::open("/")?),
+            dir_perms,
+            file_perms,
+            "/",
+        );
+    let wasi_preview2_ctx = wasi_preview2_builder.build();
+    let wasi_data = WasiData {
+        wasi_preview1: wasi_preview1_ctx,
+        wasi_preview2: wasi_preview2_ctx,
+        wasi_preview2_table: wasi_preview2::Table::new(),
```
jsturtevant, line 197: should we be going through the effort to build both of these contexts when it is only one type (either module/component)?  If not, the build of the store happened insinde the  `execute_component` then this could become something like 

```
fn prepare_wasi_ctx(
    ctx: &impl RuntimeContext,
    envs: Vec<(String, String)>,
    type: wasmbinarytype,
) -> Result<WasiData, anyhow::Error> {

    match type {
            Some(WasmBinaryType::Module) => return data { wasi1_data}
            Some(WasmBinaryType::Component) => return data {wasi2_data},
```

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -1,83 +1,89 @@
...
+        Self {
+            engine: wasmtime::Engine::new(&config)
+                .context("failed to create wasmtime engine")
+                .unwrap(),
+        }
+    }
+}
+
+/// Data that contains both wasi_preview1 and wasi_preview2 contexts.
+pub struct WasiData {
```
jsturtevant, line 33: maybe `WasiCtx`?

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -44,6 +46,35 @@ pub enum Source<'a> {
...
 }
 
+impl<'a> Source<'a> {
+    /// Returns the raw bytes of the WASM binary.
+    ///
+    /// If the source is a file, the file is read from the file system.
+    /// If the source is an OCI layer, the layer is read from the OCI spec.
+    ///
+    /// At the moment, only a single module is supported when using images with OCI layers.
+    pub fn into_wasm_binary(self) -> Result<Vec<u8>, anyhow::Error> {
```
jsturtevant, line 56: nice, I think it is still valauble to have some sense where the layers come from with the source? But I like this since, we could use this as the api to that hides composing together multiple components.

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -44,6 +46,35 @@ pub enum Source<'a> {
...
+                log::info!("loading module from file");
+                let path = path
+                    .resolve_in_path_or_cwd()
+                    .next()
+                    .context("module not found")?;
+                std::fs::read(path)?
+            }
+            Source::Oci([module]) => {
+                log::info!("loading module wasm OCI layers");
+                module.layer.clone()
```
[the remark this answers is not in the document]
Mossaka: Yes because the function returns a `Vec<u8>` but module is a `&WasmLayer`

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: We could, but I think the current appraoch is nice because we only have one wasi_ctx type that contains both wasip1 and wasip2 APIs. This could potentially benefits a future case where we want to run wasm modules with wasip2 api.

Mossaka:
Added a new commit to support `wasi:cli/command` world that exports a `run` function.

crates/containerd-shim-wasmtime/src/instance.rs
jsturtevant: > This could potentially benefits a future case where we want to run wasm modules with wasip2 api

what does this look like?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -88,7 +94,122 @@ impl Engine for WasmtimeEngine {
...
+        let mut linker = wasmtime_component::Linker::new(&self.engine);
+
+        wasi_preview2::command::sync::add_to_linker(&mut linker)?;
+
+        log::info!("instantiating component");
+
+        // This is a adapter logic that converts wasip1 `_start` function to wasip2 `run` function.
+        //
+        // TODO: think about a better way to do this.
+        if func == "_start" {
```
jsturtevant, line 154: Can we make this assumption?

crates/containerd-shim-wasmtime/src/tests.rs
```diff
@@ -101,3 +101,42 @@ fn test_has_default_devices() -> anyhow::Result<()> {
...
+
+// Test that the shim can execute a wasm component that is
+// compiled with wasip2.
+//
+// This is using the `wasi:cli/command` world to run the component.
+#[test]
+#[serial]
+fn test_wasip2_component() -> anyhow::Result<()> {
+    let (exit_code, stdout, _) = WasiTest::<WasiInstance>::builder()?
+        .with_wasm(COMPONENT_HELLO_WORLD)?
```
jsturtevant, line 133: how was this component compiled?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -88,7 +94,122 @@ impl Engine for WasmtimeEngine {
                 _ => Err(err),
             }
         })?;
+        log::info!("wasi context ready");
```
jsturtevant, line 97: copy/paste?

crates/containerd-shim-wasm/src/container/context.rs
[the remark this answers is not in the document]
jsturtevant: this was done by design to avoid coping data. Can this return &[u8] instead of vec?

crates/containerd-shim-wasmtime/src/tests.rs
Mossaka: yeah I intended to not make big changes to our exiting test infras as compiling components requires a different set of tools than compiling modules. 

This component was compiled from another place, essentially followed https://component-model.bytecodealliance.org/language-support/rust.html but added the wasip2

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: yes

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: I will document this

crates/containerd-shim-wasm/src/container/context.rs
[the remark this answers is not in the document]
Mossaka: I could roll back to the original one as I am not sure if there is a way to return &[u8] since the file is read in-function.

crates/containerd-shim-wasmtime/README.md
```diff
@@ -0,0 +1,15 @@
+## containerd-shim-wasmtime
+
+This is a [containerd] shim for running WebAssembly modules and components using [wasmtime].
+
+[containerd]: https://containerd.io/
+[wasmtime]: https://wasmtime.dev/
+
+### WASI
+
+This shim assumes that the Wasm moudles it runs are [WASI] modules. WASI is a system interface for WebAssembly. If no entrypoint is specified, the shim will look for a `_start` function in the module, which is an initial point of execution when the module is loaded in the runtime. The `_start` funciton is a WASI convention for the Command modules (see the distinction between the Command and Reactors [here]).
```
jsturtevant, line 10:
```suggestion
This shim assumes that the Wasm moudles it runs are [WASI] modules. WASI is a system interface for WebAssembly. If no entrypoint is specified, the shim will look for a `_start` function in the module, which is an initial point of execution when the module is loaded in the runtime. The `_start` funciton is a WASI convention for the Command modules (see the [distinction between the Command and Reactors](link)).
```

is more screen reader friendly

crates/containerd-shim-wasmtime/src/tests.rs
jsturtevant: maybe comment on how to compile it?  I worry about future me trying to figure this out :-)

crates/containerd-shim-wasm/src/container/context.rs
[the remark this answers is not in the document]
jsturtevant: What do you mean by roll back to the original?

I see the problem, the `Source::File` would create a temporary variable that can't be returned as borrowed.  Maybe need to rethink the source,  The read of the file could be moved to the call on `entry point` and that would make it a non-local variable but this would have cascading changes.  Another option might be to have a "intermediate step" that capture that value.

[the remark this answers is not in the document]
Mossaka: > The read of the file could be moved to the call on `entrypoint()`. 

I would be a bit more careful here since that means a huge slowdown to entrypoint invocation, which is used in `can_handle()` and `is_linux_container()` functions

crates/containerd-shim-wasmtime/src/tests.rs
Mossaka: Done.

crates/containerd-shim-wasm/src/container/context.rs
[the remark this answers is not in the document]
Mossaka: Reverted the code back.

jsturtevant (commented):
LGTM

jsturtevant:
needs some linter fixes

crates/containerd-shim-wasm/src/container/context.rs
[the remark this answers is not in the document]
Mossaka: Yup I will have a separate PR to clean that

Issue #357, opened 2023-10-12
Running complex code after the process is `clone`d in `libcontainer` can lead to UB


jprendes:
> @utam0k calling setgroups / setresgid / setresuid directly would workaround the issue.
> I am slightly concerned because in normal youki / runc, we soon call execve, which starts a clean process.
> In the case of runwasi, we don't, we keep running on the same tainted process.
> From [nix::clone](https://docs.rs/nix/latest/nix/sched/fn.clone.html) and [nix::fork](https://docs.rs/nix/latest/nix/unistd/fn.fork.html), even memory allocation might be unsafe after the `clone` call (I assume it's platform / variant dependent).
> 

_Originally posted by @jprendes in https://github.com/containerd/runwasi/issues/347#issuecomment-1755433335_

jprendes:
Some context:

The code we run in our impl of the `libcontainer`'s `Executor` is run in a `clone`d process.

In POSIX, the kind of code that is guaranteed to run safely in the child process after a `clone` is restricted to async-signal-safe code. Examples of code that are **NOT** guaranteed to be AS-safe include `malloc`, `printf` and `pthread_*`.

See for example how the snippet in [nix::fork](https://docs.rs/nix/latest/nix/unistd/fn.fork.html) uses `write(..)` in the child process instead of `println!(..)` exactly because of this.

It seems that in our case with glibc and musl, it works fine, but this is fragile.

jprendes:
I'm thinking that a possible approach would be to use the async api of containerd/rust-extensions and have the shim be a single threaded process.

This is assuming that the way rust-extensions handles async is single threaded.

Mossaka:
Hey @jprendes I know you've worked on the asyncify runwasi, do you have any updates on that one? Do you think we need another issue to track the progress of it?

jprendes:
I can think of two options to workaround this issue:

### calling exec and replacing the whole process
That's what normal for-exec does, and what a normal container does.
The **disadvantage** is that we loose any sharing between the shim and the container (e.g., anything pre-initialized by the shim won't seen by the container).
I know @cpuguy83 would like to see all containers in a pod sharing the same runtime, and that would become impossible with this approach (admitedly, that's not the direction we are heading anyway with using youki).
The **advantage** is that for windows we would need to do that anyway (IIUC, in windows there's no fork and exec, but rather fork+exec as one single thing), and having a similar behavior in all platforms might be nice.

### Running the shim async, and in only one thread
[...]

PR #405, opened 2023-11-23
Store Compiled Modules when using OCI Wasm Images

jsturtevant:
this would replace #44


jsturtevant:
fyi @cpuguy83

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -62,41 +72,220 @@ impl Client {
...
+                .ok_or_else(|| {
+                    ShimError::Containerd(format!(
+                        "no response received after write request for {}",
+                        expected.clone()
+                    ))
+                })?;
+
+            // Write and commit at same time
+            let mut labels = HashMap::new();
+            labels.insert("runwasi.io/precompiled".to_string(), "".to_string());
```
cpuguy83, line 137: Label should probably include the digest of the source content.

cpuguy83 (changes requested):
I think we need to add some GC labels so containerd does not collect this content.
See: https://github.com/containerd/containerd/blob/8459273f806e068e1a6bacfaf1355bbbad738d5e/docs/garbage-collection.md#garbage-collection-labels

How do we invalidate (and de-reference in containerd) pre-compiled modules?
As I understand it a pre-compiled module will not work across different versions of a runtime.

Also we likely need to annotate *what* runtime the content is for.

Also, when writing the content we'll need to attach a lease to it until gc labels are applied otherwise containerd will collect it.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -62,41 +72,220 @@ impl Client {
...
+    pub fn save_content(&self, data: Vec<u8>) -> Result<String> {
+        self.rt.block_on(async {
+            // create a channel to feed the stream; only sending one message at a time so we can set this to one
+            let (tx, rx) = mpsc::channel(1);
+
+            let len = data.len() as i64;
+            let expected = digest(data.clone());
+            let expected = format!("sha256:{}", expected);
+            let mut client = ContentClient::new(self.inner.clone());
+            let r#ref = "test".to_string();
```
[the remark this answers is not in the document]
jsturtevant: woops forgot to update, This can be anything but should be unique as its a key used to determine if there are open transactions happening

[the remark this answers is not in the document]
cpuguy83: I think this should be deterministic and descriptive.

Here's containerd's implementation for generating a ref key:

```
// MakeRefKey returns a unique reference for the descriptor. This reference can be
// used to lookup ongoing processes related to the descriptor. This function
// may look to the context to namespace the reference appropriately.
func MakeRefKey(ctx context.Context, desc ocispec.Descriptor) string {
	key := desc.Digest.String()
	if desc.Annotations != nil {
		if name, ok := desc.Annotations[ocispec.AnnotationRefName]; ok {
			key = fmt.Sprintf("%s@%s", name, desc.Digest.String())
		}
	}

	if v := ctx.Value(refKeyPrefix{}); v != nil {
		values := v.(map[string]string)
		if prefix := values[desc.MediaType]; prefix != "" {
			return prefix + "-" + key
		}
	}

	switch mt := desc.MediaType; {
	case mt == images.MediaTypeDockerSchema2Manifest || mt == ocispec.MediaTypeImageManifest:
		return "manifest-" + key
	case mt == images.MediaTypeDockerSchema2ManifestList || mt == ocispec.MediaTypeImageIndex:
		return "index-" + key
	case images.IsLayerType(mt):
		return "layer-" + key
	case images.IsKnownConfig(mt):
		return "config-" + key
	default:
		log.G(ctx).Warnf("reference for unknown type: %s", mt)
		return "unknown-" + key
	}
}
```

Mossaka:
@jsturtevant Started reviewing this PR. How do I run this to see if the modules are stored?

jsturtevant:
The shim logs have will have statements and you will also be able to verify in containerd using `ctr`:

Find the compiled info:

```
sudo ctr i ls | grep "runwasi.io"
ghcr.io/containerd/runwasi/wasi-demo-oci:latest                                                             application/vnd.oci.image.manifest.v1+json
               sha256:60fccd77070dfeb682a1ebc742e9d677fc452b30a6b99188b081c968992394ce 2.4 MiB   wasi/wasm                                                                                                                           
runwasi.io/precompiled=sha256:b36753ab5a46f26f6bedb81b8a7b489cede8fc7386f1398706782e225fd0a98e
```

See it in the content store:
```
sudo ctr content ls | grep "b36753ab5a46f26f6bedb81b8a7b489cede8fc7386f139870"
sha256:60fccd77070dfeb682a1ebc742e9d677fc452b30a6b99188b081c968992394ce 561B    2 months        containerd.io/gc.ref.content.0=sha256:a3c18cd551d54d3cfbf67acc9e8f7ef5761e76827fe7c1ae163fca0193be88b3,containerd.io/gc.ref.content.config=sha256:85b7f2b562fe8665ec9d9e6d47ab0b24e2315627f5f558d298475c4038d71e8b,containerd.io/gc.ref.content.precompile=sha256:b36753ab5a46f26f6bedb81b8a7b489cede8fc7386f1398706782e225fd0a98e
sha256:b36753ab5a46f26f6bedb81b8a7b489cede8fc7386f1398706782e225fd0a98e 626.4kB 3 days          runwasi.io/precompiled=sha256:60fccd77070dfeb682a1ebc742e9d677fc452b30a6b99188b081c968992394ce
```

save it to local file:
```
ctr content get sha256:b36753ab5a46f26f6bedb81b8a7b489cede8fc7386f1398706782e225fd0a98e > precompiled.wasm
```

crates/containerd-shim-wasm/src/sandbox/containerd.rs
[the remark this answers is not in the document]
jsturtevant: i've updated this to use `format!("precompile-{}-{}-{}", shim.name, shim.version, expected-sha);`

jsturtevant:
> > Also, when writing the content we'll need to attach a lease to it until gc labels are applied otherwise containerd will collect it.
> 
> Ah, I was wondering about that! There did seem to be a small moment where it could be cleaned up before I put the label on the image. I will take a look at putting a lease on it.

Added a lease during new content creation

jsturtevant:
Another question I had was if this is something we want to turn on for all images (as the code is now) or if we want to put an annotation on the image that would enable this feature.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -137,23 +431,144 @@ impl Client {
             log::info!("manifest is not in WASM OCI image format");
             return Ok((vec![], platform));
         };
+
         log::info!("found manifest with WASM OCI image format.");
+        // This label is unique across runtimes and versions
+        // a precompiled component/module will not work across different runtimes or versions
+        let label = format!(
+            "runwasi.io/precompiled/{}/{}",
```
Mossaka, line 439: Could we declare this uri `runwasi.io/precompile` as a static string to be used in the entire module?

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -137,23 +431,144 @@ impl Client {
...
+                },
+            )
+            .unwrap();
+        assert_eq!(expected, returned.digest.clone());
+
+        let data = client.read_content(returned.digest.clone()).unwrap();
+        assert_eq!(data, b"hello world");
+
+        // need to drop the lease to be able to create a second one
+        drop(returned);
```
Mossaka, line 553: I'd like to see a test case where the `drop()` function is not invoked explictly. What will happen?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -13,12 +15,22 @@ pub struct WasmtimeEngine {
     engine: wasmtime::Engine,
 }
 
+fn version_info() -> &'static RuntimeInfo {
+    static INFO: OnceLock<RuntimeInfo> = OnceLock::new();
+    INFO.get_or_init(|| RuntimeInfo {
+        name: "wasmtime",
+        version: option_env!("WASMTIME_VERSION_INFO").unwrap_or(env!("CARGO_PKG_VERSION")),
+    })
+}
```
Mossaka, line 24: Is OnceLock necessary? 

```suggestion
static INFO: RuntimeInfo = RuntimeInfo {
    name: "wasmtime",
    version: option_env!("WASMTIME_VERSION_INFO").unwrap_or(env!("CARGO_PKG_VERSION")),
};

fn version_info() -> &'static RuntimeInfo {
    &INFO
}
```

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -13,12 +15,22 @@ pub struct WasmtimeEngine {
...
+
 impl Engine for WasmtimeEngine {
-    fn name() -> &'static str {
-        "wasmtime"
+    fn info() -> &'static RuntimeInfo {
+        version_info()
     }
 
     fn run_wasi(&self, ctx: &impl RuntimeContext, stdio: Stdio) -> Result<i32> {
+        let _config = wasmtime::Config::new();
```
Mossaka, line 32: Not used?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -30,8 +42,6 @@ impl Engine for WasmtimeEngine {
             .inherit_stdio()
             .preopened_dir(root_path, "/")?;
 
-        stdio.redirect()?;
```
Mossaka, line 33: Removed by accident?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -91,4 +105,30 @@ impl Engine for WasmtimeEngine {
...
+            log::info!("using precompiled module");
+            unsafe { Module::deserialize(engine, bytes) }
+        }
+        Some(Precompiled::Component) => {
+            bail!("components not supported")
+        }
+        None => {
+            bail!("invalid precompiled module")
+        }
+    }
```
Mossaka, line 133: Since this PR only implements `load_precompiled` for wasmtime runtime, I wonder if every other wasm runtime has the capability to do something similar to `detect_precompiled`?

If we ought to impl precompiled to other runtimes, I would prefer to add `detect_precompiled()` to the `Engine` triat defined in `containerd-shim-wasm` crate.

Mossaka (commented):
Hey @jsturtevant great PR! Sorry I was late for the code review. 

Honestly it took me some time to grasp the interaction between containerd and the sandbox for saving / retriving precompiled content using leases. I am still not 100% sure the full picture of the interactions. Would be appreciate if you clarify the logic here (maybe document it somewhere?)

crates/containerd-shim-wasmtime/src/instance.rs
Mossaka: Oh I realized `detect_precompiled` is detecting a serialization format that's wasmtime specific. It may not work for other runtimes then.

Mossaka: https://github.com/bytecodealliance/wasmtime/blob/7690c500228ae005fc70ee3b6297bb4b403686d5/crates/wasmtime/src/engine/serialization.rs#L147-L163

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: you will get an error, I can add a test case

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -62,41 +129,268 @@ impl Client {
...
+            ContentClient::new(self.inner.clone())
+                .delete(req)
+                .await
+                .map_err(|err| ShimError::Containerd(err.to_string()))?;
+            Ok(())
+        })
+    }
+
+    // wrapper around lease that will create a lease and return a guard that will delete the lease when dropped
+    fn lease(&self, r#ref: String) -> Result<LeaseGuard> {
```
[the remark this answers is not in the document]
jsturtevant: it is called in `ref` in the containerd codebase which I used for reference. I can rename it to reference to avoid the need to  qualify it with `r#`

crates/containerd-shim-wasmtime/src/instance.rs
jsturtevant: I will try this but I believe it is...

crates/containerd-shim-wasmtime/src/instance.rs
jsturtevant: wasmedge has a pre-compiled mode, which was thinking of trying after this

jsturtevant:
> Another question I had was if this is something we want to turn on for all images (as the code is now) or if we want to put an annotation on the image that would enable this feature.

I've thought on this a bit more and I think it should be opt in feature so will add support for an annotation

cpuguy83:
I don't think an annotation on the image would work well.
A runtime annotation could work but...
I'm curious why do you think it should be opt in? I can see the desire to opt in for now because its kind of experimental, but then the whole stack is.

We must compile the wasm to run it at all, so we already need to store it.
Keeping it cached so the same wasm isn't compiled more than once is just an optimization.

CaptainVincent:
Yes. [This lines of the code](https://github.com/WasmEdge/wasmedge-rust-sdk/blob/fa3f3ce03ef4382f2bf69bd0a3355ca68dc4485f/src/compiler.rs#L40-L92) provides two methods about `bytes to file` and `file to file`. If needed, I can also add `bytes to bytes` inside wasmedge-rust-sdk or integration layer.

> @CaptainVincent does WasmEdge Rust SDK have an API to AOT compile a Wasm module to bytes that's similar to https://github.com/bytecodealliance/wasmtime/blob/7690c500228ae005fc70ee3b6297bb4b403686d5/crates/wasmtime/src/engine.rs#L222-L223?

jsturtevant:
> I can see the desire to opt in for now because its kind of experimental, but then the whole stack is.
>
> We must compile the wasm to run it at all, so we already need to store it.
> Keeping it cached so the same wasm isn't compiled more than once is just an optimization.

I was leaning towards it being experimental but I can keep it as.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -62,41 +129,268 @@ impl Client {
...
+        self.rt.block_on(async {
+            let mut lease_labels = HashMap::new();
+            let expire = chrono::Utc::now() + chrono::Duration::hours(24);
+            lease_labels.insert("containerd.io/gc.expire".to_string(), expire.to_rfc3339());
+            let lease_request = containerd_client::services::v1::CreateRequest {
+                id: r#ref.clone(),
```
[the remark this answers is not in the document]
jsturtevant: The reference gets moved below on line 167, so this clone is needed 

```
borrow of moved value: `reference`
move occurs because `reference` has type `std::string::String`, which does not implement the `Copy` 
```

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -62,41 +129,268 @@ impl Client {
...
+    pub fn save_content(
+        &self,
+        data: Vec<u8>,
+        original_digest: String,
+        info: &RuntimeInfo,
+    ) -> Result<WriteContent> {
+        let expected = digest(data.clone());
+        let expected = format!("sha256:{}", expected);
+        let r#ref = format!("precompile-{}-{}-{}", info.name, info.version, expected);
+        let lease = self.lease(r#ref.clone())?;
```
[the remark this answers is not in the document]
jsturtevant: this one is needed, lease is kept around and the original reference is used/moved in the async section

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -62,41 +129,268 @@ impl Client {
...
+            let (tx, rx) = mpsc::channel(1);
+
+            let len = data.len() as i64;
+
+            let mut client = ContentClient::new(self.inner.clone());
+
+            // Send write request with Stat action to containerd to let it know that we are going to write content
+            // if the content is already there, it will return early with AlreadyExists
+            let req = WriteContentRequest {
+                r#ref: r#ref.clone(),
```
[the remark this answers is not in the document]
jsturtevant: clone is needed here but removed it in the error message below.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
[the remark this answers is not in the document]
jsturtevant: i removed the extra clone in the error message below

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -1,32 +1,98 @@
...
+            }
+        });
+    }
 }
 
 // sync wrapper implementation from https://tokio.rs/tokio/topics/bridging
 impl Client {
     // wrapper around connection that will establish a connection and create a client
-    pub fn connect(address: impl AsRef<Path>, namespace: impl ToString) -> Result<Client> {
+    pub fn connect(address: &str, namespace: impl ToString) -> Result<Client> {
```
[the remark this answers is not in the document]
jsturtevant: I am guessing there is a proper way to do this but `AsRef<Path>` has no `to_string` method.  I added `impl AsRef<Path> + ToString` but get `borrow of moved value: address`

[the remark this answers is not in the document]
jsturtevant: oh, got it figured out

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -79,6 +80,8 @@ impl RuntimeContext for WasiContext<'_> {
 
         let source = if self.wasm_layers.is_empty() {
             Source::File(PathBuf::from(path))
+        } else if self.wasm_layers.len() == 1 && self.wasm_layers[0].precompiled {
+            Source::Precompiled(&self.wasm_layers[0])
         } else {
             Source::Oci(self.wasm_layers)
         };
```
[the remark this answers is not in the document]
jsturtevant: I changed how this works, so no longer relevant

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -337,6 +340,7 @@ mod tests {
             wasm_layers: &[WasmLayer {
                 layer: vec![],
                 config: Descriptor::new(oci_spec::image::MediaType::Other("".to_string()), 10, ""),
+                precompiled: false,
             }],
```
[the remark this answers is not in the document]
jsturtevant: changed how this works no longer needed

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -1,32 +1,98 @@
...
+
+        let mut client = LeasesClient::new(client.unwrap());
+
+        rt.block_on(async {
+            let req = containerd_client::services::v1::DeleteRequest { id, sync: false };
+            let req = with_namespace!(req, self.namespace);
+            let result = client.delete(req).await;
+
+            if result.is_err() {
+                log::warn!("failed to remove lease.");
```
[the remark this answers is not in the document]
jsturtevant: I'm not sure, these two errors that don't necessary mean the lease wasn't cleaned up in some way.  I've updated it to error and we can always downgrade if we end up seeing them and its not an issue.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -1,32 +1,98 @@
...
+        let mut req = Request::new($req);
+        let md = req.metadata_mut();
+        // https://github.com/containerd/containerd/blob/main/namespaces/grpc.go#L27
+        md.insert("containerd-namespace", $ns.parse().unwrap());
+        md.insert("containerd-lease", $lease_id.parse().unwrap());
+        req
+    }};
+}
+
+struct LeaseGuard {
```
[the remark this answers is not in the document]
jsturtevant: updated

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: updated

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: added an assert

crates/containerd-shim-wasmtime/src/instance.rs
jsturtevant: it is:

> cannot call non-const fn `std::option::Option::<&str>::unwrap_or` in statics
calls in statics are limited to constant functions, tuple structs and tuple variants
consider wrapping this expression in `Lazy::new(|| ...)` from the `once_cell` crate: https://crates.io/crates/once_cellrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B3%5D?3#file%3A%2F%2F%2Fhome%2Fjstur%2Fprojects%2Frunwasi%2Fcrates%2Fcontainerd-shim-wasmtime%2Fsrc%2Finstance.rs)

jsturtevant:
> Honestly it took me some time to grasp the interaction between containerd and the sandbox for saving / retriving precompiled content using leases. I am still not 100% sure the full picture of the interactions. Would be appreciate if you clarify the logic here (maybe document it somewhere?)

good idea!  I've added some documentation, you can see the rendered page by viewing the markdown: i.e: https://github.com/containerd/runwasi/blob/0135abace198e39a1f4f6aae6ae74605619e42b2/docs/oci-descision-flow.md

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -52,4 +59,14 @@ pub trait Engine: Clone + Send + Sync + 'static {
     fn supported_layers_types() -> &'static [&'static str] {
         &["application/vnd.bytecodealliance.wasm.component.layer.v0+wasm"]
     }
+
+    /// Precomiple a module
+    fn precompile(&self, _layers: &[Vec<u8>]) -> Result<Vec<u8>> {
+        bail!("precompilation not supported for this runtime")
+    }
+
+    /// Precomiple a module
```
Mossaka, line 68: Please provide more descriptive doc

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -1,44 +1,135 @@
...
+static PRECOMPILE_PREFIX: &str = "runwasi.io/precompiled";
 
 pub(crate) struct Client {
     inner: Channel,
     rt: Runtime,
     namespace: String,
+    address: String,
+}
+
+mod lease {
```
Mossaka, line 42: Why not place it in its own file?

crates/containerd-shim-wasm/src/sandbox/containerd.rs
```diff
@@ -137,23 +452,148 @@ impl Client {
...
-            })
+            .filter(|x| is_wasm_layer(x.media_type(), T::supported_layers_types()))
+            .map(|config| self.read_content(config.digest()))
             .collect::<Result<Vec<_>>>()?;
+
+        if T::can_precompile() {
+            log::info!("precompiling module");
+            let precompiled = engine.precompile(layers.as_slice())?;
+            let precompiled_content =
+                self.save_content(precompiled.clone(), image_digest.clone(), T::info())?;
```
Mossaka, line 486: Would this have visible increase on the latency of running a non-cached image?

docs/oci-descision-flow.md
```diff
@@ -0,0 +1,47 @@
+# OCI pre-compilation
+
+The OCI images layers are loaded from containerd.  If the runtime supports pre-compilation the images will be precompiled and cached using the containerd content store.  
+
+```mermaid
+graph TD
```
Mossaka, line 6: Love this diagram, thank you!

Mossaka (dismissed):
LGTM! Add a few nits and one question about the performance impact of running a non-cached image.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: We don't have a great way to measure this currently.  This will only be called on first run, then it will be cached after this.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: I moved this and the containerd client to its own module

cpuguy83:
Tested this out.

1. This doesn't work with Docker. Not your fault since Docker is not filling in `container.Image` which is how you are traversing the content. I have a patch prepared to remedy this, but it would only work when docker is using containerd-backed storage.
2. `container.Image` is not the best data to use... we should add a new field to the containerd protos (in containerd/containerd) for an immutable reference to the image manifest being used... for now we'd need to work with what we've got
3. Removing the pre-compiled content seems to make the container refuse to ever start until we wipe out the image and start over.

On the Docker usage, I've noticed some peculiar things with runwasi.
The first container start is quick (considering it is compiling the content).
Then subsequent starts of the same container are actually slower.
Removing the container in docker and recreating gives a faster startup (even faster since it has the pre-compiled content) and then the 2nd start again slows down (by an order of magnitude here).

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: I used `time` to measure the first run vs the second run and I am seeing the following: 

```
time sudo ctr run --rm --runtime=io.containerd.wasmtime.v1 ghcr.io/contain
erd/runwasi/wasi-demo-oci:latest testwasmoci wasi-demo-oci.wasm echo 'hello'
```

first run: `real    0m1.956s`
second run: `real    0m1.235s`

For an image without OCI layers I see `real   0m2.015s.`  (but it doesn't hit this code path)

These times are really slow generally and not sure if it is my machine/debug mode or what.  But I do see the increase in start up time on the second run.

jsturtevant:
> 1. This doesn't work with Docker. Not your fault since Docker is not filling in `container.Image` which is how you are traversing the content. I have a patch prepared (for docker) to remedy this, but it would only work when docker is using containerd-backed storage.

Thanks!  

> container.Image is not the best data to use... we should add a new field to the containerd protos (in containerd/containerd) for an immutable reference to the image manifest being used... for now we'd need to work with what we've got

Created an issue (#460) to look into this and possibly use something from the conatiners snapshot as you suggested offline.  I think we can improve the overall design in way Containerd passes WASM info but waiting on the discussions in the CNCF wg-wasm in regards to an offiical OCI Artifact for wasi.

> Removing the pre-compiled content seems to make the container refuse to ever start until we wipe out the image and start over.

This is a bug!  I will fix this and look at adding tests to cover this scenario more.

jsturtevant:
> On the Docker usage, I've noticed some peculiar things with runwasi.
> The first container start is quick (considering it is compiling the content).
> Then subsequent starts of the same container are actually slower.
> Removing the container in docker and recreating gives a faster startup (even faster since it has the pre-compiled content) and then the 2nd start again slows down (by an order of magnitude here).

This is interesting since I am seeing an improvement in speed with the changes when using ctr
 (https://github.com/containerd/runwasi/pull/405#discussion_r1465444917). Does the slow behavior for the second start replicate with the non-oci image?

crates/containerd-shim-wasm/src/sandbox/containerd.rs
jsturtevant: the general slowness was related to DEBUG mode. I turned on Release flag and seeing much better times:

first run: `real    0m0.089s`
second run: `real    0m0.068s`

and for image with out OCI layers: `real    0m1.471s`

lann:
You might want to include [`Engine::precompile_compatibility_hash`](https://docs.rs/wasmtime/16.0.0/wasmtime/struct.Engine.html#method.precompile_compatibility_hash) in the cache key; I added this method for precisely this purpose :slightly_smiling_face:.

Precompiled binaries are sensitive not just to runtime version but also CPU and engine config features. It doesn't seem entirely implausible that the containerd store could be migrated to a new host with different CPU features and then you'd have a very gnarly performance regression to debug...

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -52,4 +59,17 @@ pub trait Engine: Clone + Send + Sync + 'static {
     fn supported_layers_types() -> &'static [&'static str] {
         &["application/vnd.bytecodealliance.wasm.component.layer.v0+wasm"]
     }
+
+    /// Precomiples a module that in in the WASM OCI layer format
```
vdice, line 63:
```suggestion
    /// Precompiles a module that is in the WASM OCI layer format
```

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -52,4 +59,17 @@ pub trait Engine: Clone + Send + Sync + 'static {
...
+    /// This is used to precompile a module before it is run and will be called if can_precompile returns true.
+    /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
+    /// The cached, precompiled module will be reloaded on subsequent runs.
+    fn precompile(&self, _layers: &[Vec<u8>]) -> Result<Vec<u8>> {
+        bail!("precompilation not supported for this runtime")
+    }
+
+    /// Precomiple a module
```
vdice, line 71:
```suggestion
    /// Precompile a module
```

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -0,0 +1,546 @@
...
+                self.save_content(precompiled.clone(), image_digest.clone(), T::info())?;
+
+            log::debug!("updating image with compiled content digest");
+            image
+                .labels
+                .insert(label, precompiled_content.digest.clone());
+            self.update_image(image)?;
+
+            // The original image is considered a root object, by adding a ref to the new compiled content
+            // We tell to containerd to not garbage collect the new content until this image is removed from the system
```
vdice, line 442:
```suggestion
            // We tell containerd to not garbage collect the new content until this image is removed from the system
```

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -96,6 +101,17 @@ impl Engine for WasmtimeEngine {
 
         Ok(status)
     }
+
+    fn precompile(&self, layers: &[Vec<u8>]) -> Result<Vec<u8>> {
+        match layers {
+            [layer] => self.engine.precompile_module(layer),
+            _ => bail!("only a single module is supported when when precompiling"),
```
vdice, line 108:
```suggestion
            _ => bail!("only a single module is supported when precompiling"),
```

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -171,6 +185,40 @@ impl WasmtimeEngine {
...
+                self.execute_component(component, store, func)
+            }
+            None => match &self.engine.detect_precompiled(wasm_binary) {
+                Some(Precompiled::Module) => {
+                    log::info!("using precompiled module");
+                    let module = unsafe { Module::deserialize(&self.engine, wasm_binary) }?;
+                    self.execute_module(module, store, &func)
+                }
+                Some(Precompiled::Component) => {
+                    log::info!("using precompiled module");
```
vdice, line 212:
```suggestion
                    log::info!("using precompiled component");
```

vdice (commented):
Sorry for the proofreading-only review, but thought I'd submit in case these items weren't caught in other reviews.

crates/containerd-shim-wasm/src/sandbox/containerd.rs
Mossaka: > I guess this is still pretty slow

I am wondering what benchmark were you comparing to when you said its slow?

Mossaka:
```
time sudo ctr run --rm docker.io/library/hello-world:latest hello-world
```

First run: 0.345s
Second run: 0.258s

Ah I can see now, it is slow compare to running native image like hello-world in `ctr`, albert the obvious fact that I was running it in a different machine.

jsturtevant:
quick update, When I've added the tests it seems to have exposed a possible bug which I am debugging now.  Essentially the OCI test run fine but when run with the other tests they are hanging and failing _if run after_ an OCI test.  We create a new engine each time so I am not really clear on what is causing the issue yet.

jsturtevant:
> quick update, When I've added the tests it seems to have exposed a possible bug which I am debugging now. Essentially the OCI test run fine but when run with the other tests they are hanging and failing _if run after_ an OCI test. We create a new engine each time so I am not really clear on what is causing the issue yet.

I tracked this down to the call to parallelize the compile step here.  The engine should be usable across threads and each test should be creating a new "engine" so I don't know why this happens.  
 https://github.com/bytecodealliance/wasmtime/blob/b6a8abc6ce31cf6db7ad5edb9b2567620190ce35/crates/wasmtime/src/engine.rs#L142

Mossaka:
> > quick update, When I've added the tests it seems to have exposed a possible bug which I am debugging now. Essentially the OCI test run fine but when run with the other tests they are hanging and failing _if run after_ an OCI test. We create a new engine each time so I am not really clear on what is causing the issue yet.
> 
> I tracked this down to the call to parallelize the compile step here. The engine should be usable across threads and each test should be creating a new "engine" so I don't know why this happens. https://github.com/bytecodealliance/wasmtime/blob/b6a8abc6ce31cf6db7ad5edb9b2567620190ce35/crates/wasmtime/src/engine.rs#L142

Did we turn on the "parallel-compilation" feature of wasmtime?

jsturtevant:
> Did we turn on the "parallel-compilation" feature of wasmtime?

yes we turn it that feature https://github.com/containerd/runwasi/blob/ea8c04f5d8073119d6b38f360311a8ec52dc9911/crates/containerd-shim-wasmtime/Cargo.toml#L23
 and it looks to be avaliable on my machine as well as on CI

My next step is to make a small/simpler reproducible example

jsturtevant:
I was able to create a simple repo: https://github.com/jsturtevant/rayon-clone-issue/blob/master/src/main.rs.  After adding the unit tests I am effectively running into https://github.com/containerd/runwasi/issues/357 which is also called out in https://github.com/rayon-rs/rayon/issues/708#issuecomment-555699090

jsturtevant:
> You might want to include [`Engine::precompile_compatibility_hash`](https://docs.rs/wasmtime/16.0.0/wasmtime/struct.Engine.html#method.precompile_compatibility_hash) in the cache key; I added this method for precisely this purpose 🙂.
> 
> Precompiled binaries are sensitive not just to runtime version but also [CPU and engine config features](https://github.com/bytecodealliance/wasmtime/blob/e7064d4d05d9246b27915ada78c97151242929e9/crates/wasmtime/src/module.rs#L1135-L1152). It doesn't seem entirely implausible that - for example - the content store could be migrated to a new host with different CPU features and then you'd have a very gnarly performance regression to debug...

I've updated wasmtime to use this hash for the label

jsturtevant:
> I was able to create a simple repo: https://github.com/jsturtevant/rayon-clone-issue/blob/master/src/main.rs. After adding the unit tests I am effectively running into #357 which is also called out in [rayon-rs/rayon#708 (comment)](https://github.com/rayon-rs/rayon/issues/708#issuecomment-555699090)

I've added the ability to disable the parallel compilation for the tests. We won't run into this issue as is in the shim since each shim is it's own instance but we need to address it longer term with #357

jsturtevant:
test failure was #423

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -1,32 +1,49 @@
...
+impl WasiConfig for DefaultConfig {
+    fn new_config() -> Config {
         let mut config = wasmtime::Config::new();
-        config.parallel_compilation(false);
         config.wasm_component_model(true); // enable component linking
+        config
+    }
+}
+
+pub trait WasiConfig {
```
Mossaka, line 35: WasiConfig could include `Clone + Sync + Send` as supertraits

```rust
pub trait WasiConfig: Clone + Sync + Send {
    fn new_config() -> Config;
}
```

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -1,32 +1,49 @@
...
+    fn new_config() -> Config;
+}
+
+impl<T: WasiConfig + Clone + Sync + Send> Default for WasmtimeEngine<T> {
```
Mossaka, line 39: Can be simplified to 
`impl<T: WasiConfig> Default for WasmtimeEngine<T> {`

crates/containerd-shim-wasmtime/src/instance.rs
jsturtevant: ok, I thought I tried this but maybe go the syntax wrong, will try again

Mossaka (approved):
LGTM! Great work!

crates/containerd-shim-wasm/src/sandbox/containerd.rs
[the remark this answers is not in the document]
jsturtevant: no using a hash provided by runtime after learning about: https://github.com/containerd/runwasi/pull/405#issuecomment-1910462356

jsturtevant:
@cpuguy83 I've addressed the feedback, could you take another look? Thanks!

cpuguy83 (approved):

Mossaka:
🥳 Awesome! I am going to do a new release

PR #400, opened 2023-11-16
Add CONTRIBUTING.md

jsturtevant:
Adding a contributing doc which includes guidance for adding and removing new shims.

PR #408, opened 2023-11-27
Add guidance for adding and removing shims from project

jsturtevant:
Follow up to #400

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -148,7 +148,7 @@ Since we are not experts in every runtime we also need a commitment from the run
 This is a fast moving space, with lots of innovation happening and some shims may eventually need to be removed.
 
 A Shim implementation maybe subject to removal if:
- - If a shim runtime its self has not maintained for 6 months it will be subject to removal.  
+ - If a shim runtime has not been maintained for 6 months it will be subject for removal
  - If required changes to the runtime can't be merged or not supported by runtime maintainers.
  - If it falls behind in new features added to the `coantainerd-shim-wasm` due to lack of maintainance
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,7 +157,7 @@ Before removal:
 - Send notification in our slack channel 
 - make best effort to contact the maintainers (agreed to when adding the shim)
 
-After 1 month of the issue being up, if not maintainer is found we will remove the shim from the project.
+After 1 month of the issue being up, if no maintainer is found we will remove the shim from the project.
 
 ## Getting in touch
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,6 +135,30 @@ Most features will likely have most of the code in the `containerd-shim-wasm` pr
 
 A tip for developing a new feature is to implement it and test it with one runtime you are familiar with then add it to all the runtimes.  This makes it easier to test and iterate before making changes across all the runtimes.
 
+## Adding new shims
+
+We welcome new shims, though you can also host them in your own repositories as well and use the `coantainerd-shim-wasm` crate.  We recognize that the project is moving fast and having them in this repository can reduce the need for changes as well for discoverability.  
+
+Please open an issue before submitting a PR for discussion to make sure it is a good fit.  As a general rule, we want shims to be adopting WASI spec (this is after all called run*wasi* :-)). In the future we may require shims in the repository to pass WASI compliance tests when the standards mature more. See https://github.com/containerd/runwasi/issues/338 for more discussion.
+
+Since we are not experts in every runtime we also need a commitment from the runtime owners to contribute to maintenance of the shim.
+
+## Removing Shims
+
+This is a fast moving space, with lots of innovation happening and some shims may eventually need to be removed.
+
+A Shim implementation maybe subject to removal if:
+ - If a shim runtime its self has not maintained for 6 months it will be subject to removal.  
+ - If required changes to the runtime can't be merged or not supported by runtime maintainers.
+ - If it falls behind in new features added to the `coantainerd-shim-wasm` due to lack of maintainance
+
+Before removal:
+- We will create an issue in the repository, pinned to the top. 
+- Send notification in our slack channel 
+- make best effort to contact the maintainers (agreed to when adding the shim)
+
+After 1 month of the issue being up, if not maintainer is found we will remove the shim from the project.
+
 ## Getting in touch
 
 There is a lot going on in the project. If you get lost, stop by our [slack and ask questions](./README.md#community)!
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -135,6 +135,30 @@ Most features will likely have most of the code in the `containerd-shim-wasm` pr
...
+Please open an issue before submitting a PR for discussion to make sure it is a good fit.  As a general rule, we want shims to be adopting WASI spec (this is after all called run*wasi* :-)). In the future we may require shims in the repository to pass WASI compliance tests when the standards mature more. See https://github.com/containerd/runwasi/issues/338 for more discussion.
+
+Since we are not experts in every runtime we also need a commitment from the runtime owners to contribute to maintenance of the shim.
+
+## Removing Shims
+
+This is a fast moving space, with lots of innovation happening and some shims may eventually need to be removed.
+
+A Shim implementation maybe subject to removal if:
+ - If a shim runtime its self has not maintained for 6 months it will be subject to removal.
```
cpuguy83, line 151: Wording here seems a bit off.

```suggestion
 - If a shim runtime has not been maintained for 6 months it will be subject for removal
 ```

Comment (cpuguy83) [reply CONTRIBUTING.md:151]: I also think this should timeframe should be shortened for now given how things are moving.

Comment (cpuguy83) [reply CONTRIBUTING.md:151]: Maybe a callout that the shim may be removed from CI after a certain time period if it is not maintained?
Like 1 month?

**Review by cpuguy83 — CHANGES_REQUESTED** [review]

File: CONTRIBUTING.md (hunk 1 above)
Comment (squillace) [reply CONTRIBUTING.md]: "subject to" is the proper form; both suggested changes are off by one grammar oddity. :-)

has not been.... subject to.

File: CONTRIBUTING.md [hunk 2]
```diff
@@ -135,6 +135,30 @@ Most features will likely have most of the code in the `containerd-shim-wasm` pr
...
+ - If a shim runtime has not been maintained for 6 months it will be subject for removal
+ - If required changes to the runtime can't be merged or not supported by runtime maintainers.
+ - If it falls behind in new features added to the `coantainerd-shim-wasm` due to lack of maintainance
+
+Before removal:
+- We will create an issue in the repository, pinned to the top. 
+- Send notification in our slack channel 
+- make best effort to contact the maintainers (agreed to when adding the shim)
+
+After 1 month of the issue being up, if not maintainer is found we will remove the shim from the project.
```
rumpl, line 160:
```suggestion
After 1 month of the issue being up, if no maintainer is found we will remove the shim from the project.
```

CONTRIBUTING.md
```diff
@@ -135,6 +135,30 @@ Most features will likely have most of the code in the `containerd-shim-wasm` pr
...
+
+We welcome new shims, though you can also host them in your own repositories as well and use the `coantainerd-shim-wasm` crate.  We recognize that the project is moving fast and having them in this repository can reduce the need for changes as well for discoverability.  
+
+Please open an issue before submitting a PR for discussion to make sure it is a good fit.  As a general rule, we want shims to be adopting WASI spec (this is after all called run*wasi* :-)). In the future we may require shims in the repository to pass WASI compliance tests when the standards mature more. See https://github.com/containerd/runwasi/issues/338 for more discussion.
+
+Since we are not experts in every runtime we also need a commitment from the runtime owners to contribute to maintenance of the shim.
+
+## Removing Shims
+
+This is a fast moving space, with lots of innovation happening and some shims may eventually need to be removed.
```
Mossaka, line 148: One question that I brough up in the community call today is how to deal with scenarios where immediate actions are required. For example, say we introduced a breaking change in the core crate and we only implement that in one of the shims, and then obviously the CI/CD pipelines are going to break for other runtimes. How to resolve this scenario so that we can quickly iterate through new features and experiments?

CONTRIBUTING.md
```diff
@@ -135,6 +135,30 @@ Most features will likely have most of the code in the `containerd-shim-wasm` pr
...
+ - If a shim runtime has not been maintained for 6 months it will be subject for removal
+ - If required changes to the runtime can't be merged or not supported by runtime maintainers.
+ - If it falls behind in new features added to the `coantainerd-shim-wasm` due to lack of maintainance
+
+Before removal:
+- We will create an issue in the repository, pinned to the top. 
+- Send notification in our slack channel 
+- make best effort to contact the maintainers (agreed to when adding the shim)
+
+After 1 month of the issue being up, if no maintainer is found we will remove the shim from the project.
```
Mossaka, line 160: I am intepreting "remove the shim" as deleting the code from the runwasi repo. 

Alternatively, we can archive the not-maintained shim by changing the dependency of `containerd-shim-wasm` crate from a local path to the last released version where the shim is still working. This way we still host the source code of the shim in the main branch and it has visibility to folks checking out the repo, but effectively archiving the shim in a way that it's not going to get the new features from the core crate. 

The downside of that is other dependencies the shim has may or may not be out-of-date and potentially introduce CVEs and as maintainers, we still need to bump the deps for security reasons and that will take time and efforts to maintain.

Mossaka (commented):
Great work! Added the comments I was mentioning in the community call.

CONTRIBUTING.md
jsturtevant: Since we tag commits, the code is affectively archived via git in a way that someone can find it easily. Code is never fully removed. The older releases will also have the built binaries and pointers to the code.

Is the timing here the concern?  Or recognition? something else?  Timing wise, as it is worded now there is 7 months from time we see issue to deleting code.  This is a pretty generous amount of time IMO.

For recognition, maybe we can have a spot in the readme that points to "retired" shims?

CONTRIBUTING.md
jsturtevant: Great point. I will add something about this. My take is it would be case by case basis but if we can't resolve something and it is blocking then disabling and opening an issue to track (and potentially kick off the removal if no responses) fixing it.

CONTRIBUTING.md
Mossaka: > maybe we can have a spot in the readme that points to "retired" shims?

I like this

CONTRIBUTING.md
jsturtevant: fixed

CONTRIBUTING.md
jsturtevant: added a section for this

CONTRIBUTING.md
jsturtevant: I didn't add this explicitly to the doc but I think its a great idea to have a section to recognize older shims.

Mossaka (approved):

cpuguy83 (approved):

PR #434, opened 2023-12-28
Fix image name of test-image/oci

sachaos:
The image name expected in https://github.com/containerd/runwasi/blob/ec3d24d9abf0ff007ab56d07fbe040b1753d6b20/README.md#L289 is `ghcr.io/containerd/runwasi/wasi-demo-oci`.
But the image name created by `make test-image/oci` is `wasi-demo-app`.
So I've fixed the image name to `wasi-demo-oci`.

Signed-off-by: Takumasa Sakao <tsakao@zlab.co.jp>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -234,6 +234,13 @@ test/k8s-%: test/k8s/clean test/k8s/cluster-%
 	sleep 5s
 	kubectl --context=kind-$(KIND_CLUSTER_NAME) wait deployment wasi-demo --for condition=Available=True --timeout=5s
 
+.PHONY: test/k8s-oci-%
+test/k8s-oci-%: test/k8s/clean test/k8s/cluster-oci-%
+	kubectl --context=kind-$(KIND_CLUSTER_NAME) apply -f test/k8s/deploy.oci.yaml
+	kubectl --context=kind-$(KIND_CLUSTER_NAME) wait deployment wasi-demo --for condition=Available=True --timeout=90s
+	# verify that we are still running after some time
+	sleep 5s
+	kubectl --context=kind-$(KIND_CLUSTER_NAME) wait deployment wasi-demo --for condition=Available=True --timeout=5s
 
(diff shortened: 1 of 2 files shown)


Mossaka (approved):
lgtm, but I think the CI is failing

sachaos:
@Mossaka Thank you for your review!
I've fixed the CI. 😄

Makefile
```diff
@@ -234,6 +234,13 @@ test/k8s-%: test/k8s/clean test/k8s/cluster-%
 	sleep 5s
 	kubectl --context=kind-$(KIND_CLUSTER_NAME) wait deployment wasi-demo --for condition=Available=True --timeout=5s
 
+.PHONY: test/k8s-oci-%
+test/k8s-oci-%: test/k8s/clean test/k8s/cluster-oci-%
```
jprendes, line 238: Could we add this to the CI workflow?

sachaos: @jprendes Thank you for the question!
I think this is added to the CI workflow already.
We can see this is executed in the following step of the workflow.
https://github.com/containerd/runwasi/actions/runs/7442612789/job/20246359284?pr=434#step:7:1

Mossaka (approved):
Lgtm!

Makefile
jsturtevant: I think the make target above should cover this `test/k8s-%: test/k8s/clean test/k8s/cluster-%`.

jprendes: Oh, I see, before CI was triggering `test/k8s-%` and now it's triggering `test/k8s-oci-%`.

Thanks!

jprendes:
> I think this should have been an update to the README not the make files.

I prefer the current change. IIUC, before both (distinct) images were using the same name, right?  I think it's cleaner to give them different names, specially for local development.

jsturtevant:
works for me

PR #448, opened 2024-01-12
refactor(*): refactor wasm bytes preparation logic to the core crate

Mossaka:
This PR refactors the common `Source` logic from runtime shim to the core crate. To do so, the core `Engine` trait API has a breaking change:

1. `run_wasi()` now takes a `wasm_bytes: &[u8]` parameter as the wasm module / component loaded from the file system or OCI layers to execute. 

cc @jsturtevant

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// The name to use for this engine
     fn name() -> &'static str;
     /// Run a WebAssembly container
-    fn run_wasi(&self, ctx: &impl RuntimeContext, stdio: Stdio) -> Result<i32>;
+    fn run_wasi(&self, ctx: &impl RuntimeContext, wasm_bytes: &[u8], stdio: Stdio) -> Result<i32>;
     /// Check that the runtime can run the container.
     /// This checks runs after the container creation and before the container starts.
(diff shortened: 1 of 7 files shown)


jprendes:
I think I would prefer if `Source` provided a method called `to_bytes`, that would encapsulate this logic, rather than changing the `run_wasi` signature.

jsturtevant:
The background is that we had something similiar to `to_bytes` but this had a clone https://github.com/containerd/runwasi/pull/401#discussion_r1414601347.  Maybe I've got it wrong in that thread about trying to avoid a clone on that call?

I am also not a fan of these changes to the `run_wasi` trait

jprendes:
The clone could be avoided returning a `Cow<[u8]>`  IIUC.

jprendes:
Alternatively, in the case of `File` the bytes could be cached to avoid reading the file multiple times (something like a `OnceCell` or `OnceLock`), and then the `&[u8]` would refer to the bytes owned by the cache.

Mossaka:
@jprendes thanks for the suggestion of `Cow`! I re-implement the refactor in https://github.com/containerd/runwasi/pull/454

Mossaka:
Close in favor of #454

PR #454, opened 2024-01-18
refactor(*): add source.as_bytes()

Mossaka:
This PR is a preferred way to refactor the usage of `Source` into a `as_bytes()` function in Context. It does not change the APIs of the Engine trait. Closes #448

### DIFF
--- a/crates/containerd-shim-wasm/src/container/context.rs
+++ b/crates/containerd-shim-wasm/src/container/context.rs
@@ -1,8 +1,11 @@
+use std::borrow::Cow;
 use std::path::{Path, PathBuf};
 
+use anyhow::{bail, Context};
 use oci_spec::image::Platform;
 use oci_spec::runtime::Spec;
 
+use crate::container::path::PathResolve;
 use crate::sandbox::oci::WasmLayer;
 
 pub trait RuntimeContext {
@@ -42,6 +45,24 @@ pub enum Source<'a> {
     // and they will be included in this array, e.g., a `toml` file with the
     // runtime configuration.
     Oci(&'a [WasmLayer]),
+}
+
+impl<'a> Source<'a> {
+    pub fn as_bytes(&self) -> anyhow::Result<Cow<'a, [u8]>> {
+        match self {
+            Source::File(path) => {
+                let path = path
+                    .resolve_in_path_or_cwd()
+                    .next()
+                    .context("module not found")?;
+                Ok(Cow::Owned(std::fs::read(path)?))
+            }
+            Source::Oci([module]) => Ok(Cow::Borrowed(&module.layer)),
+            Source::Oci(_modules) => {
+                bail!("only a single module is supported when using images with OCI layers")
+            }
+        }
+    }
 }
 
 /// The entrypoint for a WASI module / component.
--- a/crates/containerd-shim-wasmtime/src/instance.rs
+++ b/crates/containerd-shim-wasmtime/src/instance.rs
@@ -2,7 +2,7 @@ use std::fs::File;
 
 use anyhow::{bail, Context, Result};
 use containerd_shim_wasm::container::{
-    Engine, Entrypoint, Instance, PathResolve, RuntimeContext, Source, Stdio, WasmBinaryType,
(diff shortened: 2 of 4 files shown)


crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -76,40 +77,12 @@ impl Engine for WasmtimeEngine {
...
-                }
-            }
-            Source::Oci(_modules) => {
-                bail!("only a single module is supported when using images with OCI layers")
-            }
+        let cow = source.into_cow()?;
+        let wasm_bytes = cow.deref();
+        let status = match WasmBinaryType::from_bytes(wasm_bytes) {
+            Some(WasmBinaryType::Module) => self.execute_module(wasm_bytes, store, &func)?,
+            Some(WasmBinaryType::Component) => self.execute_component(wasm_bytes, store, func)?,
```
jsturtevant, line 84: Since COW implements `Deref` you can simplify this to the following which reads a little cleaner.

```suggestion
        let wasm_bytes= source.into_bytes()?;
        let status = match WasmBinaryType::from_bytes(&wasm_bytes) {
            Some(WasmBinaryType::Module) => self.execute_module(&wasm_bytes, store, &func)?,
            Some(WasmBinaryType::Component) => self.execute_component(&wasm_bytes, store, func)?,
```

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -44,6 +47,24 @@ pub enum Source<'a> {
     Oci(&'a [WasmLayer]),
 }
 
+impl<'a> Source<'a> {
+    pub fn into_cow(self) -> anyhow::Result<Cow<'a, Vec<u8>>> {
```
jsturtevant, line 51: with comment below rename this: `into_bytes`

jsturtevant (changes requested):

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -44,6 +47,24 @@ pub enum Source<'a> {
     Oci(&'a [WasmLayer]),
 }
 
+impl<'a> Source<'a> {
+    pub fn into_bytes(self) -> anyhow::Result<Cow<'a, Vec<u8>>> {
```
jprendes, line 51: Why do we need to consume self?
Could this be `as_bytes(&self)` instead?

Mossaka: Ah you're right we don't

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -44,6 +47,24 @@ pub enum Source<'a> {
     Oci(&'a [WasmLayer]),
 }
 
+impl<'a> Source<'a> {
+    pub fn as_bytes(&self) -> anyhow::Result<Cow<'a, Vec<u8>>> {
```
jsturtevant, line 51: Changing this signature to `pub fn as_bytes(&self) -> anyhow::Result<Cow<'a, [u8]>> {`  will remove the need to do `&*wasm_bytes`  below in the wasmedge instance

jsturtevant (approved):
LGTM

crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -44,6 +47,24 @@ pub enum Source<'a> {
...
 
+impl<'a> Source<'a> {
+    pub fn as_bytes(&self) -> anyhow::Result<Cow<'a, [u8]>> {
+        match self {
+            Source::File(path) => {
+                let path = path
+                    .resolve_in_path_or_cwd()
+                    .next()
+                    .context("module not found")?;
+                Ok(Cow::Owned(std::fs::read(path)?))
```
jprendes, line 58: **thought:** We optimize for no copy in the `Oci` case by using `Cow`, but we read the file every time in the `File` case. Might be premature optimiation, but the file content could be cached in memory here to avoid multiple reads.
I'm fine with leaving it as it is because the unoptimized `File` case would be the _legacy_ workflow. Also we could revisit if we ever observe any perforance bottleneck here.

jprendes (approved):
LGTM

Issue #447, opened 2024-01-10
The newest kindest node may have the OCI containerd bits


Mossaka:

jsturtevant:
kind version https://github.com/kubernetes-sigs/kind/releases/tag/v0.20.0 does not have the correct containerd version.  I would suspect there would be containerd version bump before the do another release.

PR #472, opened 2024-02-02
Update kind and use for oci as well

jsturtevant:
Kind 0.21 came out: https://github.com/kubernetes-sigs/kind/releases/tag/v0.21.0

This has the correct containerd version for using oci images, so we can get rid of the custom kind image.


Mossaka (approved):
Nice!

Mossaka:
Close #447

PR #492, opened 2024-02-21
Pre-compile each layer individually

jsturtevant:
After some feedback we found some runtimes require ability to individually pre-compile layers.  This attempts to address that ability. 

This also fixes a bug that was introduced when adding pre-compilation where the media types were being overwritten.

### DIFF
--- a/crates/containerd-shim-wasm/src/container/engine.rs
+++ b/crates/containerd-shim-wasm/src/container/engine.rs
@@ -5,6 +5,7 @@ use anyhow::{bail, Context, Result};
 
 use super::Source;
 use crate::container::{PathResolve, RuntimeContext};
+use crate::sandbox::oci::WasmLayer;
 use crate::sandbox::Stdio;
 
 pub trait Engine: Clone + Send + Sync + 'static {
@@ -57,8 +58,8 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// This is used to precompile a module before it is run and will be called if can_precompile returns true.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
     /// The cached, precompiled module will be reloaded on subsequent runs.
-    fn precompile(&self, _layers: &[Vec<u8>]) -> Result<Vec<u8>> {
-        bail!("precompilation not supported for this runtime")
+    fn precompile(&self, _layer: &WasmLayer) -> Option<Result<Vec<u8>>> {
+        None
     }
 
     /// Can_precompile lets the shim know if the runtime supports precompilation.
--- a/crates/containerd-shim-wasm/src/container/engine.rs
+++ b/crates/containerd-shim-wasm/src/container/engine.rs
@@ -58,7 +58,7 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// This is used to precompile a module before it is run and will be called if can_precompile returns true.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
     /// The cached, precompiled module will be reloaded on subsequent runs.
-    fn precompile(&self, _layer: &WasmLayer) -> Option<Result<Vec<u8>>> {
+    fn precompile(&self, _layers: &[WasmLayer]) -> Option<Result<Vec<WasmLayer>>> {
         None
     }
 
--- a/crates/containerd-shim-wasm/src/container/engine.rs
+++ b/crates/containerd-shim-wasm/src/container/engine.rs
@@ -1,7 +1,7 @@
 use std::fs::File;
 use std::io::Read;
 
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 
 use super::Source;
 use crate::container::{PathResolve, RuntimeContext};
@@ -58,10 +58,10 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// This is used to precompile the layers before they are run and will be called if can_precompile returns true.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
     /// The cached, precompiled layers will be reloaded on subsequent runs.
-    /// The runtime is expected to runtime the same number of layers passed in.
-    /// In some cases it is possible in some edge cases that the layers may already be precompiled and the runtime can return the layers as is.  
-    fn precompile(&self, _layers: &[WasmLayer]) -> Option<Result<Vec<WasmLayer>>> {
-        None
+    /// The runtime is expected to runtime the same number of layers passed in, if the layer cannot be precompiled it should return None for that layer.
+    /// In some cases it is possible in some edge cases that the layers may already be precompiled and None should be returned in this case.
+    fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<WasmLayer>>> {
+        bail!("precompile not supported");
     }
 
     /// Can_precompile lets the shim know if the runtime supports precompilation.
--- a/crates/containerd-shim-wasm/src/container/engine.rs
+++ b/crates/containerd-shim-wasm/src/container/engine.rs
@@ -60,7 +60,7 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// The cached, precompiled layers will be reloaded on subsequent runs.
     /// The runtime is expected to runtime the same number of layers passed in, if the layer cannot be precompiled it should return None for that layer.
     /// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
-    fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<WasmLayer>>> {
+    fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<Vec<u8>>>> {
         bail!("precompile not supported");
     }
 
--- a/crates/containerd-shim-wasm/src/container/engine.rs
+++ b/crates/containerd-shim-wasm/src/container/engine.rs
@@ -1,7 +1,7 @@
 use std::fs::File;
 use std::io::Read;
 
-use anyhow::{bail, Context, Result};
+use anyhow::{Context, Result};
 
 use super::Source;
 use crate::container::{PathResolve, RuntimeContext};
--- a/crates/containerd-shim-wasm/src/container/engine.rs
+++ b/crates/containerd-shim-wasm/src/container/engine.rs
@@ -54,10 +54,12 @@ pub trait Engine: Clone + Send + Sync + 'static {
         &["application/vnd.bytecodealliance.wasm.component.layer.v0+wasm"]
     }
 
-    /// Precompiles a module that is in the WASM OCI layer format
-    /// This is used to precompile a module before it is run and will be called if can_precompile returns true.
+    /// Precompiles a passes supported OCI layers to engine for compilation
+    /// This is used to precompile the layers before they are run and will be called if can_precompile returns true.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
-    /// The cached, precompiled module will be reloaded on subsequent runs.
+    /// The cached, precompiled layers will be reloaded on subsequent runs.
+    /// The runtime is expected to runtime the same number of layers passed in.
+    /// In some cases it is possible in some edge cases that the layers may already be precompiled and the runtime can return the layers as is.  
     fn precompile(&self, _layers: &[WasmLayer]) -> Option<Result<Vec<WasmLayer>>> {
         None
     }
--- a/crates/containerd-shim-wasm/src/container/engine.rs
+++ b/crates/containerd-shim-wasm/src/container/engine.rs
@@ -59,7 +59,7 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
     /// The cached, precompiled layers will be reloaded on subsequent runs.
     /// The runtime is expected to runtime the same number of layers passed in, if the layer cannot be precompiled it should return None for that layer.
-    /// In some cases it is possible in some edge cases that the layers may already be precompiled and None should be returned in this case.
+    /// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
     fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<WasmLayer>>> {
         bail!("precompile not supported");
     }
--- a/crates/containerd-shim-wasm/src/container/engine.rs
+++ b/crates/containerd-shim-wasm/src/container/engine.rs
@@ -54,11 +54,11 @@ pub trait Engine: Clone + Send + Sync + 'static {
         &["application/vnd.bytecodealliance.wasm.component.layer.v0+wasm"]
     }
 
-    /// Precompiles a passes supported OCI layers to engine for compilation
-    /// This is used to precompile the layers before they are run and will be called if can_precompile returns true.
+    /// Precompile passes supported OCI layers to engine for compilation
+    /// This is used to precompile the layers before they are run and will be called if `can_precompile` returns `true`.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
     /// The cached, precompiled layers will be reloaded on subsequent runs.
-    /// The runtime is expected to runtime the same number of layers passed in, if the layer cannot be precompiled it should return None for that layer.
+    /// The runtime is expected to return the same number of layers passed in, if the layer cannot be precompiled it should return `None` for that layer.
     /// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
     fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<Vec<u8>>>> {
         bail!("precompile not supported");
--- a/crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
@@ -409,34 +409,26 @@ impl Client {
             None => (false, "".to_string()),
         };
 
-        match image.labels.get(&precompile_id) {
-            Some(precompile_digest) if can_precompile => {
-                log::info!("found precompiled label: {} ", &precompile_id);
-                match self.read_content(precompile_digest) {
-                    Ok(precompiled) => {
-                        log::info!("found precompiled module in cache: {} ", &precompile_digest);
-                        return Ok((
-                            vec![WasmLayer {
-                                config: image_config_descriptor.clone(),
-                                layer: precompiled,
-                            }],
-                            platform,
-                        ));
-                    }
-                    Err(e) => {
-                        // log and continue
-                        log::warn!("failed to read precompiled module from cache: {}. Content may have been removed manually, will attempt to recompile", e);
-                    }
-                }
-            }
-            _ => {}
-        }
-
+        let needs_precompile = can_precompile && !image.labels.contains_key(&precompile_id);
         let layers = manifest
             .layers()
             .iter()
             .filter(|x| is_wasm_layer(x.media_type(), T::supported_layers_types()))
-            .map(|config| self.read_content(config.digest()))
+            .map(|config| {
+                
+                let mut digest = config.digest().clone();
+                if can_precompile {
+                    let info = self.get_info(config.digest().clone())?;
+                    if info.labels.contains_key(&precompile_id) {
+                        log::info!("found precompiled layer in cache: {} ", &precompile_id);
+                        digest = info.labels.get(&precompile_id).unwrap().clone();
+                    }
+                }
+                self.read_content(digest).map(|module| WasmLayer {
+                    config: config.clone(),
+                    layer: module,
+                })
+            })
             .collect::<Result<Vec<_>>>()?;
 
         if layers.is_empty() {
@@ -444,47 +436,53 @@ impl Client {
             return Ok((vec![], platform));
         }
 
-        if can_precompile {
-            log::info!("precompiling module");
-            let precompiled = engine.precompile(layers.as_slice())?;
-            log::info!("precompiling module: {}", image_digest.clone());
-            let precompiled_content =
-                self.save_content(precompiled.clone(), image_digest.clone(), &precompile_id)?;
+        if needs_precompile {
+            log::info!("precompiling layers for image: {}", image.name);
+            
+            let mut precompiled_layers = Vec::new();
+            for (i, layer) in layers.iter().enumerate() {
+                let precompiled_layer = match engine.precompile(layer) {
+                    Some(it) => {
+                        let precompiled_layer = it?;
+                        precompiled_layers.push(WasmLayer {
+                            config: layer.config.clone(),
+                            layer: precompiled_layer.clone(),
+                        });
+                        precompiled_layer
+                    },
+                    None =>{
+                        // skip layers that don't support precompilation, and add them to the list of precompiled layers
+                         precompiled_layers.push(layer.clone()); 
+                         continue
+                        } 
+                };
 
-            log::debug!("updating image with compiled content digest");
-            image
-                .labels
-                .insert(precompile_id, precompiled_content.digest.clone());
-            self.update_image(image)?;
+                let precompiled_content = self.save_content(precompiled_layer, image_digest.clone(), &precompile_id)?;
 
-            // The original image is considered a root object, by adding a ref to the new compiled content
-            // We tell containerd to not garbage collect the new content until this image is removed from the system
-            // this ensures that we keep the content around after the lease is dropped
-            log::debug!("updating content with precompile digest to avoid garbage collection");
-            let mut image_content = self.get_info(image_digest.clone())?;
-            image_content.labels.insert(
-                "containerd.io/gc.ref.content.precompile".to_string(),
-                precompiled_content.digest.clone(),
-            );
-            self.update_info(image_content)?;
+                log::debug!("updating image with indicator that precompiled content is available");
+                image.labels
+                    .insert(precompile_id.clone(), "true".to_string());
+                self.update_image(image.clone())?;
+
+                // The original image is considered a root object, by adding a ref to the new compiled content
+                // We tell containerd to not garbage collect the new content until this image is removed from the system
+                // this ensures that we keep the content around after the lease is dropped
+                log::debug!("updating content with precompile digest to avoid garbage collection");
+                let mut image_content = self.get_info(image_digest.clone())?;
+                image_content.labels.insert(
+                    format!("containerd.io/gc.ref.content.precompile.{}",i),
+                    precompiled_content.digest.clone(),
+                );
+                self.update_info(image_content)?;
+            }
 
             return Ok((
-                vec![WasmLayer {
-                    config: image_config_descriptor.clone(),
-                    layer: precompiled,
-                }],
+                precompiled_layers,
                 platform,
             ));
         }
 
         log::info!("using module from OCI layers");
-        let layers = layers
-            .into_iter()
-            .map(|module| WasmLayer {
-                config: image_config_descriptor.clone(),
-                layer: module,
-            })
-            .collect::<Vec<_>>();
         Ok((layers, platform))
     }
 }
--- a/crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
@@ -415,26 +415,26 @@ impl Client {
             .iter()
             .filter(|x| is_wasm_layer(x.media_type(), T::supported_layers_types()))
             .map(|original_config| {
-                let mut digest = original_config.digest().clone();
+                let mut digest_to_load = original_config.digest().clone();
                 if can_precompile {
-                    let info = self.get_info(digest.clone())?;
+                    let info = self.get_info(digest_to_load.clone())?;
                     if info.labels.contains_key(&precompile_id) {
-                        digest = info.labels.get(&precompile_id).unwrap().clone();
+                        digest_to_load = info.labels.get(&precompile_id).unwrap().clone();
                         log::info!(
                             "layer {} has pre-compiled content: {} ",
                             info.digest,
-                            &digest
+                            &digest_to_load
                         );
                     }
                 }
-                log::debug!("loading digest: {} ", &digest);
-                self.read_content(digest.clone())
+                log::debug!("loading digest: {} ", &digest_to_load);
+                self.read_content(digest_to_load.clone())
                     .map(|module| WasmLayer {
                         config: original_config.clone(),
                         layer: module,
                     })
                     .or_else(|e| {
-                        if digest != *original_config.digest() {
+                        if digest_to_load != *original_config.digest() {
                             log::error!("failed to load precompiled layer: {}", e);
                             log::error!("falling back to original layer and marking for recompile");
                             needs_precompile = true;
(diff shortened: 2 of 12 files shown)


kate-goldenring:
@jsturtevant I have tested these changes out with a branch of the Spin containerd shim with pre-compilation support and it works great. I've tested Spin apps with multiple components and static assets, as well. To me, this is looking ready to take out of draft. Thank you for expanding the implementation!

kate-goldenring:
@jsturtevant through more testing, I did find some odd behavior. After pulling an app, I can rerun the shim over and over and get expected behavior (reuses precompiled components); however, if i repull the image (`ctr image pull`) and try to run (`ctr run`) it fails to execute. It looks like after the image is repulled, `runwasi` is erroneously asking the spin shim to precompile again despite knowing it already was precompiled. Then the shim fails, potentially because it is trying to precompile cwasm at that point:

```sh
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T21:17:51.312229693Z" level=info msg="Shim successfully started, waiting for exit signal..."
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T21:17:51.317710774Z" level=info msg="found manifest with WASM OCI image format"
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T21:17:51.318314715Z" level=info msg="layer sha256:35dd7338635af15592327986c45e75efc2b258f1d92e4282ced7668d827e5741 has pre-compiled content: sha256:8d7374edd04e514d5c7b35a1b4a2a334716e1bcae596dfc0944e23b921bf6cc6 "
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T21:17:51.323980565Z" level=info msg="layer sha256:65bc286f8315746d1beecd2430e178f539fa487ebf6520099daae09a35dbce1d has pre-compiled content: sha256:dabc42e0d769727d23126b7908e0ea9dd4a93a7cd8867cb36615cbc990776ba8 "
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T21:17:51.335170453Z" level=info msg="precompiling layers for image: ghcr.io/kate-goldenring/spin-hello-and-kv-explorer:latest"
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T21:17:51.335195337Z" level=info msg="Precompiling layer with Spin Engine: "sha256:35dd7338635af15592327986c45e75efc2b258f1d92e4282ced7668d827e5741""
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T21:17:51.335225245Z" level=info msg="Precompiling layer "sha256:35dd7338635af15592327986c45e75efc2b258f1d92e4282ced7668d827e5741""
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T13:17:51.502392136-08:00" level=info msg="shim disconnected" id=two namespace=default
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T13:17:51.502458979-08:00" level=warning msg="cleaning up after shim disconnected" id=two namespace=default
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T13:17:51.502475509-08:00" level=info msg="cleaning up dead shim" namespace=default
Feb 23 13:17:51 kagold-ThinkPad-X1-Carbon-6th containerd[496568]: time="2024-02-23T13:17:51.507305203-08:00" level=error msg="copy shim log" error="read /proc/self/fd/13: file already closed" namespace=default
```

jsturtevant:
@kate-goldenring did anything in the image change that caused the image sha to change?

kate-goldenring:
@jsturtevant the layers of the image are the same to `ctr` but for some reason runwasi is seems to be getting different layers. Checking if it is already precompile in the shim is an optional workaround that gets it working again:
```rust
    fn precompile(&self, layer: &WasmLayer) -> Option<Result<Vec<u8>>> {
        log::info!(
            "Precompiling layer with Spin Engine: {:?}",
            layer.config.digest()
        );

        match layer.config.media_type() {
            MediaType::Other(name) => {
                log::info!("Precompiling layer {:?}", layer.config.digest());
                if name == "application/vnd.wasm.content.layer.v1+wasm" {
                    if let Some(_) = self.wasmtime_engine.detect_precompiled(&layer.layer) {
                        log::info!("Already precompiled");
                        return None
                    }
                    let component =
                        spin_componentize::componentize_if_necessary(&layer.layer).unwrap();
                    Some(self.wasmtime_engine.precompile_component(&component))
                } else {
                    None
                }
            }
            _ => None,
        }
    }
```

jsturtevant:
ah, I see so we are passing the precompiled bits to the engine to recompile.  Makes sense it will crash, Let me see if we can fix it up in the shim library code but being defensive here might not be a bad option.

kate-goldenring:
@jsturtevant to give more context, when evaluating whether to recompile(`let mut needs_precompile = can_precompile && !image.labels.contains_key(&precompile_id);`), after logging, it is clear that ` !image.labels.contains_key(&precompile_id);` is evaluating to false despite the fact that the labels are correctly displayed on the image during a `ctr content ls`

jsturtevant:
> The change is pretty minimal on the shim implementor side.

6be79082379979b7e9d56bcd38c93e4bf64cde75

radu-matei:
To @kate-goldenring's point:

> @jsturtevant to give more context, when evaluating whether to recompile(`let mut needs_precompile = can_precompile && !image.labels.contains_key(&precompile_id);`), after logging, it is clear that `image.labels.contains_key(&precompile_id);` is evaluating to false despite the fact that the labels are correctly displayed on the image during a `ctr content ls`.

Any thouhgts why this would be happening? Nothing about the image changes in between pulls.

jsturtevant:
> To @kate-goldenring's point:
> 
> > @jsturtevant to give more context, when evaluating whether to recompile(`let mut needs_precompile = can_precompile && !image.labels.contains_key(&precompile_id);`), after logging, it is clear that `image.labels.contains_key(&precompile_id);` is evaluating to false despite the fact that the labels are correctly displayed on the image during a `ctr content ls`.
> 
> Any thouhgts why this would be happening? Nothing about the image changes in between pulls.

looking into this today

jsturtevant:
> @jsturtevant to give more context, when evaluating whether to recompile(`let mut needs_precompile = can_precompile && !image.labels.contains_key(&precompile_id);`), after logging, it is clear that ` image.labels.contains_key(&precompile_id);` is evaluating to false despite the fact that the labels are correctly displayed on the image during a `ctr content ls`

I am seeing the label on the image disappear when using `ctr i pull`.  This doesn't seem to be the case when we use `ctr import` which is used by the tests.  Working on a fix and a test to cover the scenario.

jsturtevant:
I've stored the "pre-compiled" flag on the content instead of the image, since the image is mutable (I think @cpuguy83 pointed this out previously but I didn't fully understand the implications at the time).

Note that if runwasi detects that one or more of the pre-compiled components are removed from the cache it will re-request a compilation so its good to keep the check for compilation in the shim. I don't really know a valid scenario where this would happen besides a user going in and deleting it manually.  I will see if we can come up with a way to handle this when we improve the api towards `process_layers`.

cpuguy83:
The image reference (which we read from the container object) is mutable, is what I was referring to.
But yes I think its still best to have it on the content that is being compiled rather than the image itself which is often referring to an image index and not a specific image manifest.

jsturtevant:
I was looking into the why the test failed here, I was able to reproduce it locally although flaky and identified that it is due to the way the test is importing images.  In the failed test the image gets imported twice, I expected this to be a no-op but the annotations in the image are stored in a hashmap which results in a different image digest causing the flake (since the flag isn't found on the content but the individual layers are the same).  

This comes back to my comment in https://github.com/containerd/runwasi/pull/492#issuecomment-1965541517 which is to say that we are not passing information around if we know whether it is compiled to the shim today but can possible address it with a better API.

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -401,90 +377,116 @@ impl Client {
...
+            match engine.precompile(&layers) {
+                Some(compiled_layer_result) => {
+                    let compiled_layers = compiled_layer_result?;
+
+                    for (i, precompiled_layer) in compiled_layers.iter().enumerate() {
+                        let original_layer = &layers[i];
+
+                        let precompiled_content = self.save_content(
+                            precompiled_layer.layer.clone(),
+                            original_layer.config.digest(),
```
kate-goldenring, line 444: Since we are never using `precompiled_layer.config`, should we remove it from the API return type to prevent a mismatch of config state or someone using the API and assuming they can modify a layer's config. Instead it could be:
```rust
fn precompile(&self, _layers: &[WasmLayer]) -> Option<Result<Vec<Vec<u8>>>> {
        None
    }
```

kate-goldenring (commented):
@jsturtevant I just have some preliminary questions on the `precompile` API

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -401,90 +377,116 @@ impl Client {
...
-            log::info!("no WASM modules found in OCI layers");
-            return Ok((vec![], platform));
+        if needs_precompile {
+            log::info!("precompiling layers for image: {}", image.name);
+            match engine.precompile(&layers) {
+                Some(compiled_layer_result) => {
+                    let compiled_layers = compiled_layer_result?;
+
+                    for (i, precompiled_layer) in compiled_layers.iter().enumerate() {
+                        let original_layer = &layers[i];
```
kate-goldenring, line 440: This line assumes that all N layers are returned even if only M were precompiled (say a few layers were not wasm).

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -53,12 +54,14 @@ pub trait Engine: Clone + Send + Sync + 'static {
...
+    /// This is used to precompile the layers before they are run and will be called if can_precompile returns true.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
-    /// The cached, precompiled module will be reloaded on subsequent runs.
-    fn precompile(&self, _layers: &[Vec<u8>]) -> Result<Vec<u8>> {
-        bail!("precompilation not supported for this runtime")
+    /// The cached, precompiled layers will be reloaded on subsequent runs.
+    /// The runtime is expected to runtime the same number of layers passed in.
+    /// In some cases it is possible in some edge cases that the layers may already be precompiled and the runtime can return the layers as is.  
+    fn precompile(&self, _layers: &[WasmLayer]) -> Option<Result<Vec<WasmLayer>>> {
+        None
```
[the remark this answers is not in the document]
radu-matei: IIUC, this addresses the fact that some OCI artifacts can contain layers that are required to be passes to the engine, but are not Wasm components, so they could choose to not precompile.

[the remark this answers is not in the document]
kate-goldenring: Wouldn't that be `Result<Vec<Option<WasmLayer>>>`? Or maybe i am misunderstanding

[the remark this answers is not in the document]
jsturtevant: @kate-goldenring Thanks, I think this makes more sense than the way I am doing it currently approaching it

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
jsturtevant: For now, I want avoid handling the situation where multple layers are compiled to 1.  I think this is going to need some bigger changes and would like the design to evolve it to a different api like `process_layers` or `initialize_layers` with more input for folks.  If I use `Result<Vec<Option<WasmLayer>>>` as you suggested, I can avoid looking up the layer here and just skip if `None`

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
jsturtevant: I agree that need to come up with a way to say which layers its associated with but for now I think we can handle a single layer.

crates/containerd-shim-wasm/src/container/engine.rs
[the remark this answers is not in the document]
jsturtevant: I've updated this

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
kate-goldenring: If we aren't using the precompiled layer config at all, then we should change the API to just return the content:
```
fn precompile(&self, layers: &[WasmLayer]) -> Result<Vec<Option<Vec,u8>>>>
```
What is your thought around returning the entire `WasmLayer`?

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -58,10 +58,10 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// This is used to precompile the layers before they are run and will be called if can_precompile returns true.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
     /// The cached, precompiled layers will be reloaded on subsequent runs.
-    /// The runtime is expected to runtime the same number of layers passed in.
-    /// In some cases it is possible in some edge cases that the layers may already be precompiled and the runtime can return the layers as is.  
-    fn precompile(&self, _layers: &[WasmLayer]) -> Option<Result<Vec<WasmLayer>>> {
-        None
+    /// The runtime is expected to runtime the same number of layers passed in, if the layer cannot be precompiled it should return None for that layer.
+    /// In some cases it is possible in some edge cases that the layers may already be precompiled and None should be returned in this case.
```
kate-goldenring, line 62:
```suggestion
/// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
```

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -401,90 +382,122 @@ impl Client {
...
-                    Err(e) => {
-                        // log and continue
-                        log::warn!("failed to read precompiled module from cache: {}. Content may have been removed manually, will attempt to recompile", e);
-                    }
-                }
-            }
-            _ => {}
-        }
-
+        let image_info = self.get_info(&image_digest)?;
```
kate-goldenring, line 393: Shouldn't this be checking content rather than the image?

One thing I ran into (and it also occurs with the "precompile each layer individually" API d8003bd) is that if i update a spin app to add another component, runwasi sees that the image needs precompilation again. I am assuming that is because the image digest changed so it doesn't see the label here.

This results in the Spin engine precompiling everything again, since runwasi passes the unprecompiled layers.

kate-goldenring: I see what is happening -- and I don't think there is anything we can do about it. It looks like when you pull (`ctr image pull` ) the updated image, it removed the precompiled layers (even though it retains the labels on the original layers):

After pulling a Spin app with 2 components for the first time:
```
DIGEST									SIZE	AGE	LABELS
sha256:1a2671d57364ae2946fb4e2bae0c15e195644d39280a8bc976656af925935c6f	1.215kB	12 seconds	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer
sha256:2314d31d6459abab359502dcd766bc50f09be84b8b46d4716441bf804f7a9bdf	664B	12 seconds	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer,containerd.io/gc.ref.content.0=sha256:62410ce73ff98c3587580f7659d8dae08ec8f8e86b92f757b59bd52843d3f859,containerd.io/gc.ref.content.1=sha256:65bc286f8315746d1beecd2430e178f539fa487ebf6520099daae09a35dbce1d,containerd.io/gc.ref.content.2=sha256:1a2671d57364ae2946fb4e2bae0c15e195644d39280a8bc976656af925935c6f,containerd.io/gc.ref.content.config=sha256:aabedf0b094944b344cecbffcfa0a9172e905a5d156bdf67c4c6cbb6d27d1d92,containerd.io/gc.ref.content.precompile.0=sha256:f268328f6ef7f92da3950c2d1abc9b07fc240c021b9b3b354319a09c195c0eb0,containerd.io/gc.ref.content.precompile.1=sha256:dabc42e0d769727d23126b7908e0ea9dd4a93a7cd8867cb36615cbc990776ba8,runwasi.io/precompiled/spin/15627766957371465298=true
sha256:62410ce73ff98c3587580f7659d8dae08ec8f8e86b92f757b59bd52843d3f859	1.904MB	11 seconds	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer,runwasi.io/precompiled/spin/15627766957371465298=sha256:f268328f6ef7f92da3950c2d1abc9b07fc240c021b9b3b354319a09c195c0eb0
sha256:65bc286f8315746d1beecd2430e178f539fa487ebf6520099daae09a35dbce1d	546.5kB	11 seconds	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer,runwasi.io/precompiled/spin/15627766957371465298=sha256:dabc42e0d769727d23126b7908e0ea9dd4a93a7cd8867cb36615cbc990776ba8
sha256:aabedf0b094944b344cecbffcfa0a9172e905a5d156bdf67c4c6cbb6d27d1d92	78B	12 seconds	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer
sha256:dabc42e0d769727d23126b7908e0ea9dd4a93a7cd8867cb36615cbc990776ba8	3.422MB	9 secondsrunwasi.io/precompiled/spin/15627766957371465298/original=sha256:65bc286f8315746d1beecd2430e178f539fa487ebf6520099daae09a35dbce1d
sha256:f268328f6ef7f92da3950c2d1abc9b07fc240c021b9b3b354319a09c195c0eb0	1.113MB	10 seconds	runwasi.io/precompiled/spin/15627766957371465298/original=sha256:62410ce73ff98c3587580f7659d8dae08ec8f8e86b92f757b59bd52843d3f859
```

After repulling the app that now has a third component:
```
DIGEST									SIZE	AGE		LABELS
sha256:290359e28f0878cac9793b37077c79337bdfe22e395cc809d4a99a40be2d8341	1.516kB	7 seconds	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer
sha256:478f64c114b95dad0dbe88e9755384e30ba9c8bf62e76b93feceb85311607719	821B	8 seconds	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer,containerd.io/gc.ref.content.0=sha256:62410ce73ff98c3587580f7659d8dae08ec8f8e86b92f757b59bd52843d3f859,containerd.io/gc.ref.content.1=sha256:65bc286f8315746d1beecd2430e178f539fa487ebf6520099daae09a35dbce1d,containerd.io/gc.ref.content.2=sha256:74e33464d5dae4246f077a821dd4c1514c26004e24f2b949ff4d1d95517f49e6,containerd.io/gc.ref.content.3=sha256:290359e28f0878cac9793b37077c79337bdfe22e395cc809d4a99a40be2d8341,containerd.io/gc.ref.content.config=sha256:aabedf0b094944b344cecbffcfa0a9172e905a5d156bdf67c4c6cbb6d27d1d92
sha256:62410ce73ff98c3587580f7659d8dae08ec8f8e86b92f757b59bd52843d3f859	1.904MB	About a minute	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer,runwasi.io/precompiled/spin/15627766957371465298=sha256:f268328f6ef7f92da3950c2d1abc9b07fc240c021b9b3b354319a09c195c0eb0
sha256:65bc286f8315746d1beecd2430e178f539fa487ebf6520099daae09a35dbce1d	546.5kB	About a minute	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer,runwasi.io/precompiled/spin/15627766957371465298=sha256:dabc42e0d769727d23126b7908e0ea9dd4a93a7cd8867cb36615cbc990776ba8
sha256:74e33464d5dae4246f077a821dd4c1514c26004e24f2b949ff4d1d95517f49e6	2.366MB	7 seconds	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer
sha256:aabedf0b094944b344cecbffcfa0a9172e905a5d156bdf67c4c6cbb6d27d1d92	78B	About a minute	containerd.io/distribution.source.ghcr.io=kate-goldenring/spin-hello-and-kv-explorer
```

The precompiled layers (`dabc42e0d769727d23126b7908e0ea9dd4a93a7cd8867cb36615cbc990776ba8` and `f268328f6ef7f92da3950c2d1abc9b07fc240c021b9b3b354319a09c195c0eb0`) were removed and the new component layer `74e33464d5dae4246f077a821dd4c1514c26004e24f2b949ff4d1d95517f49e6` was added.

kate-goldenring: > I can try adding the GC labels to the layers themselves as well. This might avoid the collection but since there is a new image the pre-compile will be invoked again.

@jsturtevant this would be a good addition as in the Spin shim we are checking for precompilation. This could also be done in a subsequent PR as an optimization if you are hesitant to iterate again

jsturtevant: I've added this extra GC label as well as a couple tests that cover the scenario

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
jsturtevant: The layer config is being passed through to the runtime later in the function.  For consistency, having the correct digest of the compiled layer is important and not using original_layer.  

I guess runwasi could put that together and the api could just return `Result<Vec<Option<Vec,u8>>>>`. Is there  a strong reason for this from your perspective?

kate-goldenring: I mainly think we should avoid using `original_layer.config.digest()` and instead only use the returned `WasmLayer`s as the source of truth

kate-goldenring: I understand why we are using the original layer. In a future change, I think we may consider changing to `Result<Vec<Option<Vec<u8>>>>`, but i think this is fine.

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
kate-goldenring: That did the trick. Thank you!

kate-goldenring (approved):
Thank you @jsturtevant for all the iterations of this. This really speeds up the performance of the containerd shim being able to precompile all layers in one call and preserve those precompiled layers.

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
jsturtevant: The `original_layer.config.digest()` is being used for a label, I've made a few changes that make that clearer and also updated to the `Result<Vec<Option<Vec<u8>>>>` as I think the shim shouldn't be responsible for those changes here.

jsturtevant:
> Thank you @jsturtevant for all the iterations of this. This really speeds up the performance of the containerd shim being able to precompile all layers in one call and preserve those precompiled layers.

I pushed one last change based on your feedback.  Thanks for trying out all the changes, the feedback and being patient as I adjusted the API based on the usage

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -53,12 +54,14 @@ pub trait Engine: Clone + Send + Sync + 'static {
         &["application/vnd.bytecodealliance.wasm.component.layer.v0+wasm"]
     }
 
-    /// Precompiles a module that is in the WASM OCI layer format
-    /// This is used to precompile a module before it is run and will be called if can_precompile returns true.
+    /// Precompiles a passes supported OCI layers to engine for compilation
+    /// This is used to precompile the layers before they are run and will be called if can_precompile returns true.
```
kate-goldenring, line 58:
```suggestion
    /// Precompile passes supported OCI layers to engine for compilation
    /// This is used to precompile the layers before they are run and will be called if `can_precompile` returns `true`.
```

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -53,12 +54,14 @@ pub trait Engine: Clone + Send + Sync + 'static {
...
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
-    /// The cached, precompiled module will be reloaded on subsequent runs.
-    fn precompile(&self, _layers: &[Vec<u8>]) -> Result<Vec<u8>> {
-        bail!("precompilation not supported for this runtime")
+    /// The cached, precompiled layers will be reloaded on subsequent runs.
+    /// The runtime is expected to runtime the same number of layers passed in, if the layer cannot be precompiled it should return None for that layer.
```
kate-goldenring, line 61:
```suggestion
    /// The runtime is expected to return the same number of layers passed in, if the layer cannot be precompiled it should return `None` for that layer.
```

kate-goldenring (commented):
Two doc nits

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
jsturtevant: https://github.com/containerd/runwasi/issues/504 for tracking updates

radu-matei:
I may have bumped into an issue that will turn out a blocker for the current approach — while running a Wasm component built with SpiderMonkey, the resulting precompiled component size is  larger than the maximum message size in gRPC, which means saving it to the content store fails:

```
ERRO[2024-03-06T12:52:14.701165146+01:00] (*service).Write failed                       error="rpc error: code = ResourceExhausted desc = grpc: received message larger than max (43404572 vs. 16777216)" expected="sha256:617aa5799c43c511336f0c40708e8612dd804011eb9f713a9dbcb7340e8ff9c7" ref=precompile-runwasi.io/precompiled/spin/17767192358106208183 total=43404352
time="2024-03-06T11:52:14.703413103Z" level=warn msg="Error obtaining wasm layers for container sl.  Will attempt to use files inside container image. Error: response stream error: status: ResourceExhausted, message: "grpc: received message larger than max (43404572 vs. 16777216)", details: [], metadata: MetadataMap { headers: {} }"
```

Is there a way where `runwasi` could write the file in the store directly, then use the containerd API to add the correct annotations?

devigned:
Is the message not getting chunked? If not, perhaps that is the path we should pursue.

jsturtevant:
it looks like the write failed when saving the new content. Will need to implement streaming for larger content when doing the write. Right now it writes and commits in one action: https://github.com/containerd/runwasi/blob/2171b3d841752b9dc0ac5314479059d23e9f6df8/crates/containerd-shim-wasm/src/sandbox/containerd/client.rs#L198-L201

I don't think this is a blocker on this PR and I would be inclined to merge this and make those changes separately as this changeset already has quite a bit going on.

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -53,12 +54,14 @@ pub trait Engine: Clone + Send + Sync + 'static {
...
+    /// Precompile passes supported OCI layers to engine for compilation
+    /// This is used to precompile the layers before they are run and will be called if `can_precompile` returns `true`.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
-    /// The cached, precompiled module will be reloaded on subsequent runs.
-    fn precompile(&self, _layers: &[Vec<u8>]) -> Result<Vec<u8>> {
-        bail!("precompilation not supported for this runtime")
+    /// The cached, precompiled layers will be reloaded on subsequent runs.
+    /// The runtime is expected to return the same number of layers passed in, if the layer cannot be precompiled it should return `None` for that layer.
+    /// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
+    fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<Vec<u8>>>> {
```
Mossaka, line 63: nit: type alias for `Vec<Option<Vec<u8>>>>`, perhaps `PrecompiledWasmLayers`?

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -375,6 +347,15 @@ impl Client {
         })
     }
 
+    fn get_image_manifest(&self, image_name: &str) -> Result<(ImageManifest, String)> {
+        let image = self.get_image(image_name)?;
+        let image_digest = self.extract_image_content_sha(&image)?;
+        let manifest = self.read_content(&image_digest)?;
+        let manifest = manifest.as_slice();
+        let manifest = ImageManifest::from_reader(manifest)?;
```
Mossaka, line 355:
```suggestion
            let manifest = ImageManifest::from_reader(self.read_content(&image_digest)?.as_slice())?;

```

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -401,90 +378,140 @@ impl Client {
...
         };
 
-        log::info!("found manifest with WASM OCI image format.");
+        log::info!("found manifest with WASM OCI image format");
         // This label is unique across runtimes and version of the shim running
         // a precompiled component/module will not work across different runtimes or versions
         let (can_precompile, precompile_id) = match engine.can_precompile() {
             Some(precompile_id) => (true, precompile_label(T::name(), &precompile_id)),
             None => (false, "".to_string()),
         };
```
Mossaka, line 387:
```suggestion
       let precompile_id = engine.can_precompile().map(|id| precompile_label(T::name(), &id))
```

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -401,90 +378,140 @@ impl Client {
...
+        let image_info = self.get_info(&image_digest)?;
+        let mut needs_precompile =
+            can_precompile && !image_info.labels.contains_key(&precompile_id);
         let layers = manifest
             .layers()
             .iter()
             .filter(|x| is_wasm_layer(x.media_type(), T::supported_layers_types()))
-            .map(|config| self.read_content(config.digest()))
+            .map(|original_config| {
+                let mut digest_to_load = original_config.digest().clone();
```
Mossaka, line 397: Refactor into it's own function?

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -401,90 +378,140 @@ impl Client {
...
+                if can_precompile {
+                    let info = self.get_info(&digest_to_load)?;
+                    if info.labels.contains_key(&precompile_id) {
+                        // Safe to unwrap here since we already checked for the label's existence
+                        digest_to_load = info.labels.get(&precompile_id).unwrap().clone();
```
Mossaka, line 402:
```suggestion
                    if let Some(precompiled_digest) = info.labels.get(&precompile_id) {
                        ...
                        }
```

radu-matei:
Chunking sounds great, and I agree that we can do this in a follow-up.

Thanks for the quick update!

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -401,90 +378,140 @@ impl Client {
...
-            return Ok((vec![], platform));
+        if needs_precompile {
+            log::info!("precompiling layers for image: {}", container.image);
+            match engine.precompile(&layers) {
+                Ok(compiled_layers) => {
+                    if compiled_layers.len() != layers.len() {
+                        return Err(ShimError::FailedPrecondition(
+                            "precompile returned wrong number of layers".to_string(),
+                        ));
+                    }
```
Mossaka, line 442: nit: to imporve readability, consider

```rust
let compiled_layers = match engine.precompile(&layers) {
        Ok(compiled_layers) => {
                   if compiled_layers.len() != layers.len() {
                        return Err(ShimError::FailedPrecondition(
                            "precompile returned wrong number of layers".to_string(),
                        ));
                    }
                    compiled_layers
        }
        Err(e) => {
            log::error!("precompilation failed: {}", e);
            return Err(ShimError::FailedPrecondition("precompilation failed".to_string()));
        }
    };

    
```

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -401,90 +378,140 @@ impl Client {
...
-            self.update_info(image_content)?;
-
-            return Ok((
-                vec![WasmLayer {
-                    config: image_config_descriptor.clone(),
-                    layer: precompiled,
-                }],
-                platform,
-            ));
+        if layers.is_empty() {
```
Mossaka, line 509: nit: move this condition check to right after `layers` is decalred.

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -542,4 +580,370 @@ mod tests {
             .read_content(expected)
             .expect_err("content should not exist");
     }
+
+    #[test]
+    fn test_layers_when_precompile_not_supported() {
```
Mossaka, line 585: Thanks for adding so much test cases. Love it 😍

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -542,4 +580,370 @@ mod tests {
...
+        engine.add_precompiled_bits(fake_bytes2.bytes.clone(), &fake_precompiled_bytes2);
+
+        let (layers, _) = client.load_modules(container_name2, &engine).unwrap();
+        assert_eq!(engine.precompile_called.load(Ordering::SeqCst), 2);
+        assert_eq!(layers.len(), 2);
+        assert_eq!(engine.layers_compiled_per_call.load(Ordering::SeqCst), 1);
+    }
+
+    #[test]
+    fn test_layers_are_precompiled_for_multiple_layers() {
```
Mossaka, line 776: Maybe add a case where layers are empty, or all layers contain no Wasm type?

crates/containerd-shim-wasm/src/container/engine.rs
kate-goldenring: One nit to counter the nit with: not everything returned is a precompiled layer, which is why there is an option wrapping, so it feels like a confusing name.

Mossaka:
Okay, going to merge this and figure out the comments later.

crates/containerd-shim-wasm/src/container/engine.rs
Mossaka: Or maybe `Vec<Option<WasmLayer>>`?

kate-goldenring: `WasmLayer` is already a structure. The API was that at one point but was changed to just return the layer content rather than the content and config because the original config is used as the source of truth. I think we should maybe leave it as is or `PrecompiledLayer` could work, though i prefer not using types to alias something as small as `Vec<u8>`

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
Mossaka: realized this doesn't buy us much

PR #498, opened 2024-03-01
Add a dry run option to the release workflow

devigned:
This PR adds a dry run option to the release workflow and enables workflow_dispatch and nightly cron scheduling to ensure the release workflow is consistently in a working state. 

The cron and workflow_dispatch triggers will trigger the workflow to execute in dry run mode. In this mode, the workflow will not push artifacts or crates. By default it will select the most recent tag for its ref to build.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,9 +8,24 @@ on:
   push:
     tags:
       - '**/v[0-9]+.[0-9]+.*'
+  schedule:
+    - cron: '0 0 * * *' # run every night at midnight
+  workflow_dispatch:
+    inputs:
+      dryRun:
+        description: "run the workflow without publishing any artifacts"
+        required: true
+        default: true
+        type: boolean
+      crate:
+        description: "the crate to run the dry run release"
+        default: containerd-shim-wasmtime
+        required: true
+        type: string
 
 env:
   CARGO_TERM_COLOR: always
+  DRY_RUN: ${{ inputs.dryRun ||  github.event.schedule }}
 
 jobs:
   parse:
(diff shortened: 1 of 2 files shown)


devigned:
The dry run for the release workflow passes https://github.com/containerd/runwasi/actions/runs/8124152176. Via the cron scheduling we should get early warning to when failures are introduced. 

I intend on following this release workflow with additional release flow changes that will automate the cargo versioning.

.github/workflows/release.yml
```diff
@@ -21,13 +36,29 @@ jobs:
...
+        with:
+          ref: ${{ steps.ref.outputs.ref }}
       - id: parse
         name: Parse ref
         shell: bash
-        run: ./scripts/parse_ref.sh ${{ github.ref }} >> ${GITHUB_OUTPUT}
+        run: |
+          if [[ "${DRY_RUN}" == 'true' ]]
+          then
+            CRATE="${{ inputs.crate || 'containerd-shim-wasmtime' }}"
```
Mossaka, line 55: Why do we just dry-run the wasmtime shim?

.github/workflows/release.yml
```diff
@@ -81,19 +112,21 @@ jobs:
           if stat dist/bin/* >/dev/null 2>&1; then
             cosign sign-blob --yes \
               --bundle containerd-shim-${{ needs.parse.outputs.runtime }}-v1.bundle \
+              --output-signature dist/bin/containerd-shim-${{ needs.parse.outputs.runtime }}-v1.sig \
```
Mossaka, line 115: why are these added? I thought the `bundle` contains both the signature and the cert

Mossaka (commented):
Thanks for adding the dry run option!!

.github/workflows/release.yml
devigned: The release was failing as it was trying to `cp *.sig ./dist/bin`. I wasn't clear what the expectation was, so to achieve what I interpreted to be the original intent, I added the .sig and .pem output args.

.github/workflows/release.yml
devigned: Just simplicity. Could make it random. WDYT?

Mossaka: I think we should dry run the core crate, and one of the runtime shim implementation.

.github/workflows/release.yml
Mossaka: Can you show me the failed release job?

devigned: https://github.com/containerd/runwasi/actions/runs/8118178887/job/22191957368

Mossaka: Oh I believe the failure for this one is due to that the `containerd-shim-wasm-test-modules` is a crate (not a binary) and somehow it passes the condition described here: https://github.com/containerd/runwasi/blob/main/.github/workflows/release.yml#L77

and went to assume that it contains binaries for signing.

Mossaka: What we really need to do is to skip signing the containerd-shim-wasm-test-modules and containerd-shim-wasm crates

Mossaka: Need a rebase

jsturtevant:
anything holding this up besides a rebase?  this would have been handy since our latest release failed :-( https://github.com/containerd/runwasi/actions/runs/8351251493

devigned:
> anything holding this up besides a rebase? this would have been handy since our latest release failed :-( https://github.com/containerd/runwasi/actions/runs/8351251493

I was having something of an existential crisis about this PR and if it would actually help us. Can you drop the link for the failed release? I'll make sure it is something the job would catch and prep the PR for merge.

Issue #476, opened 2024-02-07
Why is the shim running on it's own network ns?


Mossaka:
Ref: https://github.com/containerd/runwasi/blob/71f8df9cc576ea9564b3ea692432e80b454da7e5/crates/containerd-shim-wasm/src/sandbox/shim/cli.rs#L56-L57

It is not clear to me the reason why the shim start function sets up it's own network ns. I checked out both the Go and Rust implementation of runc-shim, they aren't setting the same ns at shim start. 

@jsturtevant pointed me to this https://github.com/containers/youki/issues/2473 from youki but it looks like youki was setting up all the namespaces for containers but somehow the network ns was ignored. I am raising the issue here for asking more investigation. 

FYI @jprendes

cpuguy83:
Before youki we did everything ourselves.
It may not be necessary to do this anymore.

jprendes:
From https://github.com/containerd/runwasi/pull/327#issuecomment-1730071090, removing that `setup_namespaces` call means that the network ns is not set up at all.
At the time I thought that maybe youki wasn't setting up the network ns, but it does.
It would be great if anyone has further insight.

jsturtevant:
We removed it awhile back and broke everything 🤦  https://github.com/containerd/runwasi/pull/364

cpuguy83:
> youki is only setting the Pid and User namespaces

That doesn't seem right :(

0xE282B0:
I removed it once and @jsturtevant needed to add it again [here](
https://github.com/containerd/runwasi/pull/364/files#diff-8f8704e2bb5644be97cfbd99c736cb7f432da2de20292de2b0f69e4193bd0f64R1214-R1216)

jprendes:
> > youki is only setting the Pid and User namespaces
> 
> That doesn't seem right :(

That was me misreading the code. Youki sets Pid and User ns at one point, and the rest on a different place.
IIRC, Pid and User ns need to be set up in the intermediate process, while the rest are set up in the init process.

Mossaka:
> Youki sets Pid and User ns at one point, and the rest on a different place.

Okay, if this is true, then the shim doesn't need to set up its own network ns. There must be some root causes that we haven't yet discovered.

jprendes:
I've added some logging to runwasi and youki when setting namespaces, and got:
```
level=info msg=">>>>> runwasi > unshare(CloneFlags(CLONE_NEWNET))"
level=info msg="server listen started"
[...]

PR #503, opened 2024-03-05
feat(*): remove the network ns in the shim start

Mossaka:
this commit removes the unnecessary nerwork namepsace setup in the shim start function because it is already been set in youki.

See #476


jsturtevant:
The previous run failed, Did something change?

Mossaka:
> The previous run failed, Did something change?

Mossaka:
1 / 5 runs the CI failed

Mossaka:
timeout error: https://github.com/containerd/runwasi/actions/runs/11864777519/job/33069084169?pr=503#step:7:1001

jprendes:
That timeout error is linked to installing a debian package, so it can be ignored.

jprendes:
@Mossaka , do you ha e a feel from all the runs you did if there are still failures linked to this? If so, what kind of failures? I'm inclined to think that any failures are just the sporadic timeouts we've seen in other places in CI as well.

jprendes (approved):
LGTM. 
I'm happy with merging this. We csn revert if we notice any issues afterwards.

Mossaka:
> @Mossaka , do you ha e a feel from all the runs you did if there are still failures linked to this? If so, what kind of failures? I'm inclined to think that any failures are just the sporadic timeouts we've seen in other places in CI as well. 

I believe all the failures are related to the sporadic timeouts

PR #506, opened 2024-03-06
refactor(precompile): make code a bit more readable

Mossaka:
this is to resolve my comments made in PR #492 in an effort to make the code a bit more idiomatic and readable

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
@@ -369,12 +369,10 @@ impl Client {
         })
     }
 
-    fn get_image_manifest(&self, image_name: &str) -> Result<(ImageManifest, String)> {
+    fn get_image_manifest_and_digest(&self, image_name: &str) -> Result<(ImageManifest, String)> {
         let image = self.get_image(image_name)?;
         let image_digest = self.extract_image_content_sha(&image)?;
-        let manifest = self.read_content(&image_digest)?;
-        let manifest = manifest.as_slice();
-        let manifest = ImageManifest::from_reader(manifest)?;
+        let manifest = ImageManifest::from_reader(self.read_content(&image_digest)?.as_slice())?;
         Ok((manifest, image_digest))
     }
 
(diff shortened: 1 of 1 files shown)


crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -60,7 +60,7 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// The cached, precompiled layers will be reloaded on subsequent runs.
     /// The runtime is expected to return the same number of layers passed in, if the layer cannot be precompiled it should return `None` for that layer.
     /// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
-    fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<Vec<u8>>>> {
+    fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<WasmLayer>>> {
```
kate-goldenring, line 63: @jsturtevant and I had a discussion around why `Vec<u8>` is preferable here. For reference: https://github.com/containerd/runwasi/pull/492#discussion_r1506609277

kate-goldenring: That being said if this API makes sense to both of you, thats fine with me

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -394,126 +392,140 @@ impl Client {
...
+            let mut layers_for_runtime = Vec::with_capacity(compiled_layers.len());
+            for (i, compiled_layer) in compiled_layers.iter().enumerate() {
+                if compiled_layer.is_none() {
+                    log::debug!("no compiled layer using original");
+                    layers_for_runtime.push(layers[i].clone());
+                    continue;
+                }
+
+                let compiled_layer = compiled_layer.as_ref().unwrap();
+                let original_config = &compiled_layer.config;
```
kate-goldenring, line 435: Seems weird to consider the config returned from `precompile` be "original". This is why i think keeping the API to `Option<Vec<u8>>` may make sense. Otherwise, this may want to use the `compiled_layers` above or add a comment to the API saying that it expects the returned config to be the original one.

crates/containerd-shim-wasm/src/container/engine.rs
Mossaka: ahh thanks for the context. my bad that I didn't read the thread carefully.

Mossaka: Just reverted back the return type of `precompile` to `Result<Vec<Option<Vec<u8>>>>`

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
Mossaka: Reverted back

devigned (dismissed):
+1 lgtm!

Mossaka:
I just squashed my commits and invalidated the review. Could you please approve it again? @devigned

devigned (approved):

PR #520, opened 2024-03-19
fix: use latest containerd crate with patched cgroups parsing

kate-goldenring:


Cargo.toml
```diff
@@ -22,7 +22,7 @@ homepage = "https://github.com/containerd/runwasi"
 anyhow = "1.0"
 cap-std = "1.0"
 chrono = { version = "0.4", default-features = false, features = ["clock"] }
-containerd-shim = "0.7.0"
+containerd-shim = { git = "https://github.com/containerd/rust-extensions", rev = "6b9e5b28be2870731a18ca83c13e63a98f35e149" }
```
devigned, line 25: I believe this change will cause the `containers-shim-wasm` crate to not be able publish due to the git dependency. Is this intended to be temporary, or would it be possible to get a rust-extensions release with the fix.

kate-goldenring: I was not aware that the release CI has that as a criteria. Is that a clean code criteria?

devigned: I don't think so much a clean code criteria, but rather `cargo publish` doesn't allow for crates to be published with git references. If we attempt to release the `containerd-shim-wasm` crate which takes a workspace version dependency on `containerd-shim`, I'd expect `cargo publish` to fail during the release workflow.

We can pull this in, but we just won't be able to publish `containers-shim-wasm` until `rust-extensions` is published and we update the version to point at the new version of the `rust-extensions` crate.

kate-goldenring: If for some reason in the near future, `runwasi` has changes that need to be released and `rust-extensions` does not want to release, we can revert this to unblock `runwasi`

devigned: @jsturtevant do you have a sense when `rust-extensions` is going to have another release? Also, what do you think about pulling this in and reverting if we need to release prior to `rust-extensions`?

I'm +1 for pulling in the change to unblock folks.

jsturtevant: I will work on a release for rust-extensions this afternoon, usually just a day or two to merge the version bump but we can merge this to unblock folks using it.

PR #523, opened 2024-03-20
Enable building and release of oci-tar-builder

jsturtevant:
The [0.5.0 release](https://github.com/containerd/runwasi/actions/runs/8351251493) is failing due to missing `  no matching package named `oci-tar-builder` found`.  After attempting to release this package the [release build](https://github.com/containerd/runwasi/actions/runs/8362454303/job/22893107967) failed with:

```
error: package ID specification `containerd-shim-oci-tar-builder` did not match any packages
```

This make the updates to the makefile to support building and publishing this package.  At a later time, we can build the `oci-tar-builder` binary and create a release artifact but for now we can release the library as is.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,7 +74,7 @@ jobs:
         run: make test-${{ needs.parse.outputs.runtime }}
 
       - name: Sign the binary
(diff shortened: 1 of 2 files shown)


devigned (approved):
lgtm, kind of

We (I) need to do some work on this workflow.

PR #525, opened 2024-03-20
Add Description and license to oci-tar-builder cargo.toml

jsturtevant:
[Cargo publish](https://github.com/containerd/runwasi/actions/runs/8364790324/job/22901153910) requires these fields but interestingly `publish --dry-run` doesn't tell you this when testing with it.

https://doc.rust-lang.org/cargo/reference/publishing.html#before-publishing-a-new-crate


devigned (approved):
That seems wrong of `--dry-run`. lgtm

PR #532, opened 2024-03-25
change release flow to be workflow_dispatch rather than tag based

devigned:
This PR is intended to subsume #498.

With this PR, the release process now triggers from a `workflow_dispatch` event where the input is a `crate`, a `version`, and a flag for `dry_run`. The release process will do the following:
1. Run a pre-release job
    1. extract runtime name from crate name
    2. determine if the crate is a shim
    3. verify if the owner for the crate matches with the runwasi-committers and adds them if not (this could have failed in the past if the group was already in the list)
    4. replaces the version of the crate inline and commits the change to main
    5. adds dry run if statements on parts of the job that talk to external data stores
2. build and sign
    1. does the same thing as before, but includes the outputs from the previous job as matrix variables for shorter addressing in the job.
    2. updates the signing to alway produce key and pem files for shims
    3. adds dry run if statements on parts of the job that talk to external data stores
4. release
    1. uses a single matrix build to include outputs from the pre-release job as matrix variables.
    2. adds dry run if statements on parts of the job that talk to external data stores
    
### To test
```bash
gh workflow run release.yml -f dry_run=true -f crate=containerd-shim-wasm -f version=0.4.0 --ref refactor-release
```

### Example runs
- [containerd-shim-wasmtime](https://github.com/containerd/runwasi/actions/runs/8473560851)
- [containerd-shim-wasm](https://github.com/containerd/runwasi/actions/runs/8473562187)


.github/workflows/release.yml
```diff
@@ -1,158 +1,206 @@
...
       - name: Cargo publish
-        run: cargo publish --package ${{ needs.parse.outputs.crate }} --verbose --locked
-        env:
-          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_PUBLISH_TOKEN }}
-      - name: Check crates.io ownership
-        run: |
-          cargo owner --add github:containerd:runwasi-committers ${{ needs.parse.outputs.crate }}
-          cargo owner --list ${{ needs.parse.outputs.crate }} | grep github:containerd:runwasi-committers
+        if: ${{ matrix.is_shim  != 'true' }}
+        run: cargo publish ${{ inputs.dry_run && '--dry-run' || '' }} --package ${{ matrix.crate }} --verbose --locked
```
jsturtevant, line 204: I think the step order here should be:

1. publish
2. tag repo
3. create release

if publishing fails we shouldn't be creating a release.  Also we can then add something to the release notes of the crate like 

```
Add the crate to you project:

cargo add containerd-shim-wasm@0.5.0
```

.github/workflows/release.yml
```diff
@@ -1,158 +1,206 @@
...
+        run: |
+          cargo owner --list ${{ inputs.crate }} | grep github:containerd:runwasi-committers || \
+            cargo owner --add github:containerd:runwasi-committers ${{ inputs.crate }}
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_PUBLISH_TOKEN }}
+      - name: Update crate version and commit change to main
+        if: ${{ !inputs.dry_run }}
+        run: |
+          # replace the version inline in the Cargo.toml
+          sed -i '' -E 's/^version.+=.+".+"/version = "${{ inputs.version }}"/' crates/${{ inputs.crate }}/Cargo.toml
```
jsturtevant, line 64: Interested in what others think about this. I am afraid of typing the wrong thing into the input box in the GH UI and kicking off the wrong build here.  I think I'd prefer to do a PR to update the version then grab the latest version

jsturtevant (commented):
We should also update https://github.com/containerd/runwasi/blob/main/RELEASE.md with the new steps

jsturtevant:
> ### To test
> ```shell
> gh workflow run release.yml -f dry_run=true -f crate=containerd-shim-wasm -f version=0.4.0 --ref refactor-release
> ```

love this!

.github/workflows/release.yml
```diff
@@ -1,158 +1,206 @@
...
+      version: ${{ inputs.version }}
+      ### is_shim is a string, not a boolean, so use: is_shim == 'true'
+      is_shim: ${{ steps.runtime_sub.outputs.is_shim }}
     steps:
-      - uses: actions/checkout@v3
+      # - name: Fail if branch is not main
+      #   if: github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main'
+      #   run: |
+      #     echo "This workflow should not be triggered with workflow_dispatch on a branch other than main"
+      #     exit 1
```
jsturtevant, line 39: does this need to be re-enabled?

devigned: Yes! This was intentional until we are ready to merge for testing purposes.

.github/workflows/release.yml
```diff
@@ -1,158 +1,206 @@
...
         with:
-          ref: ${{ github.ref }}
-      - id: parse
-        name: Parse ref
-        shell: bash
-        run: ./scripts/parse_ref.sh ${{ github.ref }} >> ${GITHUB_OUTPUT}
+          script: |
+            const crate = '${{ inputs.crate }}';
+            let runtime = crate.replace(/^containerd-shim-/, '');
+            core.setOutput('runtime', runtime);
```
jsturtevant, line 49: runtime seems a little mis-leading here since some of the crates aren't runtimes.  

 Another way to do this might be flip this logic to be hardcode the `shim_crates` instead of the `non_shim_crates` and set runtime if its included?

devigned: `runtime` is a carry over from previous nomenclature. I couldn't think of much else other than runtime since it is mostly used in that way.

I'm indifferent about `shim_crates` vs. `non_shim_crates`. Do you feel like one reads easier than another? I fear my feeling for it has dulled from being too close to it.

.github/workflows/release.yml
devigned: @Mossaka what say you?

devigned: I can add some sanitation to make sure the build fails based on checking the crate names and the version regex. That could provide some basic sanity checks. However, it is not going to be a second set of human eyes, which may be the root of what you are seeking.

jsturtevant: >  However, it is not going to be a second set of human eyes, which may be the root of what you are seeking.

Yes but its probably not a huge deal

.github/workflows/release.yml
jsturtevant: how about `project`?  I see how it is being used below, in some cases its the `runtime` others its the `crate`.  could be better to leave as is so we don't introduce another terminology 🤷

.github/workflows/release.yml
devigned: Check out: https://github.com/containerd/runwasi/actions/runs/8483472409/job/23244657527.

I've added some sanity checks to the early part of the release.

.github/workflows/release.yml
```diff
@@ -1,158 +1,222 @@
...
+      - uses: actions/checkout@v4
+      ### Determine the name of the runtime and if it is a binary release or crates.io
+      - name: verify version input
+        uses: actions/github-script@v7
         with:
-          ref: ${{ github.ref }}
-      - id: parse
-        name: Parse ref
-        shell: bash
-        run: ./scripts/parse_ref.sh ${{ github.ref }} >> ${GITHUB_OUTPUT}
```
jsturtevant, line 30: I think we can drop this file now

.github/workflows/release.yml
```diff
@@ -1,158 +1,222 @@
...
-        run: cargo publish --package ${{ needs.parse.outputs.crate }} --verbose --locked
-        env:
-          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_PUBLISH_TOKEN }}
-      - name: Check crates.io ownership
-        run: |
-          cargo owner --add github:containerd:runwasi-committers ${{ needs.parse.outputs.crate }}
-          cargo owner --list ${{ needs.parse.outputs.crate }} | grep github:containerd:runwasi-committers
-        env:
-          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_PUBLISH_TOKEN }}
+          RELEASE_NAME: ${{ matrix.crate }}/${{ matrix.version }}
```
jsturtevant, line 222: this needs the v infront of version or the version input should be updated to check for version?

RELEASE.md
```diff
@@ -4,16 +4,25 @@ This document describes the steps to release a new version of the crate.
...
+- Verifies ability to push crates
+- Updates the version of the crate to the version specified in the workflow input
 - Build the crate to be released (determined by the tag), including any artifacts (e.g., associated binaries)
 - Run the tests for that crate (and only that crate!)
+- Publishes to the crates.io
+- Tags the repository for the release
 - Creates a GitHub release for that crate (attaching any artifacts)
-- Publish the crate to crates.io
 
 ### Crate Release Sequence
```
jsturtevant, line 27: I missed this part when I was doing it the first time should be something like 
```suggestion
### Crate Release Sequence

Must release the creates in this order due to depenedencies: 
```

and add `oci-tar-builder` below `containerd-shim-test-modules` but before `containerd-shim-wasm`

devigned:
@jsturtevant thank you for your keen review. Nice work catching the RELEASE_NAME `v` missing.

jsturtevant:
LGTM, I tried to run it one last time and got:

```
gh workflow run release.yml -f dry_run=true -f crate=containerd-shim-wasm -f version=0.4.1 --ref refactor-release
could not create workflow dispatch event: HTTP 403: Must have admin rights to Repository.
```

jsturtevant:
we can remove https://github.com/containerd/runwasi/pull/532#pullrequestreview-1967284020 and we can try it out 🚀

.github/workflows/release.yml
Mossaka: I tried to run this command on my ubuntu devbox but it gave an error 

```
sed: can't read s/^version.+=.+".+"/version = "0.5.1"/: No such file or directory
```

it seems like in Ubuntu, the command expects no space between `-i` and the backup extension `""`, so the correct command should be 

```bash
sed -i'' -E ...
```

.github/workflows/release.yml
```diff
@@ -1,158 +1,222 @@
...
+        if: ${{ matrix.is_shim == 'true' }}
         run: |
-          make dist-${{ needs.parse.outputs.runtime }}
+          make dist-${{ matrix.runtime }}
           # Check if there's any files to archive as tar fails otherwise
           if stat dist/bin/* >/dev/null 2>&1; then
             cosign sign-blob --yes \
-              --bundle containerd-shim-${{ needs.parse.outputs.runtime }}-v1.bundle \
-              dist/bin/containerd-shim-${{ needs.parse.outputs.runtime }}-v1
+              --output-signature containerd-shim-${{ matrix.runtime }}-v1.sig \
```
Mossaka, line 133: I don't think the `output-signature` and `output-certifcate` are necessary.

.github/workflows/release.yml
```diff
@@ -1,158 +1,222 @@
...
+        uses: actions/github-script@v7
         with:
-          ref: ${{ github.ref }}
-      - id: parse
-        name: Parse ref
-        shell: bash
-        run: ./scripts/parse_ref.sh ${{ github.ref }} >> ${GITHUB_OUTPUT}
+          script: |
+            const version = '${{ inputs.version }}';
+            if(!version.match(/^[0-9]+.[0-9]+.*/)) {
```
Mossaka, line 48: I don't think this sanity check account for the full semantic versioning check. For example, "1.0" seems to be able to pass this check, no? 

The suggested Regex for semantic versioning check is: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

Mossaka (commented):
Added some comments

RELEASE.md
```diff
@@ -36,7 +48,7 @@ containerd-shim-wasm = { path = "crates/containerd-shim-wasm", version = "0.4.0"
 
 1. Open a PR to bump crate versions and dependency versions in `Cargo.toml` for that crate
```
Mossaka, line 49: No longer needed?

RELEASE.md
```diff
@@ -36,7 +48,7 @@ containerd-shim-wasm = { path = "crates/containerd-shim-wasm", version = "0.4.0"
...
 2. PR can be merged after 2 LGTMs
```
Mossaka, line 50: No longer needed?

RELEASE.md
```diff
@@ -4,21 +4,33 @@ This document describes the steps to release a new version of the crate.
...
 - Run the tests for that crate (and only that crate!)
+- Publishes to the crates.io
+- Tags the repository for the release
 - Creates a GitHub release for that crate (attaching any artifacts)
-- Publish the crate to crates.io
 
 ### Crate Release Sequence
+
+Must release the creates in this order due to dependencies:
 1. `containerd-shim-wasm-test-modules`
```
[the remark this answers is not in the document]
jsturtevant: I had the same expectation, I tried it and got error messages like 

```
Caused by:
  failed to select a version for the requirement `containerd-shim-wasm-test-modules = "^0.4.0"`
  candidate versions found which didn't match: 0.3.1, 0.3.0
```
```
no matching package namedoci-tar-builder found
```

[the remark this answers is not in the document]
jsturtevant: https://github.com/containerd/runwasi/actions/runs/8351251493/job/22859463521 (the first run ran into this)

.github/workflows/release.yml
jsturtevant: this is really needed any more we ended up matching it to the crate file: https://github.com/containerd/runwasi/blob/fc7aa53b0f58f1945afdc46bc29ceb7b5bd307d5/.github/workflows/release.yml#L79.  We could remove it or add it to a PR check with suggested regular expression

Since the PR needs to bump the version, maybe a check with something like https://github.com/obi1kenobi/cargo-semver-checks or that regex would be helpful

RELEASE.md
jsturtevant: after several iterations see https://github.com/containerd/runwasi/pull/539#pullrequestreview-1971615202

PR #533, opened 2024-03-29
Fix for running sed on Linux for release pipeline

jsturtevant:
Release failed because sed is different on Linux/macos 🤕 

Also adds https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#run-name so can see what ran with out having to open the action logs

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,5 @@
 name: release
+run-name: Release ${{ inputs.crate }} with v@${{ inputs.version }} (DryRun - ${{ inputs.dry_run }})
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,4 +1,5 @@
 name: release
+run-name: Release ${{ inputs.crate }} with v@${{ inputs.version }} (DryRun - ${{ inputs.dry_run }})
```
devigned, line 2: 🙌

devigned (approved):
lgtm!

PR #536, opened 2024-03-29
Remove auto-commit and verify version

jsturtevant:
The main branch is protected so we cannot do commits to it


devigned (approved):
lgtm!

Mossaka:
With this change in, we need to modify the RELEASE.md to remove the step `- Updates the version of the crate to the version specified in the workflow input`

PR #539, opened 2024-03-30
docs: delete outdated release steps in RELEASE.md

Mossaka:
this commit deleted the first two steps in the release process which are not necessary anymore since we changed tag-based release to workload-dispatch release


jsturtevant (changes requested):
leave steps as is, PR's are still required to bump crate versions

PR #546, opened 2024-04-03
ci(*): merge build action

Mossaka:
this commit merges the build step in the release action and the one in the ci action since they share a large overlap code.

the build step not only builds the binary, but signs it, and upload as action artifacts. this step is now running for every ci run, including PR and release pipeline.

### DIFF
--- a/.github/workflows/action-build.yml
+++ b/.github/workflows/action-build.yml
@@ -1,3 +1,4 @@
+# this action will require write permission to id-token
 name: Build shims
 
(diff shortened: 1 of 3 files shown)


Mossaka:
I realized whenever we want to sensitive operations with permissions like `write` on the Github actions, it is always required to create the PR from branches in the upstream repo, not from forked repos. Hence I am closing this PR as it went no where and created a new one #548

PR #547, opened 2024-04-03
ci(*): publish oci-tar-builder binary in the ci

### LINKED ISSUE #524: Create Release binary for oci-tar-builder
As follow up to https://github.com/containerd/runwasi/pull/523, we could build the executable and publish as part of the release: https://github.com/containerd/runwasi/blob/main/crates/oci-tar-builder/README.md#executable-usage

Mossaka:
closes #524 

I didn't like the fact that the published artifacts has prefixed name `containerd-shim-` which doesn't really fit the purpose of the `oci-tar-builder`.

The `oci-tar-builder` binary is built for each Linux, Windows and Macos-x86_64 and Macos-arm64


jsturtevant:
I was thinking https://github.com/containerd/runwasi/issues/524 would be part of the release CI as well.  Would we want to sign it too?

Mossaka:
> I was thinking #524 would be part of the release CI as well.

Ah yea I intend to work on another PR to add the release of the bianry as part of the release ci. 

>  Would we want to sign it too?

sure :)

jsturtevant:
> I didn't like the fact that the published artifacts has prefixed name containerd-shim- which doesn't really fit the purpose of the oci-tar-builder.

I think I have a preference for modify the `build` action to handle this verses having a separate file and process. This would give the same steps for the signing too

.github/workflows/action-build-oci-tar-builder.yml
```diff
@@ -0,0 +1,41 @@
...
+          # Check if there's any files to archive as tar fails otherwise
+          if stat dist/bin/* >/dev/null 2>&1; then
+            tar -czf dist/oci-tar-builder-${{ inputs.os }}.tar.gz -C dist/bin .
+          else
+            tar -czf dist/oci-tar-builder-${{ inputs.os }}.tar.gz -T /dev/null
+          fi
+      - name: Upload artifacts
+        uses: actions/upload-artifact@master
+        with:
+          name: oci-tar-builder-${{ inputs.os }}
```
jprendes, line 40: This would publish `oci-tar-builder-macos-12` and `oci-tar-builder-macos-14`
Could we have
* `macos-12` -> `macos-x86_64`
* `macos-14` -> `macos-aarch64`
* `ubuntu-22.04` -> `linux-glibc`
* `windows-latest` -> `windows`

or some similar mapping?

I think that would make finding the artifact to download wasier for our users.

jprendes: ah, wait, this is the CI artifact, not the published release, never mind then :-)

jprendes:
> > I didn't like the fact that the published artifacts has prefixed name containerd-shim- which doesn't really fit the purpose of the oci-tar-builder.
> 
> I think I have a preference for modify the build action to handle this verses having a separate file and process. This would give the same steps for the signing too

I agree with both of these comments.
I would like the build action to handle both cases. We could either:
* have a flag in the build action to conditionally add the `containerd-shim-` prefix, or
* just drop the prefix, and release unprefixed artifacts for the shims as well.

I prefer the first option. I would even defer that logic to the makefile, which already has similar logic.

jprendes (dismissed):
LGTM

.github/workflows/action-build.yml
```diff
@@ -73,13 +81,13 @@ jobs:
           make dist-${{ inputs.runtime }}
           # Check if there's any files to archive as tar fails otherwise
           if stat dist/bin/* >/dev/null 2>&1; then
-            tar -czf dist/containerd-shim-${{ inputs.runtime }}-${{ inputs.slug }}.tar.gz -C dist/bin .
+            tar -czf dist/$PREFIX{{ inputs.runtime }}-${{ inputs.slug }}.tar.gz -C dist/bin .
```
jprendes, line 84:
```suggestion
            tar -czf dist/$PREFIX${{ inputs.runtime }}-${{ inputs.slug }}.tar.gz -C dist/bin .
```

jprendes (approved):
LGTM!

I would have a slight preference towards the musl version instead of gnu, but it's not a big deal.

Also, should we include this in release.yml as well?

Mossaka:
> Also, should we include this in release.yml as well?
I am hoping the release.yml should JUST WORK https://github.com/containerd/runwasi/blob/main/.github/workflows/release.yml#L19

PR #548, opened 2024-04-03
ci(*): merge build action

Mossaka:
this commit merges the build step in the release action and the one in the ci action since they share a large overlap code.

the build step not only builds the binary, but signs it, and upload as action artifacts. this step is now running for every ci run, including PR and release pipeline.

### DIFF
--- a/.github/workflows/action-build.yml
+++ b/.github/workflows/action-build.yml
@@ -1,3 +1,4 @@
+# this action will require write permission to id-token
 name: Build shims
 
 on:
@@ -22,7 +23,7 @@ on:
 jobs:
   build-sign-upload:
     permissions:
-      # cosign uses the GitHub OIDC token
+      contents: write
       id-token: write
     name: build for ${{ inputs.slug }}
     runs-on: ${{ inputs.os }}
@@ -60,7 +61,6 @@ jobs:
           make test-${{ inputs.runtime }}
         if: ${{ inputs.arch == 'x86_64' }}
       - name: Sign the binary
-        shell: bash
         if: ${{ inputs.runtime != 'common' && inputs.slug != 'windows' }}
         run: |
           make dist-${{ inputs.runtime }}
--- a/.github/workflows/action-build.yml
+++ b/.github/workflows/action-build.yml
@@ -1,3 +1,6 @@
+
+# yaml-language-server: $schema=https://json.schemastore.org/github-action.json
+
 # this action will require write permission to id-token
 name: Build shims
 
@@ -19,6 +22,9 @@ on:
       arch:
         required: false
         type: string
+      sign:
+        default: false
+        type: boolean
 
 jobs:
   build-sign-upload:
@@ -45,7 +51,7 @@ jobs:
         if: runner.os == 'Linux'
         run: ./scripts/setup-cross.sh ${{ inputs.target }}
       - name: Setup cosign for signing
-        if: ${{ inputs.runtime != 'common' }}
+        if: ${{ inputs.runtime != 'common' && inputs.sign }}
         uses: sigstore/cosign-installer@v3.3.0
         with:
           cosign-release: 'v2.2.2'
@@ -60,7 +66,7 @@ jobs:
           make test-${{ inputs.runtime }}
         if: ${{ inputs.arch == 'x86_64' }}
       - name: Sign the binary
-        if: ${{ inputs.runtime != 'common' && inputs.slug != 'windows' }}
+        if: ${{ inputs.runtime != 'common' && inputs.slug != 'windows' && inputs.sign }}
         run: |
           make dist-${{ inputs.runtime }}
           # Check if there's any files to archive as tar fails otherwise
--- a/.github/workflows/action-build.yml
+++ b/.github/workflows/action-build.yml
@@ -23,7 +23,6 @@ on:
 jobs:
   build-sign-upload:
     permissions:
-      contents: write
       id-token: write
     name: build for ${{ inputs.slug }}
     runs-on: ${{ inputs.os }}
--- a/.github/workflows/action-build.yml
+++ b/.github/workflows/action-build.yml
@@ -45,7 +45,7 @@ jobs:
         if: runner.os == 'Linux'
         run: ./scripts/setup-cross.sh ${{ inputs.target }}
       - name: Setup cosign for signing
-        if: ${{ matrix.is_shim == 'true' }}
+        if: ${{ inputs.runtime != 'common' }}
         uses: sigstore/cosign-installer@v3.3.0
         with:
           cosign-release: 'v2.2.2'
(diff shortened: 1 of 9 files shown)


jsturtevant:
Are there implications for signing the binary every PR?

.github/workflows/release.yml
```diff
@@ -85,92 +85,22 @@ jobs:
...
-      - name: Upload artifacts
-        if: ${{ matrix.is_shim == 'true' }}
-        uses: actions/upload-artifact@master
-        with:
-          name: containerd-shim-${{ matrix.runtime }}-${{ matrix.arch }}
-          path: dist/containerd-shim-${{ matrix.runtime }}-${{ matrix.arch }}.tar.gz
+    uses: ./.github/workflows/action-build.yml
+    with:
+      os: "ubuntu-22.04"
+      runtime: ${{ matrix.runtime }}
```
jsturtevant, line 100: there will be times when this value won't be `common` as the check in the action is expecting

Mossaka:
> Are there implications for signing the binary every PR?

Since we are signing the artifacts that we publish on every build, I can imagine increased security and potentially degraded performance and increased complexity of managing keys. 

Although stated in the release doc for how to verify the binaries, we probably want to make that even clearer by copying the exact verification command in the release doc everytime we release signed artifacts. 

I am not sure if anyone would want to verify build artifacts for each PR / commit, but if they want, it's do-able.

.github/workflows/release.yml
```diff
@@ -85,92 +85,22 @@ jobs:
 
   build-and-sign:
     permissions:
-      # cosign uses the GitHub OIDC token
       id-token: write
+      contents: write
```
jsturtevant, line 89: What was content write used for? I don't see anything that introduced that

Mossaka: That was added for testing purposes. Good catch! will remove that

.github/workflows/action-test-smoke.yml
```diff
@@ -22,13 +22,13 @@ jobs:
       - name: Download artifacts
         uses: actions/download-artifact@master
         with:
-          name: containerd-shim-${{ inputs.runtime }}-linux-musl
+          name: containerd-shim-${{ inputs.runtime }}-x86_64-linux-musl
```
jsturtevant, line 25: interesting, I don't see where this arch was being included in the `inputs.runtime` before.  Or is this extra field new?

.github/workflows/release.yml
Mossaka: Sorry, can you elaborate your concerns here?

.github/workflows/action-test-smoke.yml
Mossaka: Before this PR, only the release pipeline runs both `x86_64` and `aarch64`. The ci does not have a matrix for different archs. I added that to make the build job on par with the one in the release CI.

.github/workflows/release.yml
jsturtevant: This value in the release.yml will be something like `oci-tar-builder` https://github.com/containerd/runwasi/blob/8c98100eb74008844975aa48a6521f618f0eef50/.github/workflows/release.yml#L66-L69

and I think signing will fail in the new signing step at https://github.com/containerd/runwasi/pull/548/files#diff-43035769cde349b680ebf72f18e67b049e730d0d0ec91174e6455a685b3c2accR61-R64

marosset:
If you add `# yaml-language-server: $schema=https://json.schemastore.org/github-action.json` to the top of all the github action files you'll get schema validation in a lot of editors like vs code

.github/workflows/action-test-smoke.yml
```diff
@@ -22,13 +22,13 @@ jobs:
...
           path: dist
       - name: Unpack artifats
         shell: bash
         run: |
           mkdir -p dist/bin
-          tar -xzf dist/containerd-shim-${{ inputs.runtime }}-linux-musl.tar.gz -C dist/bin
+          tar -xzf dist/containerd-shim-${{ inputs.runtime }}-x86_64-linux-musl.tar.gz -C dist/bin
```
marosset, line 31: nit: Maybe make this a makefile target since the tar command is repeated in multiple workflows?

Mossaka:
> Are there implications for signing the binary every PR?

Now that I thought more about the implications, and realized that I missed one of the biggest implication in my previous comment - signing requires feature branches to be pushed to upstream, because they require `id-token: write` permission which will not grant to public forked repos (the max permission for `id-token` is `read` for forked repos, so even though you specified `write` permission, the pipeline won't grant you the context you want). 

I am okay to put a condition on the signing to differentiate build / release pipelines so we can disable signing on build. Wdyt?

.github/workflows/release.yml
Mossaka: Good point! I have modified the logic of `runtime_sub` step to replace the name of non-shim crate to `common`.

.github/workflows/action-build.yml
```diff
@@ -15,13 +19,24 @@ on:
...
+  build-sign-upload:
+    permissions:
+      id-token: write
     name: build for ${{ inputs.slug }}
     runs-on: ${{ inputs.os }}
     steps:
-      - uses: actions/checkout@v3
+      - name: describe runner
+        run: |
+          echo "Running job with os: '${{ inputs.os }}', arch: '${{ inputs.arch }}', slug: '${{ inputs.slug }}', runtime: '${{ inputs.runtime }}', target: '${{ inputs.target }}'"
```
cpuguy83, line 38: Maybe you can write this such that it shows up as an annotation? https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-setting-a-notice-message

.github/workflows/action-build.yml
```diff
@@ -35,12 +50,51 @@ jobs:
...
+        shell: bash
+        run: echo "OPT_PROFILE=release" >> ${GITHUB_ENV}
       - name: Build
         run: make build-${{ inputs.runtime }}
       - name: Run tests
         timeout-minutes: 5
         run: |
           make test-${{ inputs.runtime }}
+        if: ${{ inputs.arch == 'x86_64' }}
+      - name: Sign the binary
```
cpuguy83, line 68: Maybe we can move signing to a separate action that we call from here?
We can move the `Setup cosign for signing` part in there as well.

cpuguy83 (changes requested):

.github/workflows/action-build.yml
Mossaka: I am not sure if I understand the benefits of creating annotations. The example showed in the doc has a file name and a line number, which isn't applied to what we want to show here. 

Perhaps, a job summary would be better suited? GITHUB_STEP_SUMMARY

Mossaka: I've added `error` and `notice` annotations to logs.

.github/workflows/action-build.yml
Mossaka: Done

Mossaka:
Could you please this PR again? @cpuguy83 @devigned 🙏

devigned (approved):
lgtm!

Nice work, @Mossaka!

Mossaka:
Going to merge this now and any follow up can be addressed later

PR #549, opened 2024-04-05
ci(*): automate the version bump process

Mossaka:
this PR automates the version bump process to
ask the bot to create a PR to bump the respective
crate's version to the given one on behalf of humans.

the PR created by the bot contains magic message that
will be parsed by another workflow `publish` which triggers
on merge to the main branch.

The `publish` workflow parses the commit message to get
crate, version and dry-run information to be able to
publish the crate, release binary and push tags.

the motivation for this commit is to further simplify the
release process

builds on top of #548

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,7 +18,6 @@ jobs:
       version: ${{ steps.parse.outputs.version }}
       is_binary: ${{ steps.parse_crate.outputs.is_binary }}
       is_crate: ${{ steps.parse_crate.outputs.is_crate }}
-      #  e0ef1cf000bfb3f12e5b8613977ac9d901e39fbd...1ab3eb4318fb824cc44e36530b48e600fe18f258
     steps:
       - uses: actions/checkout@v4
         with:
@@ -51,9 +50,9 @@ jobs:
           echo "crate: $crate"
           echo "version: $version"
           
-          echo "dry_run:$dry_run" >> $GITHUB_OUTPUT
-          echo "crate:$crate" >> $GITHUB_OUTPUT
-          echo "version:$version" >> $GITHUB_OUTPUT
+          echo "dry_run=$dry_run" >> $GITHUB_OUTPUT
+          echo "crate=$crate" >> $GITHUB_OUTPUT
+          echo "version=$version" >> $GITHUB_OUTPUT
         
       - name: parse crate
         id: parse_crate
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -73,152 +58,48 @@ jobs:
...
+          ${{ inputs.crate }} v${{ inputs.version }} when merged. The commit
+          message for this PR has a marker that is detected by CI to create
+          tags and publish crate artifacts.
+
+          When first opened this PR will not have CI run because it is generated
+          by a bot. A maintainer should close this PR and then reopen it to
+          trigger CI to execute which will then enable merging this PR.
+          EOF
+      - name: Make a PR
+        run: gh pr create -B "$PR_BASE" -H "$PR_HEAD" --title "$PR_TITLE" --body "$(cat ./pr-body)"
```
cpuguy83, line 103: So I recalled the problem with using the github actions token for opening pull requests: Jobs will not trigger on that pull request.
So you can open the PR but CI will not run.

Mossaka: Ahh yes, but I think that's fine as long as we document it in the release.md something like "you need to close the PR and re-open the PR to trigger the CIs". How does this sound?

Mossaka: "When the PR opens, close it then reopen it. Don't ask questions."

devigned (dismissed):
lgtm

.github/workflows/release.yml
cpuguy83: If that works that's fine.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,106 @@
+
+name: Publish Artifacts
+
+on:
+  push:
+    branches: [main]
```
cpuguy83, line 6: This doesn't seem right.
Looks like its going to run this on every merge to main?

cpuguy83 (changes requested):

.github/workflows/publish.yml
Mossaka: Perhaps, a release branch would make more sense?

cpuguy83: It seems like the intent would be to run this on tag rather than on a branch.

Mossaka: The intention is to run the `publish` action after the version bump PR merged to `main` branch

Mossaka:
This PR may need some love, but let me first release a new version of runwasi

jsturtevant:
Going to close this up, but feel free to refresh at some point

PR #564, opened 2024-04-17
feat(tracing): add tracing instruments to containerd-shim-wasm

Mossaka:
this commit adds tracing instrument macros to functions in the containerd-shim-wasm crate to capture spans of each function including its parameters and results.

these spans can be in turn be collected using tracing-opentelemetry and opentelemetry SDK at the shim binary level and output to collectors like Jeager endpoint.

Demo: https://github.com/Mossaka/runwasi/tree/otel-wasmtime

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ ttrpc = "0.8.0"
 wat = "1.206"
 windows-sys = "0.52"
 serial_test = "2"
+tracing = "0.1"
 
 [profile.release]
 panic = "abort"
--- a/crates/containerd-shim-wasm/src/sandbox/cli.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/cli.rs
@@ -3,6 +3,7 @@ use std::sync::mpsc::channel;
 use std::sync::Arc;
 
 use containerd_shim::{parse, run, Config};
+use tracing::{instrument, Span};
 use ttrpc::Server;
 
 use crate::sandbox::manager::Shim;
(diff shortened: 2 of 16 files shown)


crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -36,6 +37,7 @@ macro_rules! revision {
     };
 }
 
+#[instrument(skip_all, parent = Span::current(), level= "Info")]
```
jsturtevant, line 40: Do we want all of these at `Info`?  Is there a performance overhead?

Cargo.toml
```diff
@@ -44,6 +44,7 @@ ttrpc = "0.8.0"
 wat = "1.206"
 windows-sys = "0.52"
 serial_test = "2"
+tracing = "0.1"
```
jsturtevant, line 47: Should this be an optional feature in the case downstream shims don't use it? https://github.com/containerd/rust-extensions/issues/260

jsturtevant (commented):
A few questions about the level and ability to[ opt/in out](https://github.com/containerd/rust-extensions/issues/260#issuecomment-2091046206).  Although we support it not everyone might have a OTEL environment set up, if they don't do they pay performance hit for it being enabled?

jsturtevant:
@Mossaka and I chatted offline, there isn't a major overhead (>1ms) so we will go ahead with this and can always adjust as need.

jsturtevant (approved):
LGTM

PR #582, opened 2024-05-06
feat(containerd-shim-wasm): add OpenTelemetry tracing library and feature

Mossaka:
This PR introduces OpenTelemetry feature and new APIs on the core crate to add tracing capabilities to the shim. 

-  Added OpenTelemetry tracing to the main entry point of the shim. It now parses the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable to determine if the shim should be started with OpenTelemetry tracing.
- Enable that in the wasmtime shim 

It builds on top of #564


crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -48,7 +86,17 @@ pub fn shim_main<'a, I>(
     I: 'static + Instance + Sync + Send,
     I::Engine: Default,
 {
+    if cfg!(feature = "opentelemetry") {
+        // read TRACECONTEXT env var that's set by the parent process
+        if let Ok(ctx) = std::env::var("TRACECONTEXT") {
```
Mossaka, line 91: NOTE: It uses `TRACECONTEXT` env var to pass tracing context from the parent shim process to the child process.

Mossaka:
Could you please take a look on this PR 🙏🏻 @devigned @jsturtevant @jprendes

docs/opentelemetry.md
```diff
@@ -0,0 +1,68 @@
...
+
+You may also use the `containerd_shim_wasm::sandbox::shim::OtelConfig` struct to configure the OpenTelemetry tracing manually.
+
+### Running containerd with OpenTelemetry
+
+You can run containerd with OpenTelemetry tracing by setting the following environment variables.
+
+```sh
+OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
+OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc
```
cpuguy83, line 32: Why GRPC?
otlp default is http and I think jaeger expects http.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,144 @@
...
+use opentelemetry_sdk::{runtime, trace as sdktrace, Resource};
+use tracing::Span;
+use tracing_opentelemetry::OpenTelemetrySpanExt as _;
+use tracing_subscriber::layer::SubscriberExt as _;
+use tracing_subscriber::{EnvFilter, Registry};
+
+pub const OTEL_EXPORTER_OTLP_ENDPOINT: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
+
+/// Configuration struct for OpenTelemetry setup.
+pub struct OtelConfig {
```
cpuguy83, line 43: I think this should be called just `Config` or perhaps `OTLPConfig`

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn init_tracer(&self) -> Result<opentelemetry_sdk::trace::Tracer, TraceError> {
+        opentelemetry_otlp::new_pipeline()
+            .tracing()
+            .with_exporter(
+                opentelemetry_otlp::new_exporter()
+                    .tonic()
+                    .with_endpoint(&self.otel_endpoint),
+            )
+            .with_trace_config(sdktrace::config().with_resource(Resource::new(vec![
+                KeyValue::new("service.name", format!("containerd-shim-{}", self.name)),
```
cpuguy83, line 68: This should be taken from env (which I expect the otel libs to do by default).
I guess it should be ok to specify a name but then we should look at that before applying the config.
Also not sure we should prepend `containerd-shim-` in that case.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,144 @@
...
+
+impl Drop for ShutdownGuard {
+    fn drop(&mut self) {
+        // Give tracer provider a chance to flush any pending traces.
+        opentelemetry::global::shutdown_tracer_provider();
+    }
+}
+
+#[derive(Default)]
+pub struct OtelConfigBuilder {
```
cpuguy83, line 117: In the same vein as above `ConfigBuilder`

cpuguy83 (changes requested):

docs/opentelemetry.md
Mossaka: jeager can work on both. but your point is valid, I will add grpc env var as well

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
Mossaka: Good point, it looks like the library will check for `OTEL_SERVICE_NAME` first

docs/opentelemetry.md
```diff
@@ -0,0 +1,68 @@
...
+
+[OpenTelemetry](https://opentelemetry.io/) is a set of libraries, agents, and instrumentation to provide observability (metrics, logs and traces) in applications.
+
+`containerd-shim-wasm` crate has a set of APIs to enable OpenTelemetry tracing in the shim. This document is a guide on how to use OpenTelemetry tracing in your shim.
+
+## Usage
+
+To use OpenTelemetry tracing in your shim, you need to use the `opentelemetry` feature in the `containerd-shim-wasm` crate and the `tokio` crate.
+```toml
+containerd-shim-wasm = { workspace = true, features = ["opentelemetry"] }
```
jsturtevant, line 11: will the `tracing` feature and `opentemetry` feature always be turned on together? should they be the same?

docs/opentelemetry.md
jsturtevant, line 1: Thanks for adding docs!

docs/image.png
jsturtevant, line 1: can you name this something relevant?

docs/image-1.png
jsturtevant, line 1: update the name of this to what it represents

docs/opentelemetry.md
```diff
@@ -0,0 +1,68 @@
...
+### Demo
+
+Assuming you have built or installed the `containerd-shim-wasmtime-v1` shim binary and the demo wasm image following README.md instructions,
+you can run a the wasmtime shim with OpenTelemetry tracing by running the following command
+
+```sh
+sudo ctr run  --net-host --rm --runtime=io.containerd.wasmtime.v1 ghcr.io/containerd/runwasi/wasi-demo-app:latest testwasm /wasi-demo-app.wasm sleep 3
+```
+
+![alt text](image.png)
```
jsturtevant, line 56: add alt text that makes sense for this image

docs/opentelemetry.md
```diff
@@ -0,0 +1,68 @@
...
+
+![alt text](image-1.png)
```
jsturtevant, line 58: same here

crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -37,7 +39,43 @@ macro_rules! revision {
...
+                .otel_endpoint(otel_endpoint)
+                .name(name.to_string())
+                .build()
+                .expect("Failed to build OtelConfig.");
+            let _guard = otel_config
+                .init()
+                .expect("Failed to initialize OpenTelemetry.");
+            shim_main::<I>(name, version, revision, shim_version, config);
+        }
+        Err(_) => {
```
jsturtevant, line 72: should we log as a trace that the value wasn't set to help with debugging?

crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -48,7 +86,18 @@ pub fn shim_main<'a, I>(
     I: 'static + Instance + Sync + Send,
     I::Engine: Default,
 {
+    #[cfg(feature = "opentelemetry")]
```
jsturtevant, line 89: why not put this code in `fn shim_main_with_otel` where this is already guarded by this feature?

crates/containerd-shim-wasmtime/src/main.rs
```diff
@@ -1,6 +1,7 @@
-use containerd_shim_wasm::sandbox::cli::{revision, shim_main, version};
+use containerd_shim_wasm::sandbox::cli::{revision, shim_main_with_otel, version};
 use containerd_shim_wasmtime::WasmtimeInstance;
 
-fn main() {
-    shim_main::<WasmtimeInstance>("wasmtime", version!(), revision!(), "v1", None);
+#[tokio::main]
```
jsturtevant, line 4: Is this needed?  I thought we were injecting the async runtime into the `.install_batch(runtime::Tokio)`.  Similar to how we do it for the containerd client code

jsturtevant (commented):
Great work, awesome to see that this has already been used to solve our startup perf!

docs/opentelemetry.md
Mossaka: yeah I need to update this doc it's not update to date with the work that turns tracing to an optional feature.

crates/containerd-shim-wasm/src/sandbox/cli.rs
Mossaka: Oh it's because `shim_main_with_otel` does not have a span yet, and thus `OtelConfig::get_trace_context()` will return an empty string for the context. 

We could add span to `shim_main_with_otel` though but that will add one more layer of tracing, which I don't think it's bad necessarily.

crates/containerd-shim-wasmtime/src/main.rs
Mossaka: Yes, unfortunately opentelemetry SDK in rust all require tokio runtime

docs/opentelemetry.md
jsturtevant: is the plan to merge the features or keep them seperate?

docs/opentelemetry.md
Mossaka: Just added `OTEL_EXPORTER_OTLP_PROTOCOL` to configure HTTP vs. gRPC

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
Mossaka: Reverted this change to use the default service.name from OTEL_SERVICE_NAME

docs/opentelemetry.md
Mossaka: I am inclined to merge them, but this may change once we use `tracing::log` to replace `log` crate

crates/containerd-shim-wasm/src/sandbox/cli.rs
Mossaka: Can't log a msg here since this will be called by the parent process while the logger is setup at the child process at a later time.

Mossaka:
Did a rebase of the PR and resolved all the comments. Could you please take another look, @cpuguy83 , @jsturtevant 🙏?

crates/containerd-shim-wasm/src/sandbox/shim/cli.rs
```diff
@@ -70,9 +69,6 @@ where
             .and_then(|a| a.get("io.kubernetes.cri.sandbox-id"))
             .unwrap_or(&id);
 
-        setup_namespaces(&spec)
```
jsturtevant, line 73: Let's not make this change in this PR and keep it to https://github.com/containerd/runwasi/pull/503

crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -36,6 +40,43 @@ macro_rules! revision {
...
+    revision: impl Into<Option<&'a str>>,
+    shim_version: impl Into<Option<&'a str>>,
+    config: Option<Config>,
+) where
+    I: 'static + Instance + Sync + Send,
+    I::Engine: Default,
+{
+    match std::env::var(OTEL_EXPORTER_OTLP_ENDPOINT) {
+        Ok(otel_endpoint) => {
+            let mut otel_config = OTLPConfig::builder().otel_endpoint(otel_endpoint);
```
jsturtevant, line 63: I don't see this checking for the "/// If the environment variable is empty, the shim will be started without tracing." `otel_endpoint()` wraps the blank value with `Some()`. Is the empyt value handled by the otel libraries?

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
```diff
@@ -229,8 +229,6 @@ impl<T: Instance + Send + Sync, E: EventSender> Local<T, E> {
             .context("could not spawn thread to wait exit")
             .map_err(Error::from)?;
 
-        debug!("started: {:?}", req);
```
jsturtevant, line 232: Is there a reason to drop this?

crates/containerd-shim-wasmtime/src/main.rs
jsturtevant: I am not familiar enough with rust async, will this cause any issues with shim or is it ok to sprinkle async in various areas?

calebschoepp:
@Mossaka have you tried this with the latest version of the `containerd-shim-spin`? I'm concerned that they're going to conflict b/c they're both setting the tracing subscriber.

Mossaka:
>  I'm concerned that they're going to conflict b/c they're both setting the tracing subscriber.

I will make sure to test the spin shim with these changes.

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
Mossaka: nope will add it back

crates/containerd-shim-wasm/src/sandbox/shim/cli.rs
Mossaka: sounds good

Mossaka:
An interesting error from the build CI

```bash
Dockerfile.musl:4
--------------------
   2 |     FROM $CROSS_BASE_IMAGE
   3 |     
   4 | >>> COPY --from=jorgeprendes[42](https://github.com/containerd/runwasi/actions/runs/9485474098/job/26137500555?pr=582#step:8:43)0/apk-anywhere / /
   5 |     ENV MARCH=${CROSS_CMAKE_SYSTEM_PROCESSOR}
   6 |     RUN apk-init ${MARCH} ${CROSS_SYSROOT}
--------------------
ERROR: failed to solve: jorgeprendes420/apk-anywhere: failed to resolve source metadata for docker.io/jorgeprendes420/apk-anywhere:latest: unexpected status from HEAD request to https://registry-1.docker.io/v2/jorgeprendes420/apk-anywhere/manifests/latest: 503 Service Unavailable
Error: 
   0: could not run container
   1: when building custom image
   2: when building dockerfile
   3: `docker buildx build --progress auto --label 'org.cross-rs.for-cross-target=x86_64-unknown-linux-musl' --label 'org.cross-rs.runs-with=x86_64-unknown-linux-gnu' --label 'org.cross-rs.workspace_root=/home/runner/work/runwasi/runwasi' --tag localhost/cross-rs/cross-custom-runwasi:x86_64-unknown-linux-musl-7e7ca --build-arg 'CROSS_DEB_ARCH=musl-linux-amd64' --build-arg 'CROSS_BASE_IMAGE=ghcr.io/cross-rs/x86_64-unknown-linux-musl:main' --file cross/Dockerfile.musl --output 'type=docker' /home/runner/work/runwasi/runwasi` failed with exit status: 1

Location:
   src/docker/custom.rs:177
```

crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -36,6 +40,47 @@ macro_rules! revision {
...
+    name: &str,
+    version: &str,
+    revision: impl Into<Option<&'a str>>,
+    shim_version: impl Into<Option<&'a str>>,
+    config: Option<Config>,
+) where
+    I: 'static + Instance + Sync + Send,
+    I::Engine: Default,
+{
+    match std::env::var(OTEL_EXPORTER_OTLP_ENDPOINT) {
```
cpuguy83, line 61: There is also `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` that we should look at, which is an alternative to OTEL_EXPORTER_OTLP_ENDPOINT.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn init_tracer_grpc(&self) -> SpanExporterBuilder {
+        opentelemetry_otlp::new_exporter()
+            .tonic()
+            .with_endpoint(&self.otel_endpoint)
+            .into()
+    }
+
+    fn init_tracer(&self) -> Result<opentelemetry_sdk::trace::Tracer, TraceError> {
+        let exporter = match self.otel_protocol.as_str() {
+            OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_PROTOBUF => self.init_tracer_http_protobuf(),
```
cpuguy83, line 109: The sdk should already have a default protocol set in the exporter config.
Do we need to have a custom setup for this?
ref: https://docs.rs/opentelemetry-otlp/latest/src/opentelemetry_otlp/exporter/mod.rs.html#81-93

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,164 @@
...
+};
+pub use opentelemetry_otlp::{OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_PROTOCOL};
+use opentelemetry_sdk::propagation::TraceContextPropagator;
+use opentelemetry_sdk::{runtime, trace as sdktrace};
+use tracing::Span;
+use tracing_opentelemetry::OpenTelemetrySpanExt as _;
+use tracing_subscriber::layer::SubscriberExt as _;
+use tracing_subscriber::{EnvFilter, Registry};
+
+const OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_PROTOBUF: &str = "http/protobuf";
```
cpuguy83, line 42: These are already defined here: https://docs.rs/opentelemetry-otlp/latest/src/opentelemetry_otlp/exporter/mod.rs.html#48

cpuguy83 (changes requested):
I'm not all that familiar with the otel rust SDK but this generally seems good, but we should add support for the extra env var as mentioned in my comments.

crates/containerd-shim-wasm/src/sandbox/cli.rs
Mossaka: This is not supported in the opentelemetry crate. ([issue](https://github.com/open-telemetry/opentelemetry-rust/issues/1892) to track)

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
Mossaka: Yeah but that's not a public facing const (cannot use it outside of the crate)

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
Mossaka: The SDK determines the default protocol based on static features, but if we want to support dynamic protocol , which is to parse environment variable to determine which protocol to use, we can't use the `default()` for `ExportConfig`.

crates/containerd-shim-wasm/src/sandbox/cli.rs
Mossaka: But i guess we still can support it ourselves.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
cpuguy83: Ah you are right, I was thinking this was coming from env.
Kinda frustrating the rust sdk isn't setup for this.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
cpuguy83: Ah right! I guess I'm a bit rusty! 🦀

jsturtevant:
> > I'm concerned that they're going to conflict b/c they're both setting the tracing subscriber.
> 
> I will make sure to test the spin shim with these changes.

Have you tested this with spin shim?  otherwise I think we are good to go

Mossaka:
> Have you tested this with spin shim? 

Yeah I've started testing this with the Spin Shim. I was debugging some issues of not seeing logs in the Spin Shim today.

Mossaka:
I wasn't able to enable opentelemetry on the Spin Shim, as there is a conflict of setting global tracing subscribers at both places which is not allowed:
- in the `main` function of the Spin Shim
- in the `init()` call of `spin-telemetry` crate

The error message looks like

```
thread '<unnamed>' panicked at /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-subscriber-0.3.18/src/util.rs:91:14:
failed to set global default subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Mossaka:
**Updateon the above issue** : 

It seems to me that the container sandbox is blocking traces being exported to the shim's OTLP collector, which lives ouside of the sandbox. So now we have these scenarios
1. If the shim process doesn't enable telemetry, but container does, the Spin runtime traces can be seen from OTLP endpoint.
2. If the shim process enables telemetry, but not container, the shim traces can be seen from OTLP endpoint, but none of trhe traces from inside the container can. 
3. If both are enabled, there is the issue of setting global subscriber twice. The shim needs to set a global subscriber because the shim process is multi-threaded. Unless each TTRPC handler creates a thread-local subscriber, the shim needs a global one. The same applies to the Spin runtime since it's running multi-threaded. 

Based on that, it appears to me that the telemetry in the shim and the container are mutually exclusive. Moving forward, what I hope to achieve is 
1. When the shim process doesn't enable telemetry but container does, the default case, the wasm host's traces are exported to the OTLP endpoint.
2. When the shim process does enable telemetry, it will export both the shim and wasm host's traces to the endpoint. Essentially overriding the container's OTLP endpoint and warns in the logs.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,164 @@
...
+            .tracing()
+            .with_exporter(exporter)
+            .with_trace_config(sdktrace::config())
+            .install_batch(runtime::Tokio)
+    }
+}
+
+/// Shutdown of the open telemetry services will automatically called when the OtelConfig instance goes out of scope.
+#[must_use]
+pub struct ShutdownGuard;
```
jprendes, line 126: **thought:** It's a bit odd that a drop guard has not members and is guarding a global object. Is there any way we can get a hold of the provider we need to shutdown, instead of relying on the global one?

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn drop(&mut self) {
+        // Give tracer provider a chance to flush any pending traces.
+        opentelemetry::global::shutdown_tracer_provider();
+    }
+}
+
+#[derive(Default)]
+pub struct ConfigBuilder {
+    otel_endpoint: Option<String>,
+    otel_protocol: Option<String>,
```
jprendes, line 138: **suggestion:** Remove `otel_` from the members and from the methods in this struct.
```suggestion
    endpoint: Option<String>,
    protocol: Option<String>,
```

crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -36,6 +40,47 @@ macro_rules! revision {
...
+        Ok(otel_endpoint) => {
+            if otel_endpoint.is_empty() {
+                shim_main::<I>(name, version, revision, shim_version, config);
+                return;
+            }
```
jprendes, line 66: **nit:**
```suggestion
        Ok(otel_endpoint) if !otel_endpoint.is_empty() => {
```

crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -36,6 +40,47 @@ macro_rules! revision {
...
 
+/// Main entry point for the shim with OpenTelemetry tracing.
+///
+/// It parses the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable to determine
+/// if the shim should be started with OpenTelemetry tracing.
+///
+/// If the environment variable is not set, the shim will be started without tracing.
+/// If the environment variable is empty, the shim will be started without tracing.
+#[cfg(feature = "opentelemetry")]
+pub fn shim_main_with_otel<'a, I>(
```
jprendes, line 51: **thought:** I think I would prefer another generic argument (with a default value) to `shim_main`, which is a trait for a tracing provider. Then the otel tracer is just one impl of that tracer. The default value could be a no-op provider. Or maybe make the tracing trait part of the shim's `config` argument. I think that would also help with mocking in tests. But it's not a blocker, I am happy for this to go in as is as there's a lot of value in this PR, we can tweak things in a follow up PR.

crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -34,22 +34,44 @@ tokio = { version = "1.38.0", features = [ "full" ] }
 futures = { version = "0.3.30" }
 wasmparser = "0.209.1"
 tokio-stream = { version = "0.1" }
-prost-types = "0.12" # should match version in containerd-shim
+prost-types = "0.12"                                                      # should match version in containerd-shim
```
jprendes, line 37: typo?
```suggestion
prost-types = "0.12" # should match version in containerd-shim
```

crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -34,22 +34,44 @@ tokio = { version = "1.38.0", features = [ "full" ] }
...
 sha256 = { workspace = true }
 
 # tracing
 tracing = { workspace = true, optional = true }
+tracing-subscriber = { version = "0.3", features = [
```
jprendes, line 42: **nit:** bump versions, also a comment about keeping these versions in sync would be nice.

crates/containerd-shim-wasmtime/src/main.rs
jprendes, line 4: If we create the tokio runtime manually in `shim_main_with_otel`, then we woulnd't need to impose tokio as a dependency for runtime crates.
Also it's a bit odd that we are calling a sync/blocking function inside an async function. IIUC, this would break if we use `#[tokio::main(flavor = "current_thread")]` as we would never yield back to the tokio runtime.

jprendes (commented):
Thank you @Mossaka for working on this.
I always learn a lot reviewing your PRs!

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,164 @@
...
+pub use opentelemetry_otlp::{OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_PROTOCOL};
+use opentelemetry_sdk::propagation::TraceContextPropagator;
+use opentelemetry_sdk::{runtime, trace as sdktrace};
+use tracing::Span;
+use tracing_opentelemetry::OpenTelemetrySpanExt as _;
+use tracing_subscriber::layer::SubscriberExt as _;
+use tracing_subscriber::{EnvFilter, Registry};
+
+const OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_PROTOBUF: &str = "http/protobuf";
+const OTEL_EXPORTER_OTLP_PROTOCOL_GRPC: &str = "grpc";
```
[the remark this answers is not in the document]
Mossaka: As you've mentioned these two `String` constants are only used in L.108 so I don't see the value to define an extra enum for this. They are also used for string comparison with the value of environment variables to determine what protocol we are using.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,164 @@
...
+
+/// Configuration struct for OpenTelemetry setup.
+pub struct Config {
+    otel_endpoint: String,
+    otel_protocol: String,
```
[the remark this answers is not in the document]
Mossaka: I've changed them to `traces_*` to differentiate from `metrics` and `logs`.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub fn init(&self) -> anyhow::Result<ShutdownGuard> {
+        let tracer = self.init_tracer()?;
+        let telemetry = tracing_opentelemetry::layer().with_tracer(tracer);
+        set_text_map_propagator(TraceContextPropagator::new());
+
+        let filter = EnvFilter::try_new("info,h2=off")?;
+
+        let subscriber = Registry::default().with(telemetry).with(filter);
+
+        tracing::subscriber::set_global_default(subscriber)?;
```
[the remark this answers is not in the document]
Mossaka: Yeah that's a good point. Since both the shim and spin runtime are multi-threaded, I don't see a good solution here. If we defer setting the default to runtimes, then we will lose the traces for the shim functions. 

I am experimenting moving away from `tracing` library and use `opentelemetry-sdk` to instrument spans. In this way, we don't need to call `set_global_default` and can use named `tracer`.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
Mossaka: I think this particular SDK usage

```rust
opentelemetry_otlp::new_pipeline()
            .tracing()
            .with_exporter(exporter)
            .with_trace_config(sdktrace::config())
            .install_batch(runtime::Tokio)
```

sets a global tracer provider, and thus the only way to shutdown that is to call `global::shutdown_tracer_provider()`.

crates/containerd-shim-wasm/src/sandbox/cli.rs
Mossaka: Oh I think the main reason I have a seperate function for `shim_main_with_otel` is to be able to trace `shim_main` function which is doing a lot of work. Initializing tracer provider **in the function** does not export trace of that function iiuc.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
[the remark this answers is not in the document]
jprendes: Well, they are not used only in L.108, they are used by whoever calls `pub fn otel_protocol(mut self, otel_protocol: String) -> Self;`. Because they are strings, you need an error condition there
```rust
            _ => Err(TraceError::from(
                "Invalid OTEL_EXPORTER_OTLP_PROTOCOL value",
            ))?,
```
If it was an enum, it makes it cleas to consumers as to what are the valid options.

Although I also see your point that this is coming from an env var, and using an enum would defer handling the string matching to the consumer.

What do you think of having that method `OtelConfig::build_from_env() -> Result<OtelConfig>`? that parses the standard env-vars and builds the config acordingly. That would let us encapsulate all the otel logic in this module, and that could also be in charge of doing the string parcing.

crates/containerd-shim-wasmtime/src/main.rs
Mossaka: I've moved tokio runtime creation inside `shim_main_with_otel`.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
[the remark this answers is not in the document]
jprendes: I might be missing something, but I don't understand why we need a prefix, `metrics` and `logs` are members of a different struct, right?
Anyway, it's just a nit (it's not even a public member ^^), feel free to ignore it :-)

crates/containerd-shim-wasm/src/sandbox/cli.rs
jprendes: Could we rename
`pub fn shim_main_with_otel<'a, I>(...)` -> `pub fn shim_main<'a, I>(...)`
`pub fn shim_main(...)` -> `fn shim_main_inner(...)` (not pub)
Then we only expose one of the methods to consumers.
Happy to discuss this on slack :-)

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -54,12 +56,25 @@ pub struct Config {
...
-    /// Creates a new builder for `OtelConfig`.
+    /// Creates a new builder for `Config`.
     pub fn builder() -> ConfigBuilder {
         ConfigBuilder::default()
     }
 
+    /// Returns `true` if traces are enabled, `false` otherwise.
+    ///
+    /// Traces are enabled if either `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` or `OTEL_EXPORTER_OTLP_ENDPOINT` is set and not empty.
+    pub fn traces_enabled() -> bool {
```
jprendes, line 67: **suggestion:** make this a member method instead of a static method, then you can use `self.traces_endpoint`
```suggestion
    pub fn traces_enabled(&self) -> bool {
```

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -134,31 +151,208 @@ impl Drop for ShutdownGuard {
...
+    traces_endpoint: Option<String>,
+    traces_protocol: String,
 }
 
 impl ConfigBuilder {
-    /// Sets the OpenTelemetry endpoint.
-    pub fn otel_endpoint(mut self, otel_endpoint: String) -> Self {
-        self.otel_endpoint = Some(otel_endpoint);
+    /// Sets the OTLP endpoint from environment variables.
+    pub fn traces_endpoint_from_env(mut self) -> Self {
```
jprendes, line 160: **praise:** Nice! I like this, it hides the otel env-vars handling to this module :-)

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -134,31 +151,208 @@ impl Drop for ShutdownGuard {
...
     pub fn build(self) -> Result<Config, &'static str> {
-        let otel_endpoint = self.otel_endpoint.ok_or("otel_endpoint is required")?;
-        let otel_protocol = self
-            .otel_protocol
-            .unwrap_or_else(|| OTEL_EXPORTER_OTLP_PROTOCOL_DEFAULT.to_owned());
         Ok(Config {
-            otel_endpoint,
-            otel_protocol,
+            traces_endpoint: self.traces_endpoint.ok_or("otel_endpoint is required")?,
+            traces_protocol: self.traces_protocol,
```
jprendes, line 180: **question:** protocol is also required, otherwise it ends hitting the error in L.128, right?

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -134,31 +151,208 @@ impl Drop for ShutdownGuard {
...
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
```
jprendes, line 186: **praise:** Thanks!

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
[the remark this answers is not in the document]
Mossaka: Okay, I am actually going to delete the builder and expose `build_from_env` from `Config`.

crates/containerd-shim-wasm/src/sandbox/shim/cli.rs
```diff
@@ -70,11 +69,9 @@ where
             .and_then(|a| a.get("io.kubernetes.cri.sandbox-id"))
             .unwrap_or(&id);
 
+        let (_child, address) = shim::spawn(opts, grouping, vec![])?;
         setup_namespaces(&spec)
             .map_err(|e| shim::Error::Other(format!("failed to setup namespaces: {}", e)))?;
-
-        let (_child, address) = shim::spawn(opts, grouping, vec![])?;
-
```
jsturtevant, line 77: Why did this move up?

jsturtevant (commented):
LGTM

cpuguy83:
One thing to consider, the `Wait` api is a long-running API.
I don't think we want to keep a span running forever.
Something I am looking at for another project is to end the request span once everything is wired up, then start a new span (with the original span as the parent) and on an interval end the span and create a new one (with the previous span as the parent). I'm looking at using 60s for that interval.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
Mossaka: I actually made `traces_enabled` a function instead of a method, since the complexity of it increases after I added `OTEL_SDK_DISABLED`

crates/containerd-shim-wasm/src/sandbox/shim/cli.rs
Mossaka: That was a mistake. Reverted

crates/containerd-shim-wasm/src/sandbox/cli.rs
Mossaka: Done

Mossaka:
> Something I am looking at for another project is to end the request span once everything is wired up, then start a new span (with the original span as the parent) and on an interval end the span and create a new one (with the previous span as the parent). I'm looking at using 60s for that interval.

Interesting idea! I created a minimal example to demonstrate how it could be implemented ([code](https://github.com/Mossaka/SpanSplitter/blob/main/src/main.rs#L10-L37)) and it was more complicated than I had thought originally, and I am not sure if it was a good idea to turn `wait()` function to an async function. 

Do you mind sharing what's the benefits of splitting a long running span to multiple child spans? Also, for each interval child span, do they share the same parent span?

This is the result: 
image

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -83,34 +106,33 @@ impl Config {
...
-        let exporter = match self.otel_protocol.as_str() {
-            OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_PROTOBUF => self.init_tracer_http_protobuf(),
-            OTEL_EXPORTER_OTLP_PROTOCOL_GRPC => self.init_tracer_grpc(),
-            _ => Err(TraceError::from(
-                "Invalid OTEL_EXPORTER_OTLP_PROTOCOL value",
-            ))?,
+        let exporter = match self.traces_protocol {
+            Protocol::HttpBinary => self.init_tracer_http(),
+            Protocol::HttpJson => self.init_tracer_http(),
+            Protocol::Grpc => self.init_tracer_grpc(),
```
jprendes, line 135: **praise:** Great! Now we get a compile error if upstream adds a new protocol, and we can fix it :-)

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
```diff
@@ -132,33 +154,217 @@ impl Drop for ShutdownGuard {
...
+    let traces_protocol = env::var(OTEL_EXPORTER_OTLP_TRACES_PROTOCOL).unwrap_or(
+        env::var(OTEL_EXPORTER_OTLP_PROTOCOL)
+            .unwrap_or(OTEL_EXPORTER_OTLP_PROTOCOL_DEFAULT.to_owned()),
+    );
+    let protocol = match traces_protocol.as_str() {
+        OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_PROTOBUF => Protocol::HttpBinary,
+        OTEL_EXPORTER_OTLP_PROTOCOL_GRPC => Protocol::Grpc,
+        _ => Err(TraceError::from(
+            "Invalid OTEL_EXPORTER_OTLP_PROTOCOL value",
+        ))?,
```
jprendes, line 174: **question:** We never create the `HttpJson` variant?

jprendes (dismissed):
LGTM!
I would still like to eventually see the tracing setup as an injected configuration, but happy to leave that for another PR.

crates/containerd-shim-wasm/src/sandbox/shim/otel.rs
Mossaka: added

crates/containerd-shim-wasm/src/sandbox/shim/mod.rs
```diff
@@ -14,4 +14,7 @@ mod task_state;
 pub use cli::Cli;
 pub(crate) use local::Local;
 #[cfg(feature = "opentelemetry")]
-pub use otel::{Config, ShutdownGuard, OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_PROTOCOL};
+pub use otel::{
+    traces_enabled as otel_traces_enabled, Config as OTLPConfig,
+    ShutdownGuard as OTLPShutdownGuard, OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_PROTOCOL,
```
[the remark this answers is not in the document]
Mossaka: removed them

jprendes (dismissed):
LGTM

Mossaka:
Going to merge this PR in and I will work on the follow ups

1. splitting long-running function to 60s interval spans
2. tracing setup as an injected configuration
3. figure out how to integrate with Spin

PR #589, opened 2024-05-14
wasmtime 20.0.2, enabled async, networking and ip name lookup

vescoc:


crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -29,6 +28,7 @@ impl WasiConfig for DefaultConfig {
     fn new_config() -> Config {
         let mut config = wasmtime::Config::new();
         config.wasm_component_model(true); // enable component linking
+        config.async_support(true);
```
jsturtevant, line 31: What features are you looking to enable with the async support?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -251,36 +260,38 @@ fn prepare_wasi_ctx(
...
-    let wasi_data = WasiCtx {
-        wasi_preview1: wasi_preview1_ctx,
-        wasi_preview2: wasi_preview2_ctx,
-        resource_table: ResourceTable::default(),
-    };
-    Ok(wasi_data)
+        .inherit_network()
+        .allow_tcp(true)
+        .allow_udp(true)
+        .allow_ip_name_lookup(true)
```
jsturtevant, line 294: What's the use case you are trying to unblock?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -251,36 +260,38 @@ fn prepare_wasi_ctx(
...
-            "/",
-        );
-    let wasi_preview2_ctx = wasi_preview2_builder.build();
-    let wasi_data = WasiCtx {
-        wasi_preview1: wasi_preview1_ctx,
-        wasi_preview2: wasi_preview2_ctx,
-        resource_table: ResourceTable::default(),
-    };
-    Ok(wasi_data)
+        .inherit_network()
```
jsturtevant, line 291: same here, what are the use cases?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -251,36 +260,38 @@ fn prepare_wasi_ctx(
...
+        .allow_tcp(true)
+        .allow_udp(true)
```
jsturtevant, line 293: These two are on by default, but being explicit is probably good idea.

jsturtevant (commented):
Thanks for the for upgrade to wasmtime. I am interested in the use cases that require the additional features you've enabled.  Ultimately, we might need a way customize the runtimes engines.  This has come up in https://github.com/spinkube/containerd-shim-spin/issues/115 as well

vescoc:
> Thanks for the for upgrade to wasmtime. I am interested in the use cases that require the additional features you've enabled. Ultimately, we might need a way customize the runtimes engines. This has come up in [spinkube/containerd-shim-spin#115](https://github.com/spinkube/containerd-shim-spin/issues/115) as well

Well, the use case that triggered these configuration sets is trivial. I'm investigating the current limits of using WASI on non-trivial "tokio-type" basic applications. I'm converting my [protohackers](https://protohackers.com/) problem solutions  which are tokio based, to solutions based on WASI 0.2 components. To run the solutions I need IP name lookup, networking, tcp and udp enabled. Simple 😄 

If you want to see what I'm doing on protohackers, [this is my repository](https://github.com/vescoc/protohackers).  Into the directory wasi there are the solutions using Rust / WASI.

I have currently converted the first 4 solutions. I'll proceed with the others in the next few days. I need to implement the missing pieces in my wasi_async framework (next step: UDP).

crates/containerd-shim-wasmtime/src/instance.rs
vescoc: I tried a sync implementation but got an error when running the demo app. If I remember correctly the error looked like [this](https://stackoverflow.com/questions/77696960/how-to-disable-async-in-wasmtimecomponent). I solved the problem by enabling async and using async linker and so on.

Mossaka:
Yeah containerd config options for the shim would be a good place to put configurations that might change the runtime behaviours.

crates/containerd-shim-wasmtime/Cargo.toml
```diff
@@ -12,26 +12,8 @@ oci-spec = { workspace = true, features = ["runtime"] }
...
-    "pooling-allocator",
-    "demangle",
-    "addr2line",
-    "coredump",
-    "debug-builtins",
-    'component-model',
-]}
-wasmtime-wasi = { version = "17.0", features = ["exit"] }
-wasi-common = "17.0"
+wasmtime = { version = "20.0.2", features = ["async"] }
```
vescoc, line 15: [Wasmtime 21 was released](https://github.com/bytecodealliance/wasmtime/blob/main/RELEASES.md)
I saw that by updating the version the repository compiles and the tests pass. If you prefer, I open a new PR or rename this one with the transition directly to wasmtime 21

jsturtevant: Sure, Would you mind doing the bump to Wasmtime 21 as a separate PR?  I am still not sure the implications of turning on the async support and a few of the other options enabled here.  I need a little more time to understand what turning these features means.  Maybe other reviewers have some thoughts too

jsturtevant:
> Well, the use case that triggered these configuration sets is trivial. I'm investigating the current limits of using WASI on non-trivial "tokio-type" basic applications. I'm converting my [protohackers](https://protohackers.com/) problem solutions which are tokio based, to solutions based on WASI 0.2 components. To run the solutions I need IP name lookup, networking, tcp and udp enabled. Simple 😄

>
> If you want to see what I'm doing on protohackers, [this is my repository](https://github.com/vescoc/protohackers). Into the directory wasi there are the solutions using Rust / WASI.

This is really cool!

crates/containerd-shim-wasmtime/Cargo.toml
vescoc: > Sure, Would you mind doing the bump to Wasmtime 21 as a separate PR? I am still not sure the implications of turning on the async support and a few of the other options enabled here. I need a little more time to understand what turning these features means. Maybe other reviewers have some thoughts too

For the transition to wasmtime 21 in my opinion we can wait as long as, we simply need to update the dependencies of Cargo.toml.

I'm currently testing the new version on my k8s.

Please take the time to evaluate the impacts of the proposed changes. As already indicated I attempted to keep the code with async disabled but I had got failed tests so I opted to switch to async mode. This should not impact other shims.

I hope other people can try my modifications on more serious implementations than the my toy one with protohackers.

Mossaka (approved):
LGTM!

Mossaka:
Need a rebase

Mossaka (approved):

Mossaka:
Thanks, again! @vescoc

PR #604, opened 2024-05-22
fix(release): fix another syntax error on runtime_sub

Mossaka:
well, it turns out that #603 does not fix the whole issue. This PR is an attempt to address it.


jsturtevant (approved):
LGTM

PR #508, opened 2024-03-07
feat(wamr): add wasm-mico-runtime shim implementation

Mossaka:
this commit adds an additional shim implementation: Wamr using wamr-rust-sdk. At the moment, the shim isn't working with an error `"thread signal env initialized failed"` which I am not sure exactly sure what it means. Will ask the Wamr team for clarification. There are also some pain points using the newly created [wamr-rust-sdk](https://github.com/bytecodealliance/wamr-rust-sdk) which I wrote as TODOs in the comment. 

This PR should close #337

FYI @squillace @0xE282B0 @lum1n0us

PR #642, opened 2024-07-13
add wasm-mico-runtime shim implementation 

macko99:
this is an updated [PR](https://github.com/containerd/runwasi/pull/508) using  adding addtional shim implementation: Wamr using latest [wamr-rust-sdk](https://github.com/bytecodealliance/wamr-rust-sdk). 

It is still not working, the shim exits with one of the errors:
-"thread signal env initialized failed"
-without any error
-"error running start function: Failed to create instance: InstantiationFailure("error while pre-opening mapped directory: invalid map" 
-"Running "_start"" then "Error: ExecutionError("Exception: unreachable")"

My module has only the main function defined.
I have tried disabling hardware  checks in WAMR since I'm working on the VM, [commit](https://github.com/macko99/wamr-rust-sdk/commit/b89f685eba764fd784be103a7659fe19bb215482)

FYI @squillace @0xE282B0 @lum1n0us @Mossaka

### DIFF
--- a/crates/containerd-shim-wamr/src/instance.rs
+++ b/crates/containerd-shim-wamr/src/instance.rs
@@ -6,10 +6,7 @@ use wamr_rust_sdk::function::Function;
 use wamr_rust_sdk::instance::Instance as WamrInstnace;
 use wamr_rust_sdk::module::Module;
 use wamr_rust_sdk::runtime::Runtime;
-
-// use wamr_rust_sdk::value::WasmValue;
 use wamr_rust_sdk::wasi_context::WasiCtxBuilder;
-// use wamr_rust_sdk::RuntimeError;
 
 pub type WamrInstance = Instance<WamrEngine>;
 
@@ -24,7 +21,6 @@ unsafe impl Sync for WamrEngine {}
 
 impl Default for WamrEngine {
     fn default() -> Self {
-        log::info!("Create a WAMR runtime");
         let runtime = Runtime::new().unwrap();
         Self { runtime }
     }
@@ -32,7 +28,6 @@ impl Default for WamrEngine {
 
 impl Clone for WamrEngine {
     fn clone(&self) -> Self {
-        log::info!("Clone a WAMR runtime");
         let runtime = Runtime::new().unwrap();
         Self { runtime }
     }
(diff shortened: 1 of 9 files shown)


crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,102 @@
...
+impl Default for WamrEngine {
+    fn default() -> Self {
+        let runtime = Runtime::new().unwrap();
+        Self { runtime }
+    }
+}
+
+impl Clone for WamrEngine {
+    fn clone(&self) -> Self {
+        let runtime = Runtime::new().unwrap();
```
Mossaka, line 31: **question**:  is this a deep clone or shallow clone? Runwasi clones the runtime so that it could share to different instances. (i.e. https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sandbox/instance.rs#L98)

lum1n0us: In wamr-rust-sdk, it'll execute +1 on runtime's reference counter.

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,102 @@
...
+        let mod_name = name.unwrap_or_else(|| "main".to_string());
+
+        let mut module = Module::from_buf(&self.runtime, &wasm_bytes, &mod_name).map_err(|e| {
+            anyhow::Error::msg(format!("Failed to create module from bytes: {:?}", e))
+        })?;
+
+        log::info!("Create a WASI context");
+
+        let wasi_ctx = WasiCtxBuilder::new()
+        .set_pre_open_path(vec!["/"], vec!["/"])
```
lum1n0us, line 69:
```suggestion
        .set_pre_open_path(vec!["/"], vec![])
```

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,102 @@
...
+        .set_env_vars(envs.iter().map(String::as_str).collect())
+        .build();
```
lum1n0us, line 71:
```suggestion
        .set_arguments(args.iter().map(String::as_str).collect())
        .build();
```

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,102 @@
...
+        // TODO: bug: failed at line above saying: `thread signal env initialized failed`
+
+        log::info!("redirect stdio");
+        stdio.redirect()?;
+
+        log::info!("Running {func:?}");
+        let function = Function::find_export_func(&instance, &func)
+            .map_err(|e| anyhow::Error::msg(format!("Failed to find function: {:?}", e)))?;
+        let status = function
+            .call(&instance, &Vec::new())
```
lum1n0us, line 92: just in case. 

```suggestion
            .call(&instance, &vec![]))
```

lum1n0us:
Need to sync up latest wamr-rust-sdk. But sill need to keep the workaround: `.define("WAMR_DISABLE_HW_BOUND_CHECK", "1")`.

After all above modification, we still have one problem that happens when using `println!()` in wasm. You will see more details in https://github.com/containerd/runwasi/pull/508#issuecomment-2249263378

crates/containerd-shim-wamr/src/instance.rs
lum1n0us: in order to pass arguments (like "wasi-demo-oci.wasm echo 'hi'") to wasm functions.

lum1n0us:
https://github.com/bytecodealliance/wamr-rust-sdk/releases/tag/v1.1.0

lum1n0us:
@Mossaka any update? Let me know if you need any help.

PR #653, opened 2024-07-23
feat(otel): split long-running wait function to child spans

Mossaka:
this commit splits the long-running wait function and exports child spans every 60 seconds.

Example:
image

Note: in the example, I changed the interval time to 2s instead of 60s.

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/shim/local.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/shim/local.rs
@@ -387,7 +387,31 @@ impl<T: Instance + Sync + Send, E: EventSender> Task for Local<T, E> {
     #[cfg_attr(feature = "tracing", tracing::instrument(parent = tracing::Span::current(), skip_all, level = "Info"))]
     fn wait(&self, _: &TtrpcContext, req: WaitRequest) -> TtrpcResult<WaitResponse> {
         debug!("wait: {:?}", req);
-        Ok(self.task_wait(req)?)
+
+        #[cfg(feature = "opentelemetry")]
+        {
+            use tracing::{span, Level, Span};
+
+            let (tx, rx) = std::sync::mpsc::channel();
+            // Start a thread to export interval span for long wait
+            let parent_span = Span::current();
+            let _ = thread::spawn(move || loop {
+                let current_span =
+                    span!(parent: &parent_span, Level::INFO, "task wait 60s interval");
+                let _enter = current_span.enter();
+                if rx.recv_timeout(Duration::from_secs(60)).is_ok() {
+                    break;
+                }
+            });
+            let result = self.task_wait(req)?;
+            tx.send(()).unwrap();
+            Ok(result)
+        }
+
+        #[cfg(not(feature = "opentelemetry"))]
+        {
+            Ok(self.task_wait(req)?)
+        }
     }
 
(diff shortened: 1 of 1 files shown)


crates/containerd-shim-wasm/src/sandbox/shim/local.rs
```diff
@@ -387,7 +387,31 @@ impl<T: Instance + Sync + Send, E: EventSender> Task for Local<T, E> {
...
+        #[cfg(feature = "opentelemetry")]
+        {
+            use tracing::{span, Level, Span};
+
+            let (tx, rx) = std::sync::mpsc::channel();
+            // Start a thread to export interval span for long wait
+            let parent_span = Span::current();
+            let _ = thread::spawn(move || {
+                while rx.recv_timeout(Duration::from_secs(60)).is_err() {
+                    let current_span =
```
cpuguy83, line 400: I think in order to make sure each span is sent we need to call `exit()` on the previous span?

cpuguy83 (changes requested):

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
Mossaka: when `_enter` goes out of scope here, span automatically exits

cpuguy83: Oh yeah that's nice, but then when does `_enter` go out of scope here? My intuition says on the next loop iteration.

jprendes: IIUC this goes out of scope immediately. Does it generate a tiny span every 60 seconds? Is that what we want? Or would it be better a span taking the whole 60 seconds?

Mossaka: Ah yes it generates tiny spans. I agree a 60s span makes more sense 👍

Mossaka: Done

Mossaka: Is this current change loog good to y'all?

jprendes: LGTM.
I wonder however if we can achieve this spawning a task instead of a thread. 
Also, I don't fully understand why we need this change.

cpuguy83: It was something that came up in another project for long-running requests.
Basically the span can't be exported until it exits.

cpuguy83: I think in this case though, if we don't exit the parent span first then we'll have an orphan span until the parent exits.

devigned (approved):
LGTM!

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
Mossaka: getting back to this PR, is there anything you wanna change or are you good with merging as is? @cpuguy83

cpuguy83: It may be good to exit the parent span, but honestly this  is probably fine as is.

cpuguy83 (approved):

PR #654, opened 2024-07-23
Use RPITIT in `PathResolve`

primoly:
As the return-position impl trait in trait (RPITIT) feature is stable since [Rust 1.75.0
](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html#async-fn-and-return-position-impl-trait-in-traits)

### DIFF
--- a/crates/containerd-shim-wasm/src/container/path.rs
+++ b/crates/containerd-shim-wasm/src/container/path.rs
@@ -1,10 +1,6 @@
 use std::path::{Path, PathBuf};
 
 pub trait PathResolve {
-    // TODO: Once RPITIT lands in stable, change the return types from
-    //       `-> Box<dyn Trait>` to `-> impl Trait`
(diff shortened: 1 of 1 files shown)


crates/containerd-shim-wasm/src/container/path.rs
```diff
@@ -13,10 +9,10 @@ pub trait PathResolve {
     // Further contraints can be added by calling filtering the returned iterator.
     fn resolve_in_dirs<'a>(
```
jprendes, line 10: I think `'a` is not needed anymore, right?
```suggestion
    fn resolve_in_dirs(
```

jprendes (commented):
Thanks for your contribution!
Could you also update the `impl` block? The methods from line 35 onwards.

Mossaka (approved):
LGTM, thanks!

Issue #637, opened 2024-07-07
ctr: failed to create shim task, wasmtime can't handle spec?


matsbror:
I have compiled and installed the wasmtime shim (and wasmedge and wasmer), and I am able to run a wasm module using docker:

```
docker run --rm --runtime io.containerd.runtime.wasmtime.v1 matsbror/serwasm2:0.1
```

However, I would like to run it directly using `ctr` if possible, but I get this error:

```
sudo ctr run --rm --runtime io.containerd.runtime.wasmtime.v1 docker.io/matsbror/serwasm2:0.1 myapp bogus
ctr: failed to create shim task: Others("exec process failed with error error in executing process : wasmtime executor can't handle spec"): unknown
```

I can run a spin app directly using `ctr` with the spin shim from spinkube.

Mossaka:
It's hard to tell what's going on with this error, and perhaps it would be helpful to add `containerd` logs. 

If you are running containerd as a service in Linux, you can check containerd logs by `journalctl -u containerd`. It seems like this error msg is from https://github.com/Mossaka/runwasi/blob/main/crates/containerd-shim-wasm/src/sys/unix/container/executor.rs#L40

dvperez-grad:
Same error here but inside a pod of Kubernetes (Kind) and using WasmEdge. Kubectl logs empty. 

![imagen](https://github.com/user-attachments/assets/0c3e1ece-2269-48e0-8c4e-39f5bba0c340)

Any ideas?

Mossaka:
The error message should be more explicity after #655.

PR #655, opened 2024-07-23
Improved error message for `Executor::exec`

Mossaka:
this commit imrpoves the error message for exec function, particularlly when the user forgot to provide Wasm file path as the arg0 for the container, it will log an error message in containerd

Related https://github.com/containerd/runwasi/issues/637

### DIFF
--- a/crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
+++ b/crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
@@ -90,12 +90,20 @@ impl<E: Engine> Executor<E> {
 
     fn inner(&self, spec: &Spec) -> &InnerExecutor {
         self.inner.get_or_init(|| {
-            if is_linux_container(&self.ctx(spec)).is_ok() {
-                InnerExecutor::Linux
-            } else if self.engine.can_handle(&self.ctx(spec)).is_ok() {
-                InnerExecutor::Wasm
-            } else {
-                InnerExecutor::CantHandle
+            let ctx = &self.ctx(spec);
+            match is_linux_container(ctx) {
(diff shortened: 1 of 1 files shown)


crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
```diff
@@ -90,12 +90,20 @@ impl<E: Engine> Executor<E> {
...
+            match is_linux_container(ctx) {
+                Ok(_) => InnerExecutor::Linux,
+                Err(err) => {
+                    log::debug!("error checking if linux container: {err}. Fallback to wasm container");
+                    match self.engine.can_handle(ctx) {
+                        Ok(_) => InnerExecutor::Wasm,
+                        Err(err) => {
+                            // log an error and return
+                            log::error!("error checking if wasm container: {err}. Note: arg0 must be a path to a Wasm file");
+                            InnerExecutor::CantHandle
```
jsturtevant, line 103: it would be nice if we could pass some context to this error message so it got bubbled up to the CTR/cri error message with out having to look into the logs. 

It looks like we pass the name as context, maybe we could extend it take in a message too?

https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm/src/sys/unix/container/executor.rs#L40

jsturtevant (approved):
/lgtm

This is an improvement and we can follow up with a change to the libcontainer api if we want to

crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
Mossaka: Yeah good point! I was thinking about that. I will make a PR against youki anad see if they like this idea.

PR #656, opened 2024-07-24
Remove dependency on prost-types

jprendes:
This change removes the dependency on `prost-types`, which otherwise needs to be kept in sync with the version in `containerd-client` (see for example  https://github.com/containerd/runwasi/pull/651)


Mossaka (approved):
That's a smart way of dealing with it. Thanks and LGTM!

PR #665, opened 2024-08-13
Updating Windows getting started doc and Makefile misc

danbugs:


devigned (approved):
We should consider some comment linting. Thank you!

Issue #619, opened 2024-06-12
Where do we define the Wasm "container" boundry


jsturtevant:
We are doing a bunch of set up work for the wasm runtime inside the container and using rust log crate.  We are effectively still operating as the "shim" until we invoke the Wasm binary for example https://github.com/containerd/runwasi/blob/7cc5e5a8025f1508c264dee1faf1c8c57cf645a1/crates/containerd-shim-wasmtime/src/instance.rs#L163

This leads to some confusion from an end user perspective in some scenarios. For instance if a user is using a Rust wasm application and they set `RUST_LOG` on their podspec, then it will switch logging for everything in https://github.com/containerd/runwasi/blob/7cc5e5a8025f1508c264dee1faf1c8c57cf645a1/crates/containerd-shim-wasm/src/container/engine.rs#L16.  

This could also happen when we start to enable tracing in https://github.com/containerd/runwasi/pull/582

Since there is some additional set up of the wasm runtime, this leads an issue of where the boundary should live for the container vs wasm.

jsturtevant:
some further conversation and context in https://cloud-native.slack.com/archives/C06PC7JA1EE/p1717097248450109

Mossaka:
Let me clarify why we are seeing "if a user is using a Rust wasm application and they set RUST_LOG on their podspec, then it will switch logging for everything in `run_wasi`"

There are two shim processes when started:
- Parent shim process: started by containerd and uses containerd's env vars (e.g. `RUST_LOG` from containerd)
- Child shim process: started by containerd and inherits parent shim process's env vars (e.g. `RUST_LOG` from containerd)

When the child shim process invokes youki's `ContainerBuilder::build()` it reads the env vars from the Container Spec, hence from the PodSepc, and changes the process's env vars (e.g. [this line](https://github.com/containers/youki/blob/e19a81979c5d2e09697ccd09a17e00f6e8f7fe9b/crates/libcontainer/src/process/container_init_process.rs#L574)).

This happens prior to `run_wasi`. Hence everything after `run_wasi` is using the env vars from PodSpec.

---

Back to your question: "where the boundary should live for the shim vs wasm?". I think from the user's perspective, they would have assumed that their application environment variables (i.e. from PodSpec) should only be seen by their application (i.e. WASI environment consumed by Wasm modules). 

The next question is how do we achieve that?

I am thinking about either
[...]

PR #668, opened 2024-09-09
core/container/context: container env vars are only passed to WASI

Mossaka:
this commit uses the new setup_envs() method from libcontainer and brings the following changes:

1. the Executor trait now provides a no-op for setup_envs, which means youki is no longer changing shim process's environment variables.
2. the container env vars are passed down to WASI context, meaning only the Wasm modules can see container env vars.
3. This sets a clear boundary between shim process and wasm workloads.

Note that the wasm runtime like Wasmtime is still running on the shim process, which means they will still see shim's env vars passed down from containerd

see more details at #619 

TODO:
- [x] we should document the expectations of handling env vars for shim authors using the `containerd_shim_wasm::container::Engine` trait

### DIFF
--- a/crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
+++ b/crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
@@ -1,4 +1,5 @@
 use std::cell::OnceCell;
+use std::collections::HashMap;
 use std::fs::File;
 use std::io::Read;
 use std::os::unix::prelude::PermissionsExt;
@@ -8,7 +9,7 @@ use anyhow::{bail, Context, Result};
 use libcontainer::workload::default::DefaultExecutor;
 use libcontainer::workload::{
     Executor as LibcontainerExecutor, ExecutorError as LibcontainerExecutorError,
-    ExecutorValidationError,
+    ExecutorSetEnvsError, ExecutorValidationError,
 };
 use oci_spec::image::Platform;
 use oci_spec::runtime::Spec;
@@ -64,6 +65,21 @@ impl<E: Engine> LibcontainerExecutor for Executor<E> {
                 };
             }
         }
+    }
+
+    // This is an no-op for the Wasm `Executor`. Instead of youki's libcontainer setting the envs
+    // in the shim process, the shim will manage the envs itself. The expectation is that the shim will
+    // call `RuntimeContext::envs()` to get the container's envs and set them in the `Engine::run_wasi`
+    // function. This way, the shim can decide how to pass the envs to the WASI context.
+    //
+    // See the following issues for more context:
+    // https://github.com/containerd/runwasi/issues/619
+    // https://github.com/containers/youki/issues/2815
+    fn setup_envs(
+        &self,
+        _: HashMap<String, String>,
+    ) -> std::result::Result<(), ExecutorSetEnvsError> {
+        Ok(())
     }
 }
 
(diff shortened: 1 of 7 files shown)


Mossaka:
cc @kate-goldenring this PR will make the boundary between shim process and wasi context clearer. For the spin shim configuration as described by [003-skips](https://github.com/spinkube/skips/tree/main/proposals/003-skips), I think what we need to do

1. call `ctx.envs()` to get a list of container env vars
2. take out spin execution vars like `SPIN_APPLICATION_VARIABLE_PREFIX` and use the to configure spin runtime
3. for the rest of vars, prefix with `SPIN_VARIABLE` to make them Spin application variables. 

the nice thing is you no longer need to invoke process's `env::var` and `env::set_var`.

Mossaka:
This benchmark CI failure is strange, not related to the changes I made and I even can't reproduce in my local...

crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
```diff
@@ -65,6 +66,14 @@ impl<E: Engine> LibcontainerExecutor for Executor<E> {
             }
         }
     }
+
+    fn setup_envs(
```
cpuguy83, line 70: Can you write a comment why this needs to be a no-op?

kate-goldenring:
> the nice thing is you no longer need to invoke process's env::var and env::set_var.

@Mossaka, the Spin environment variables provider for application variables dynamically looks at the process env for `SPIN_VARIABLE_` prefixed items. So we need to prefix the env vars and then reset them in the execution environment. I do think we will still need to use `env::set_var` unless we change the provider to support a static list of values

Mossaka:
> > the nice thing is you no longer need to invoke process's env::var and env::set_var.
> 
> @Mossaka, the Spin environment variables provider for application variables dynamically looks at the process env for `SPIN_VARIABLE_` prefixed items. So we need to prefix the env vars and then reset them in the execution environment. I do think we will still need to use `env::set_var` unless we change the provider to support a static list of values

I should clarify, after this PR merges, the container env vars should be in `ctx.envs()`. `env::var()` only gives you env vars from the shim process, which come from containerd, instead of the OCI spec. What you really want for the application variables are these from the OCI spec, IIUC.

kate-goldenring:
> What you really want for the application variables are these from the OCI spec, IIUC.

I agree that these are the ones we want. But i am pointing out that the env var provider is tightly tied to the assumption that it gets the application variables from it's environment (using `env::var`). 

> Can we somehow declare all the vars in ctx.envs() minus execution variables are Spin application variables?

We will need to update the env var provider to not fetch application variables from the environment rather from a different method. See here how the current provider references the environment for the application var:
https://github.com/fermyon/spin/blob/c4d6f456d51419296a6cae05722d66639dfbd7f6/crates/factor-variables/src/spin_cli/env.rs#L33-L34

crates/containerd-shim-wasm/src/sys/unix/container/executor.rs
Mossaka: yeah I added a comment explaining why this is a no-ope

Mossaka:
> i am pointing out that the env var provider is tightly tied to the assumption that it gets the application variables from it's environment (using env::var)

I see. If that's the case, I have no problem with calling `env::set_var()` to get them into the shim process. This will be called from shim's `run_wasi` method which the spin shim can decide what are app vars and what are execution vars since it can see both the shim process's vars and OCI spec vars.

kate-goldenring:
> I see. If that's the case, I have no problem with calling env::set_var() to get them into the shim process.

Sounds good. That is an easy start -- just get the vars from `ctx.envs()` first. If we find it to be odd behavior, we can refactor the provider in the future

jprendes (approved):
LGTM

utam0k:
Thanks, @Mossaka !

PR #680, opened 2024-09-30
Updated Scripts for setup new dev environment

fawadasaurus:
I setup a new dev environment and wanted to get the project running on my OSX machine. I installed a fresh Linux VM and was able to run `make build` and `make test` with this setup.

I encountered one test failure. I don't believe this failure would only impact me as the oci_spec default for environment variables sets 2 environment variables. So I set the 
```rust
impl Default for Process {
    fn default() -> Self {
        Process {
....
            // Sets linux default enviroment for binaries and default xterm emulator
            env: vec![
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin".into(),
                "TERM=xterm".into(),
            ]
```

### DIFF
--- a/scripts/setup-linux.sh
+++ b/scripts/setup-linux.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 sudo apt -y update
-sudo apt install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev libzstd-dev protobuf-compiler
+sudo apt install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev libzstd-dev protobuf-compiler libssl-dev
 
 if [ ! -z "$CI" ] && ! mount | grep cgroup; then
     echo "cgroup is not mounted" 1>&2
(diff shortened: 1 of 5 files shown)


Mossaka:
Thanks for fixing this bug! I just realized that our build pipeline is not set up correctly to fail when the `make test-` fails...

fawadasaurus:
> Thanks for fixing this bug! I just realized that our build pipeline is not set up correctly to fail when the `make test-` fails...

I can try to address that as well in this PR if you would like.

scripts/setup-linux.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 sudo apt -y update
-sudo apt install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev libzstd-dev protobuf-compiler
+sudo apt install -y pkg-config libsystemd-dev libdbus-glib-1-dev build-essential libelf-dev libseccomp-dev libclang-dev libzstd-dev protobuf-compiler libssl-dev
```
Mossaka, line 3: Can you please add `libssl` dep to `cross/Dockerfile_*` as well?

Mossaka (approved):

PR #686, opened 2024-10-02
add forced exit into makefile

fawadasaurus:


jprendes:
I believe this is already the default behavior of `make`, i.e., to stop after any non-zero exit code.

I think the problem with CI is somewhere else.

fawadasaurus:
Yea, I have tested different variations and I believe that the issue is not with Make or Github Actions.

That means that the issue is likely with cross-rs not properly propagating errors from inside the docker container. I will do more experimentation to try to figure it out.

jprendes:
Fixed by #737

PR #699, opened 2024-10-08
.github/workflows/release: no cargo publish if the repo is not upstream

Mossaka:
forked repo cannot publish to crates.io


jprendes (approved):
LGTM

PR #700, opened 2024-10-11
Fix wasmtime precompilation

andreiltd:
Check the `wasm` binary type and choose proper precompilation method.

### DIFF
--- a/crates/containerd-shim-wasmtime/src/instance.rs
+++ b/crates/containerd-shim-wasmtime/src/instance.rs
@@ -117,7 +117,17 @@ impl<T: WasiConfig> Engine for WasmtimeEngine<T> {
                 continue;
             }
 
-            let compiled_layer = self.engine.precompile_module(&layer.layer)?;
+            use WasmBinaryType::*;
+
+            let compiled_layer = match WasmBinaryType::from_bytes(&layer.layer) {
+                Some(Module) => self.engine.precompile_module(&layer.layer)?,
+                Some(Component) => self.engine.precompile_component(&layer.layer)?,
+                None => {
+                    log::warn!("Unknow WASM binary type");
+                    continue;
+                }
+            };
+
             compiled_layers.push(Some(compiled_layer));
         }
 


crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -117,7 +117,17 @@ impl<T: WasiConfig> Engine for WasmtimeEngine<T> {
...
+
+            let compiled_layer = match WasmBinaryType::from_bytes(&layer.layer) {
+                Some(Module) => self.engine.precompile_module(&layer.layer)?,
+                Some(Component) => self.engine.precompile_component(&layer.layer)?,
+                None => {
+                    log::warn!("Unknow WASM binary type");
+                    continue;
+                }
+            };
+
```
Mossaka, line 130: suggest: move this logic to Engine and add a method called `fn precompile()`. It is the job of the engine to handle the type of `compiled_layer`

jprendes: So instead of `Engine::precompile` like we have now, it would be `Engine::precompile_module` and `Engine::precompile_component`?
Sounds god to me.
Alternatively, a layer could have a `is_component(&self) -> bool` method, WDYT?

jsturtevant: > It is the job of the engine to handle the type of compiled_layer

Not sure I understand this statement.  note we have this type of logix else where in this shim https://github.com/andreiltd/runwasi/blob/e07255e7d1b6ea07ff3daa783782defd317d18b5/crates/containerd-shim-wasmtime/src/instance.rs#L258-L259

> it would be Engine::precompile_module and Engine::precompile_component?
I am not excited about chaning the api or making the engine have to iterate through the layers.  We delegate that to the instance so shims can handle there own layer types.

> Alternatively, a layer could have a is_component(&self) -> bool method, WDYT?

Something like this makes sense if we want to move the logic our of here.  Maybe a enum instead of a bool though?  In spin shim it currently supports of ther layer types.  We don't really wanto

Mossaka: > In spin shim it currently supports other layer types

Interesting. I'd suggest we rename `WasmLayer` to `OCILayer` to be more representative of different layer types in the artifact. 

I'd suggest we can actually have a `WasmLayer` that corresponds to `layer.config.media_type() == "application/vnd.wasm.content.layer.v1+wasm"`.

And then it makes sense to add `is_component()` to `WasmLayer`.

jprendes (approved):
We discussed this issue in the community meeting, and we agreed that we are happy to proceed with this PR in its current form.

crates/containerd-shim-wasmtime/src/instance.rs
jprendes: Marking this as resolved as per the community meeting discussion.
Feel free to open a separate issue to address any remaining concerns.

PR #705, opened 2024-10-25
Add graceful server shutdown

andreiltd:
This patch adds graceful shutdown functionality to the Wasmtime shim server loop. The shutdown process works as follows:

- The code that spawns the top-level loop creates a `CancellationToken`, which is shared across all cancellable tasks.
- A tracker is used to manage any tasks spawned within the server loop.
- A signal handler is installed to listen for signals. Upon receiving the first `Interrupt` signal, it calls cancel and waits for the server task to finish gracefully, allowing all spawned tasks to complete.
- On receiving a second `Interrupt` signal, the server loop exits immediately.
- Similarly, upon receiving a `Terminate` signal, the server loop finishes abruptly.
- The return code indicates which of these shutdown scenarios occurred. If the server has finished abruptly the `128 + SIG` will be used. See https://tldp.org/LDP/abs/html/exitcodes.html for more details.

### DIFF
--- a/crates/containerd-shim-wasmtime/Cargo.toml
+++ b/crates/containerd-shim-wasmtime/Cargo.toml
@@ -6,10 +6,11 @@ edition.workspace = true
 [dependencies]
 anyhow = { workspace = true }
 containerd-shim-wasm = { workspace = true, features = ["opentelemetry"] }
+libc = { workspace = true }
 log = { workspace = true }
 hyper = { workspace = true }
 tokio = { workspace = true, features = ["signal", "macros"] }
-libc = { workspace = true }
+tokio-util = { workspace = true, features = ["rt"] }
 
 wasmtime = { workspace = true }
 wasmtime-wasi = { workspace = true }
--- a/crates/containerd-shim-wasmtime/src/http_proxy.rs
+++ b/crates/containerd-shim-wasmtime/src/http_proxy.rs
@@ -5,10 +5,12 @@ use std::collections::HashMap;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
+use std::time::Duration;
 
 use anyhow::{bail, Result};
 use containerd_shim_wasm::container::RuntimeContext;
 use hyper::server::conn::http1;
+use tokio::net::{TcpListener, TcpStream};
 use tokio_util::sync::CancellationToken;
 use tokio_util::task::TaskTracker;
 use wasmtime::component::ResourceTable;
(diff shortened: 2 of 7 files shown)


crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -302,9 +306,32 @@ where
         log::debug!("loading wasm component");
 
         wasmtime_wasi::runtime::in_tokio(async move {
-            tokio::select! {
-                status = self.execute_component_async(ctx, component, func, stdio) => { status }
-                sig = wait_for_signal() => { sig }
+            let mut done = false;
```
Mossaka, line 309: question: what is this used for?

crates/containerd-shim-wasmtime/Cargo.toml
```diff
@@ -6,10 +6,11 @@ edition.workspace = true
 [dependencies]
 anyhow = { workspace = true }
 containerd-shim-wasm = { workspace = true, features = ["opentelemetry"] }
+libc = { workspace = true }
 log = { workspace = true }
 hyper = { workspace = true }
 tokio = { workspace = true, features = ["signal", "macros"] }
-libc = { workspace = true }
+tokio-util = { workspace = true, features = ["rt"] }
```
Mossaka, line 13: I was reading the difference between `JoinSet` and `TaskTracker` but I didn't understand if our use cases require some of the unique properties that `TaskTracker` brings. If not, I'd suggest we go with `JoinSet` since it meant one less dependency for runwasi.

Mossaka:
Please consider adding integration tests that simulate different shutdown scenarios

crates/containerd-shim-wasmtime/Cargo.toml
andreiltd: The property we leverage is

> 4. A TaskTracker can be cloned to share it with many tasks.

The tracker is cloned and passed to handler to track the tasks spawned there as well:
https://github.com/containerd/runwasi/pull/705/files#diff-b2d23b367806b354ef48f11bbfd796edae1c54af02e2a28f83fbbdf273b955b6R70

I think `JoinSet` is not `Clone` but I will double check.

crates/containerd-shim-wasmtime/src/instance.rs
andreiltd: I think we can come up with a better name for this, like `force_shutdown`. This variable tracks if we already initiated a shutdown and upon receiving a second `ctrl_c` signal, we cancel waiting for the tasks and exit immediately.

Mossaka: agreed that `force_shutdown` would be a better name

Mossaka (dismissed):

crates/containerd-shim-wasmtime/src/http_proxy.rs
```diff
@@ -59,45 +62,65 @@ pub(crate) async fn serve_conn(
...
-            {
-                log::error!("error: {e:?}");
+        tokio::select! {
+            // listen to cancellation requests
+            _ = cancel.cancelled() => {
+                break;
             }
-        });
+            res = listener.accept() => {
+                let (stream, _) = res?;
```
jprendes, line 79: **question:** What does it mean for `res` to  be `Err`?
The current behaviour triggers a forced shutdown, is that what we want?
Otherwise we can do something like
```suggestion
                let Ok(stream, _) = res else {
                    log::info!("Error accepting new connection, triggering graceful shutdown.");
                    break;
                };
```

and follow the graceful shutdown path.

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -302,9 +306,32 @@ where
...
         wasmtime_wasi::runtime::in_tokio(async move {
-            tokio::select! {
-                status = self.execute_component_async(ctx, component, func, stdio) => { status }
-                sig = wait_for_signal() => { sig }
+            let mut force_shutdown = false;
+            let exec = self.execute_component_async(ctx, component, func, stdio);
+
+            tokio::pin!(exec);
+
+            loop {
```
jprendes, line 314: **comment:** I got a bit confused by this being a loop, but then realized this will only run one, and at most two iterations. I wonder if this can be rewritten without a loop.

jprendes (commented):
Thanks for working on this @andreiltd !

crates/containerd-shim-wasmtime/src/http_proxy.rs
andreiltd: Good catch! I now changed this and followed `axum` strategy to deal with errors in accept connection.

andreiltd:
The last commit introduces a few changes:
- Added an integration test to verify forced shutdown using a terminate signal.
- Addressed errors in accepting connections by implementing a backoff strategy, which is a common pattern found in libraries such as `axum` and `mini-redis`.

crates/containerd-shim-wasmtime/src/instance.rs
andreiltd: I think the problem is that we want to continue with server future after we receive a first signal. I 'm not sure if it's possible without loop.

Mossaka: Can we do this to make sure that it at most loops twice? 
```rust
for _ in 0..2 {
  ...
}
```

Mossaka (dismissed):
Nice work, thanks for adding the integration test!

jprendes (dismissed):
LGTM

crates/containerd-shim-wasmtime/src/instance.rs
andreiltd: It is now rewritten without a loop. Thanks @jprendes for suggestion.

jprendes (approved):
Thanks @andreiltd !

PR #716, opened 2024-11-03
feat(wamr): add wasm-micro-runtime shim implementation

Mossaka:
Follow up on the work done by #508 and #642. The runtime execution `"thread signal env initialized failed"` error was resolved by the upstream Wamr SDK and thus I used the latest sdk to re-build the wamr shim. 

Now it actually worked!

```sh
sudo ctr run  --net-host --rm --runtime=io.containerd.wamr.v1 ghcr.io/containerd/runwasi/wasi-demo-app:latest testwasm /wasi-demo-app.wasm echo "hi"
hi
exiting
```

TODO
- [x] pass all the tests, see https://github.com/containerd/runwasi/commit/5030afa3475c44ad92a353db7fadf481a18d867c


crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,97 @@
...
+        log::info!("Running {func:?}");
+        let function = Function::find_export_func(&instance, &func)
+            .map_err(|e| anyhow::Error::msg(format!("Failed to find function: {:?}", e)))?;
+        let status = function
+            .call(&instance, &vec![])
+            .map(|_| 0)
+            .or_else(|err| {
+                log::error!("Error: {:?}", err);
+                Err(err)
+            })
```
Mossaka, line 92: we should parse the RuntimeError to appropriate exit code. Any thoughts on this one, @lum1n0us?

lum1n0us: Regarding that, I just noticed that WAMR only uses error strings to convey information if something goes wrong during the execution of a Wasm function. So, if it's okay, can we continue to display the error string in the log and use a simple -1 and 0 to indicate failure and success, respectively?

Mossaka: I was looking for exit code that represents https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasmer/src/tests.rs#L79-L88

Mossaka: For example, how can I get exit code 42 from running this wasm module?

```wat
(module
    ;; Import the required proc_exit WASI function which terminates the program with an exit code.
    ;; The function signature for proc_exit is:
    ;; (exit_code: i32) -> !
    (import "wasi_snapshot_preview1" "proc_exit" (func $proc_exit (param i32)))
    (memory 1)
    (export "memory" (memory 0))
    (func $main (export "_start")
        (call $proc_exit (i32.const 42))
        unreachable
    )
)

```

lum1n0us: We have [a fix](https://github.com/bytecodealliance/wamr-rust-sdk/pull/79) for this situation. Please have a look.

Mossaka: Approved

Mossaka:
@lum1n0us do you know why the entrypoint_test failed?

```console
thread 'wamr_tests::test_custom_entrypoint' panicked at crates/containerd-shim-wamr/src/tests.rs:58:5:
assertion `left == right` failed
  left: "hello world\n[00:08:01:896 - 7FA1CA910640]: warning: a module with WASI apis should be either a command or a reactor\n"
 right: "hello world\n"
```

Mossaka:
Also, it doesn't seem like it runs on Windows: https://github.com/containerd/runwasi/actions/runs/11692971119/job/32563482903?pr=716#step:8:167

crates/containerd-shim-wamr/src/instance.rs
lum1n0us: Merged

lum1n0us:
Yes, that's because WAMR enforces a strict [application ABI](https://github.com/WebAssembly/WASI/blob/main/legacy/application-abi.md) check. Every wasm32-wasi .wasm should be classified as either a command or a reactor.

For your case,
- either patch the [case](https://github.com/containerd/runwasi/blob/main/crates/containerd-shim-wasm-test-modules/src/modules/custom_entrypoint.wat) as a command, for example. Add a function like `(func $start (export "_start"))`
- or WAMR turns off the WARNING messages by setting a higher log level.

lum1n0us:
So, which option?

Mossaka:
I disabled that test for now and I don't think we need to pass it to get this PR merged. Let me fix the linting issue.

Mossaka:
@jprendes @andreiltd can you give this PR a review please 🙏

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    fn run_wasi(&self, ctx: &impl RuntimeContext, stdio: Stdio) -> Result<i32> {
+        let args = ctx.args();
+        let envs = ctx.envs();
+        let Entrypoint {
+            source,
+            func,
+            arg0: _,
+            name,
```
jprendes, line 46: **suggestion:**
```suggestion
            name,
            ..
```

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        log::info!("Create a WAMR module");
+
+        // TODO: error handling isn't ideal
+
+        let mod_name = name.unwrap_or_else(|| "main".to_string());
+
+        let mut module = Module::from_buf(&self.runtime, &wasm_bytes, &mod_name).map_err(|e| {
+            anyhow::Error::msg(format!("Failed to create module from bytes: {:?}", e))
+        })?;
```
jprendes, line 61:
```suggestion
        let mut module = Module::from_buf(&self.runtime, &wasm_bytes, &mod_name)
            .context("Failed to create module from bytes")?;
```

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,97 @@
...
+            .build();
+
+        module.set_wasi_context(wasi_ctx);
+
+        // TODO: no way to register a named module with bytes?
+
+        log::info!("Create a WAMR instance");
+
+        let instance = WamrInstnace::new(&self.runtime, &module, 1024 * 64)
+            .map_err(|e| anyhow::Error::msg(format!("Failed to create instance: {:?}", e)))?;
```
jprendes, line 78:
```suggestion
            .context("Failed to create instance")?;
```

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        log::info!("redirect stdio");
+        stdio.redirect()?;
+
+        log::info!("Running {func:?}");
+        let function = Function::find_export_func(&instance, &func)
+            .map_err(|e| anyhow::Error::msg(format!("Failed to find function: {:?}", e)))?;
```
jprendes, line 85:
```suggestion
            .context("Failed to find function")?;
```

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,97 @@
...
+        let status = function
+            .call(&instance, &vec![])
+            .map(|_| 0)
+            .map_err(|err| {
+                log::error!("Error: {:?}", err);
+                err
+            })
+            .map_err(|e| anyhow::Error::msg(format!("Failed to call function: {:?}", e)))?;
```
jprendes, line 93:
```suggestion
            .context("Failed to call function")?;
```

crates/containerd-shim-wamr/src/tests.rs
```diff
@@ -0,0 +1,117 @@
...
+        .wait(Duration::from_secs(10))?;
+
+    assert_eq!(exit_code, 0);
+
+    Ok(())
+}
+
+// test_exit_code is disabled because it requires the wamr SDK to expose exit code
+// See https://github.com/containerd/runwasi/pull/716#discussion_r1827086060
+fn test_exit_code() -> anyhow::Result<()> {
```
jprendes, line 91:
```suggestion
#[test]
#[ignore = "disabled because the WAMR SDK doesn't expose exit code yet"]
// See https://github.com/containerd/runwasi/pull/716#discussion_r1827086060
fn test_exit_code() -> anyhow::Result<()> {
```

crates/containerd-shim-wamr/src/tests.rs
```diff
@@ -0,0 +1,117 @@
...
+        .wait(Duration::from_secs(10))?;
+
+    assert_eq!(exit_code, 42);
+
+    Ok(())
+}
+
+// test_custom_entrypoint is disabled
+// See https://github.com/containerd/runwasi/pull/716#issuecomment-2458200081
+fn test_custom_entrypoint() -> anyhow::Result<()> {
```
jprendes, line 105:
```suggestion
#[test]
#[ignore]
// See https://github.com/containerd/runwasi/pull/716#issuecomment-2458200081
fn test_custom_entrypoint() -> anyhow::Result<()> {
```

jprendes (approved):
LGTM, left a few comments and suggestions

crates/containerd-shim-wamr/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition.workspace = true
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = { workspace = true }
+containerd-shim-wasm = { workspace = true }
+log = { workspace = true }
+
+[target.'cfg(unix)'.dependencies]
+wamr-rust-sdk = { git = "https://github.com/bytecodealliance/wamr-rust-sdk", tag = "v1.1.0" }
```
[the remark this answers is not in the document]
lum1n0us: Currently, there isn't a plan in place.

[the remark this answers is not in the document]
Mossaka: It would be nice to have, but this is not a blocker since we are not publishing individual shims to crates.io.

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,97 @@
...
+use wamr_rust_sdk::wasi_context::WasiCtxBuilder;
+
+pub type WamrInstance = Instance<WamrEngine>;
+
+pub struct WamrEngine {
+    runtime: Runtime,
+}
+
+unsafe impl Send for WamrEngine {}
+unsafe impl Sync for WamrEngine {}
```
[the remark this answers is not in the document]
Mossaka: I believe the `Runtime` should be made `Send` / `Sync` upstream

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -0,0 +1,97 @@
...
+        Self { runtime }
+    }
+}
+
+impl Clone for WamrEngine {
+    fn clone(&self) -> Self {
+        let runtime = Runtime::new().unwrap();
+        Self { runtime }
+    }
+}
```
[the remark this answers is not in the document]
Mossaka: We want the `Engine` to be `Clone` because we want cheap copies of the Runtime for each container in the pod. This is the case for Wasmtime, WasmEdge and Wasmer. I believe it should be the case for Wamr as well.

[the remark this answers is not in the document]
Mossaka: yeah I agree with you that we clone the entire process so there is no point to clone the engine. 

> In any case, if we create a new runtime with each clone,

at least for wasmtime, `Engine::clone()` is a cheap operation: https://docs.wasmtime.dev/api/wasmtime/struct.Engine.html#engines-and-clone

PR #723, opened 2024-11-08
Include cli implementations for adapter components

jsturtevant:
Some languages (c# and go) always include a few imports like environment and exit.  Others components that are composed with the p2 adapter will have all the imports.  This makes it so those components can run by default.

https://github.com/dotnet/runtime/issues/107405

todo:
- [x] add a test


crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -246,8 +246,10 @@ where
         // This is a adapter logic that converts wasip1 `_start` function to wasip2 `run` function.
         let status = match target {
             ComponentTarget::HttpProxy => {
+                log::info!("Found HTTP proxy target");
                 let mut linker = component::Linker::new(&self.engine);
-                wasmtime_wasi_http::add_to_linker_async(&mut linker)?;
+                wasmtime_wasi::add_to_linker_async(&mut linker)?;
+                wasmtime_wasi_http::add_only_http_to_linker_async(&mut linker)?;
```
Mossaka, line 252: nice!

Mossaka (approved):

Makefile
```diff
@@ -28,7 +28,7 @@ override CARGO = cross
 endif
 
 ifeq ($(CARGO),cross)
-override TARGET_DIR := $(or $(TARGET_DIR),./target/build/$(TARGET)/)
+override TARGET_DIR := $(or $(TARGET_DIR),./target/)
```
jprendes, line 31: **question:** Why has this changed? I think this is the cause of the CI failures.

jprendes (commented):
Thanks @jsturtevant !

Makefile
jsturtevant: I was getting errors on my local machine for the wrong build targets must be a configuration issue

jsturtevant: When I run `make test/k8s/cluster-wasmtime`

I get:

```
...
install: cannot stat './target/build/x86_64-unknown-linux-musl/x86_64-unknown-linux-musl/debug/containerd-shim-wasmtime-v1'
```

devigned (approved):
lgtm!

devigned (dismissed):
I haven't dug into the broken test runs, but besides that, lgtm.

Mossaka:
Idea: We should allow users to apply environment variables or containerd configs to mutate WASI host capabilities, similar to how the Wasmtime CLI allows flags like '-Shttp' and '-Scli'.

Not a blocker to this PR though.

Makefile
jsturtevant: I do not know why this was occurring but after cleaning up the build folders (`target` and `dist`) and doing `cargo clean`.  I no longer have the issue 🤷

jprendes (approved):
Thanks @jsturtevant !
LGTM

jsturtevant:
I had to remove the changes I made to the `makefile` completely to pass CI. I still need to clean the `dist/` folder locally or it won't update the binary after changes but due to the way CI is set up it needs the `dist/` folder in place.  It would be nice to have the tooling do the right thing locally and in CI but for now will remove those changes.

jprendes:
I feel our makefile is overly complex. 
I wonder what would be a good solution. 
How do you feel about a rust program that drives the build / test / ci workflows?

jprendes:
@jsturtevant , do you want to add tests on this PR, or on a different one?

Mossaka (dismissed):

devigned (approved):
lgtm!

Mossaka (approved):

PR #724, opened 2024-11-08
Remove special handling of the pause container

jprendes:
Since we can run native containers, there's no point in having a special case for the pause container.


Mossaka (approved):
Excited to see so much lines removed!!

Mossaka (commented):
Could you please add this change to the CHANGELOG?

Mossaka (approved):

jsturtevant:
This definitely simplifies a lot of the code. Just a note that it might add a slight perf difference here since we are doing more work (i.e. creating a container vs just returning).  We would need to measure it to know if it was significant in anyway. I would suspect it would be minor and won't be an issue in most scenarios and the complexity reduction is worth it.

jprendes:
That's true, it is more work.
It is worth mentioning that the pause container runs only once per pod, so the impact should be rather small.

PR #730, opened 2024-11-13
Upgrade wasmer version to v5.0.1

andreiltd:
Depends on https://github.com/wasmerio/wasmer/pull/5245


Cargo.toml
```diff
@@ -56,3 +56,7 @@ wasmtime-wasi-http = { version = "26.0.0" }
 
 [profile.release]
 panic = "abort"
+
+[patch.crates-io]
```
andreiltd, line 60: Remove this section before merging.

jprendes:
Now that your wasmer PR gor merged, we should try with a git dependency.
Following what we did for wamr, using a git dependency for the shim binaries should be fine, as we don't publish those crates

jprendes:
Yay, upgrading wasmer fixes the issue with bumping the rust version! @Mossaka

jprendes:
I would be happy to use the git dependency as the actual dependency (instead of being a patch), and have it merged.
That would unblock bumping rust, and the crate is not published to a registry anyway.
We can always move back to published wasmer versions when they release a new one with the fix.

Mossaka (approved):

PR #741, opened 2024-11-18
Remove sync wrappers for containerd client

jprendes:
This PR removes all the sync wrapping to the containerd client API, and embraces exposing a fully async API.
This delegates to the consumer code the responsibility of handling the async nature of the API.

A sticking point of this rewrite is that the `LeaseGuard` relied in `Drop` to release the lease. At times the code assumes that the lease would be released before continuing executing, which we can't expose asynchronously from drop. This forces us to implement a release method, and manually calling it before dropping the `LeaseGuard`.


crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -38,124 +37,117 @@ static MAX_WRITE_CHUNK_SIZE_BYTES: i64 = 1024 * 1024 * 15;
...
         unique_id: &str,
         labels: HashMap<String, String>,
     ) -> Result<WriteContent> {
         let expected = format!("sha256:{}", digest(data.clone()));
         let reference = format!("precompile-{}", unique_id);
-        let lease = self.lease(reference.clone())?;
+        let lease = self.lease(reference.clone()).await?;
 
-        let digest = self.rt.block_on(async {
+        let digest = 'digest: {
```
Mossaka, line 150: Oh interesting, I've never seen this rust usage before!!

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -177,13 +169,12 @@ impl Client {
...
             let request_stream = ReceiverStream::new(rx);
-            let request_stream =
-                with_lease!(request_stream, self.namespace, lease.lease_id.clone());
+            let request_stream = with_lease!(request_stream, self.namespace, lease.id());
             let mut response_stream = match client.write(request_stream).await {
                 Ok(response_stream) => response_stream.into_inner(),
                 Err(e) if e.code() == Code::AlreadyExists => {
                     log::info!("content already exists {}", expected.clone().to_string());
-                    return Ok(expected);
+                    break 'digest expected;
```
Mossaka, line 177: Same for this one

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -501,20 +484,22 @@ impl Client {
                 log::debug!(
                     "updating image content with precompile digest to avoid garbage collection"
                 );
-                let mut image_content = self.get_info(&image_digest)?;
+                let mut image_content = self.get_info(&image_digest).await?;
                 image_content.labels.insert(
                     format!("containerd.io/gc.ref.content.precompile.{}", i),
-                    precompiled_content.digest,
+                    precompiled_content.digest.clone(),
```
Mossaka, line 490: why is there a extra clone() here?

scripts/cleanup-tests.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
Mossaka, line 1: what's the use cases for this script?

crates/containerd-shim-wasm/src/sandbox/async_utils.rs
```diff
@@ -0,0 +1,22 @@
+#![cfg_attr(windows, allow(dead_code))] // this is currently used only for linux
+
+use std::future::Future;
+use std::sync::LazyLock;
+
+static RUNTIME: LazyLock<tokio::runtime::Runtime> = LazyLock::new(|| {
```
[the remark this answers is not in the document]
jprendes: Now that I think about it, I think it would be better to use a thread local for this.
Will update it.

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
jprendes: hehe, named blocks, a rather infrequent feature indeed.
wasm also had them, well, rather numbered than named :-)

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
jprendes: It was there so that I could do `precompiled_content.release().await` later.
without the clone `precompiled_content` would be partially destructured, and I wouldn't be able to call methods on it.

But instead of doing `precompiled_content.release()` I can do `precompiled_content.lease.release()` and save the clone, indeed :-)
Thanks!

jsturtevant (approved):
LGTM

Mossaka (approved):

PR #744, opened 2024-11-19
Add Wasmedge benchmark

dblnz:
- Enable wasmedge benchmark that runs cases that take less than 5s #626

### DIFF
--- a/benches/containerd-shim-benchmarks/benches/wasmedge-benchmarks.rs
+++ b/benches/containerd-shim-benchmarks/benches/wasmedge-benchmarks.rs
@@ -1,0 +1,154 @@
+use std::borrow::Cow;
+use std::time::Duration;
+
+use containerd_shim_wasm::container::Instance;
+use containerd_shim_wasm::sandbox::Error;
+use containerd_shim_wasm::testing::WasiTest;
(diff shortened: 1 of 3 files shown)


benches/containerd-shim-benchmarks/benches/wasmedge-benchmarks.rs
```diff
@@ -0,0 +1,154 @@
...
+        secretbox2,
+        secretbox_easy,
+        secretstream_xchacha20poly1305,
+        shorthash,
+        siphashx24,
+        stream3,
+        stream4
+}
+
+criterion_main!(benches);
```
Mossaka, line 153: Can we share this code with the wasmtime-benchmark?

Mossaka (commented):
Thanks for your contribution! It looks good to me, just added one comment

benches/containerd-shim-benchmarks/benches/wasmedge-benchmarks.rs
dblnz: I thought about that, but some of the cases do differ.
I'll try and see if I can at least define a common list.

Mossaka (approved):

Mossaka:
Thanks for your first contribution to Runwasi!! 🥂🥂

PR #745, opened 2024-11-22
Use async pid fd instead of blocking waitid to wait for a child process to exit

jprendes:
This PR replaces the blocking waitid call with an async implementation based on pid fd.
There's a race condition where containerd-shim reaps child processes.
If the child process has already been reaped, query containerd-shim to get the process status.

Note that this race condition is already present in the current implementation, but runwasi's waitid is very likely to win the race, the introduction of async evens out the odds between runwasi and containerd-shim.

This PR is in preparation to move the whole shim implementation to async.

### DIFF
--- a/crates/containerd-shim-wasm/src/sys/unix/pid_fd.rs
+++ b/crates/containerd-shim-wasm/src/sys/unix/pid_fd.rs
@@ -1,0 +1,44 @@
+use std::os::fd::{AsFd, FromRawFd as _, OwnedFd, RawFd};
+
+use libc::pid_t;
+use nix::sys::wait::{waitid, Id, WaitPidFlag, WaitStatus};
+use tokio::io::unix::AsyncFd;
+
+pub(super) struct PidFd {
+    fd: OwnedFd,
+}
+
+impl PidFd {
+    pub(super) fn new(pid: impl Into<pid_t>) -> anyhow::Result<Self> {
+        use libc::{syscall, SYS_pidfd_open, PIDFD_NONBLOCK};
+        let pidfd = unsafe { syscall(SYS_pidfd_open, pid.into(), PIDFD_NONBLOCK) };
+        if pidfd == -1 {
+            return Err(std::io::Error::last_os_error().into());
+        }
+        let fd = unsafe { OwnedFd::from_raw_fd(pidfd as RawFd) };
+        Ok(Self { fd })
+    }
+
+    pub(super) async fn wait(self) -> std::io::Result<WaitStatus> {
+        let fd = AsyncFd::new(self.fd)?;
+        loop {
+            // Check with non-blocking waitid before awaiting on fd.
+            // On some platforms, the readiness detecting mechanism relies on
+            // edge-triggered notifications.
+            // This means that we could miss a notification if the process exits
+            // before we create the AsyncFd.
+            // See https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html
+            match waitid(
+                Id::PIDFd(fd.as_fd()),
+                WaitPidFlag::WEXITED | WaitPidFlag::WNOHANG,
+            )? {
+                WaitStatus::StillAlive => {
+                    let _ = fd.readable().await?;
+                }
+                status => {
+                    return Ok(status);
+                }
+            }
+        }
+    }
+}
--- a/crates/containerd-shim-wasm/src/sys/unix/pid_fd.rs
+++ b/crates/containerd-shim-wasm/src/sys/unix/pid_fd.rs
@@ -1,22 +1,30 @@
 use std::os::fd::{AsFd, FromRawFd as _, OwnedFd, RawFd};
+use std::time::Duration;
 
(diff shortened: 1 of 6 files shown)


andreiltd (approved):
LGTM!

Mossaka (approved):
LGTM!

jprendes:
I still need to figure out why CI is failing. I think the http_poxy container is not receiving the SiGINT for some reason. But I don't see how that can relate to this. I'll keep digging tomorrow.

andreiltd (approved):

Mossaka (approved):
Great! Thanks for working on this 

What I'd like to see, perhaps as a follow up, is some unit tests for edge cases like 
1. What would happen if the pid does not exist
1. what would happen if the orignal PID is reused by a new process after the first one exists. 
1. what would happen to call `wait()` when containerd-shim has already reaped the process

crates/containerd-shim-wasm/src/sys/unix/pid_fd.rs
```diff
@@ -0,0 +1,58 @@
...
+impl PidFd {
+    pub(super) fn new(pid: impl Into<pid_t>) -> anyhow::Result<Self> {
+        use libc::{syscall, SYS_pidfd_open, PIDFD_NONBLOCK};
+        let pid = pid.into();
+        let pidfd = unsafe { syscall(SYS_pidfd_open, pid, PIDFD_NONBLOCK) };
+        if pidfd == -1 {
+            return Err(std::io::Error::last_os_error().into());
+        }
+        let fd = unsafe { OwnedFd::from_raw_fd(pidfd as RawFd) };
+        let subs = monitor_subscribe(Topic::Pid)?;
```
Mossaka, line 25: question: shouldn't we start subscription monitor **before** the syscalls?

crates/containerd-shim-wasm/src/sys/unix/pid_fd.rs
```diff
@@ -0,0 +1,58 @@
...
+                Ok(WaitStatus::StillAlive) => {
+                    let _ = fd.readable().await?;
+                }
+                Ok(status) => {
+                    return Ok(status);
+                }
+                Err(Errno::ECHILD) => {
+                    // The process has already been reaped by the containerd-shim reaper.
+                    // Get the status from there.
+                    let status = wait_pid(self.pid, self.subs);
```
Mossaka, line 51: If containerd has already reaped the process, will this be blocking the async runtime forever? 

If so, I would suggest to run this in a spawn_blocking call. I am not entirely sure how whether or not containerd-shim handles subscription and how reliable is that.

Mossaka (commented):
Great! Thanks for working on this 

What I'd like to see, perhaps as a follow up, is some unit tests for edge cases like 
1. What would happen if the pid does not exist
1. what would happen if the orignal PID is reused by a new process after the first one exists. 
1. what would happen to call `wait()` when containerd-shim has already reaped the process

Mossaka:
> Note that this race condition is already present in the current implementation, but runwasi's waitid is very likely to win the race, the introduction of async evens out the odds between runwasi and containerd-shim.

Just to clarify, in this PR, you did not resolve this issue, right? If yes, could you please raise an issue against the repo so that we can track this one.

crates/containerd-shim-wasm/src/sys/unix/pid_fd.rs
jprendes: Good point.
Normally, this should never happen.
The only way this could happen is if the child process died before we created the subscription, which would be a bug regardless.
I'll rework this to not block the runtime.

jprendes:
> Just to clarify, in this PR, you did not resolve this issue, right?

It doesn't resolve the underlying race condition, but it covers all possible outcomes, so it's not an issue for runwasi.

> could you please raise an issue against the repo so that we can track this one

To be fair, we could just use the containerd_shim reaper alone, and we wouldn't need the pidfd part.
I think that's how containerd_shim intendeds it to be used.
However, that breaks our tests, as we don't bootstrap the containerd_shim reaper in tests.
If we do bootstrap the containerd_shim reaper in tests, then we would break tests that use `process::Command`, as they would hit the same race condition.

jprendes:
> 1. What would happen if the pid does not exist

During wait, it would hit the ECHILD branch, and then wait forever in the c8d_shim reaper. Fixed it so that we wait with a timeout.

> 2. what would happen if the orignal PID is reused by a new process after the first one exists.

That's not a problem.
* PIDFD is resilient to that by using a dedicated FD instead of the PID.
* The reaper should send the correct event regardless. Assuming the reaper always wins, that would mean that the reaper would broadcast two events with the same PID, but we stop listening after the first hit.

> 3. what would happen to call wait() when containerd-shim has already reaped the process

As long as the struct is created *before* the process exits, that's not a problem. That's why we create the struct BEFORE calling `container.start()`. I added a comment explaining this in the code.

Mossaka (dismissed):
LGTM!

Mossaka (approved):

PR #746, opened 2024-11-23
Fix description of can_handle

z63d:
Match the description of can_handle in README with the actual code.
https://github.com/containerd/runwasi/blob/251575a68dd334c8de936e15c2b704634009d254/crates/containerd-shim-wasm/src/container/engine.rs#L21


jprendes (approved):
LGTM, Thanks again for contributing!

Issue #755, opened 2024-12-02
Issue with handling signals in containers


jprendes:
We currently use tokio for the communication with containerd in the shim.
Many of the shims built on top of runwasi also use tokio, for example the wasmtime shim.
With our current implementation `tokio::signal` is broken for wasm containers created by runwasi shims.

The reason for this is a bit involved.
With out current implementation, a container inherits the global state of the shim when the container is created.
Tokio's signal handling depends on some global state, and in particular uses the self pipe trick.
This means that the pipe used for handling signals will be shared by the shim and all the wasm containers spawned by it.
We currently don't do any special signal handling inside of the shim, so this doesn't affect us where there's one container per shim.
The problem appears when one shim is managing more than one container.

PR #756, opened 2024-12-02
Add test for signals

jprendes:
This PR adds test for sending signals to containers.
This is the first step towards fixing #755.

Running:
```
USE_LIBC=0 cargo test -p containerd-shim-wasm -- test::signals::test_handling_signals --exact --show-output --nocapture --ignored
```
is very likely to stall

While running:
```
USE_LIBC=1 cargo test -p containerd-shim-wasm -- test::signals::test_handling_signals --exact --show-output --nocapture --ignored
```
does succeed without problems.

The main difference is that the first one uses `tokio::signal`, while the second one uses `libc::signal`.

Commenting out the `containerd::Client` code in `crates/containerd-shim-wasm/src/sys/unix/container/instance.rs` prevents tokio from being used in the shim, and both versions of the test pass reliably.

### DIFF
--- a/crates/containerd-shim-wasm/src/lib.rs
+++ b/crates/containerd-shim-wasm/src/lib.rs
@@ -11,6 +11,12 @@ pub mod sandbox;
 pub(crate) mod sys;
 
 #[cfg(any(test, feature = "testing"))]
+/// Utilities for writing shims tests.
+/// You can use this to test your runwasi based shim.
 pub mod testing;
+
+#[cfg(test)]
+/// Tests for runwasi's containerd-shim-wasm.
+mod test;
 
 pub use containerd_shim::Config;
--- a/crates/containerd-shim-wasm/src/test/signals.rs
+++ b/crates/containerd-shim-wasm/src/test/signals.rs
@@ -1,0 +1,157 @@
+//! This module includes a test for sending signals to containers when
+//! the shim is managing two or more containers.
+//!
+//! See https://github.com/containerd/runwasi/issues/755 for context.
+//!
+//! This test is currently broken for the reasons explained in #755.
+//! Running the test will result in a failure:
+//! ```
+//! cargo test -p containerd-shim-wasm -- test::signals::test_handling_signals --exact --show-output --nocapture --ignored
+//! ```
+//!
+//! This is because the current implementation breaks `tokio::signal`.
+//! You can verify this by using `libc::signal` instead, and the test will succeed
+//! ```
+//! USE_LIBC=1 cargo test -p containerd-shim-wasm -- test::signals::test_handling_signals --exact --show-output --nocapture --ignored
+//! ```
+//!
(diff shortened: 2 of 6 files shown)


crates/containerd-shim-wasm/src/lib.rs
```diff
@@ -13,4 +13,7 @@ pub(crate) mod sys;
 #[cfg(any(test, feature = "testing"))]
 pub mod testing;
 
+#[cfg(test)]
+mod test;
+
```
Mossaka, line 18: Can you add docs to differentiate "testing" and "test" mods? I think many people would find this confusing.

crates/containerd-shim-wasm/src/test/signals.rs
```diff
@@ -0,0 +1,143 @@
...
+
+use anyhow::{bail, Result};
+use containerd_shim_wasm_test_modules::HELLO_WORLD;
+use tokio::sync::Notify;
+
+use crate::container::{Engine, Instance, RuntimeContext};
+use crate::sandbox::Stdio;
+use crate::testing::WasiTest;
+
+#[derive(Clone)]
```
Mossaka, line 15:
```suggestion
#[derive(Clone, Default)]
```

crates/containerd-shim-wasm/src/test/signals.rs
```diff
@@ -0,0 +1,143 @@
...
+pub struct SomeEngine;
+
+impl Default for SomeEngine {
+    fn default() -> Self {
+        SomeEngine
+    }
+}
```
Mossaka, line 22: remove

crates/containerd-shim-wasm/src/test/signals.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    fn run_wasi(&self, ctx: &impl RuntimeContext, stdio: Stdio) -> Result<i32> {
+        stdio.redirect()?;
+        let name = ctx.entrypoint().func;
+        tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()?
+            .block_on(async move {
+                use tokio::time::sleep;
+                let use_libc = std::env::var("USE_LIBC").unwrap_or_default();
```
Mossaka, line 52: we need to document how to run this test since it introduces a new environment variable "USE_LIBC". Do you want to add some instructions to the /docs?

Mossaka (commented):
Great to see more tests being added!

I do believe this change should be appended as a "Added" section to the CHANGELOG since it's adding a new test feature.

crates/containerd-shim-wasm/src/test/signals.rs
jprendes: I'll add a commend in the test file.
Ideally we would fix the problem, and then we can remove the `libc` implementation, and remove the `ignore` attribute from the test.

Mossaka (approved):
LGTM!

PR #763, opened 2024-12-05
[chore] minor cleanup

jprendes:
This PR reuses the container returned by `libcontainer` instead of reloading it from disk every time.

Note: some of the functions removed here were part of the public API, probably incorrectly so.

### DIFF
--- a/crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
+++ b/crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
@@ -1,5 +1,6 @@
 use std::marker::PhantomData;
-use std::path::{Path, PathBuf};
+use std::path::Path;
+use std::sync::Mutex;
 use std::thread;
 use std::time::Duration;
 
@@ -16,7 +17,7 @@ use oci_spec::image::Platform;
 
 use crate::container::Engine;
 use crate::sandbox::async_utils::AmbientRuntime as _;
-use crate::sandbox::instance_utils::{determine_rootdir, get_instance_root, instance_exists};
+use crate::sandbox::instance_utils::determine_rootdir;
 use crate::sandbox::sync::WaitableCell;
 use crate::sandbox::{
     containerd, Error as SandboxError, Instance as SandboxInstance, InstanceConfig, Stdio,
@@ -27,7 +28,7 @@ static DEFAULT_CONTAINER_ROOT_DIR: &str = "/run/containerd";
 
 pub struct Instance<E: Engine> {
     exit_code: WaitableCell<(u32, DateTime<Utc>)>,
-    rootdir: PathBuf,
+    container: Mutex<Container>,
     id: String,
     _phantom: PhantomData<E>,
 }
@@ -54,7 +55,7 @@ impl<E: Engine> SandboxInstance for Instance<E> {
                 (vec![], Platform::default())
             });
 
-        ContainerBuilder::new(id.clone(), SyscallType::Linux)
+        let container = ContainerBuilder::new(id.clone(), SyscallType::Linux)
             .with_executor(Executor::new(engine, stdio, modules, platform))
             .with_root_path(rootdir.clone())?
             .as_init(&bundle)
@@ -64,7 +65,7 @@ impl<E: Engine> SandboxInstance for Instance<E> {
         Ok(Self {
             id,
             exit_code: WaitableCell::new(),
-            rootdir,
+            container: Mutex::new(container),
             _phantom: Default::default(),
         })
     }
@@ -78,8 +79,7 @@ impl<E: Engine> SandboxInstance for Instance<E> {
         // make sure we have an exit code by the time we finish (even if there's a panic)
         let guard = self.exit_code.set_guard_with(|| (137, Utc::now()));
 
-        let container_root = get_instance_root(&self.rootdir, &self.id)?;
-        let mut container = Container::load(container_root)?;
+        let mut container = self.container.lock().expect("Poisoned mutex");
         let pid = container.pid().context("failed to get pid")?.as_raw();
 
         container.start()?;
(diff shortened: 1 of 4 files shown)


andreiltd (approved):

Mossaka (commented):
LGTM can you add this change to the CHANGELOG please?

Mossaka:
Hold on merging this PR till I finish releasing the shim binaries today!

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
```diff
@@ -78,8 +79,7 @@ impl<E: Engine> SandboxInstance for Instance<E> {
         // make sure we have an exit code by the time we finish (even if there's a panic)
         let guard = self.exit_code.set_guard_with(|| (137, Utc::now()));
 
-        let container_root = get_instance_root(&self.rootdir, &self.id)?;
-        let mut container = Container::load(container_root)?;
+        let mut container = self.container.write().unwrap();
```
Mossaka, line 82: Is there a way to remove the usage of `unwrap()`?

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
```diff
@@ -115,11 +115,8 @@ impl<E: Engine> SandboxInstance for Instance<E> {
         let signal = Signal::try_from(signal as i32).map_err(|err| {
             SandboxError::InvalidArgument(format!("invalid signal number: {}", err))
         })?;
-        let container_root = get_instance_root(&self.rootdir, &self.id)?;
-        let mut container = Container::load(container_root)
-            .with_context(|| format!("could not load state for container {}", self.id))?;
 
-        container.kill(signal, true)?;
+        self.container.write().unwrap().kill(signal, true)?;
```
Mossaka, line 119: Is there a way to remove the usage of `unwrap()`?

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
```diff
@@ -129,23 +126,7 @@ impl<E: Engine> SandboxInstance for Instance<E> {
...
-        let container_root = get_instance_root(&self.rootdir, &self.id)?;
-        match Container::load(container_root) {
-            Ok(mut container) => {
-                container.delete(true)?;
-            }
-            Err(err) => {
-                log::error!("could not find the container, skipping cleanup: {}", err);
-            }
-        }
+        self.container.write().unwrap().delete(true)?;
```
Mossaka, line 129: Is there a way to remove the usage of `unwrap()`?

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
```diff
@@ -64,7 +65,7 @@ impl<E: Engine> SandboxInstance for Instance<E> {
         Ok(Self {
             id,
             exit_code: WaitableCell::new(),
-            rootdir,
+            container: RwLock::new(container),
```
Mossaka, line 68: question: Why is a lock required for container?

Mossaka (commented):
Added some comments and I'd suggest to add this change to the `Changed` section in CHANGELOG.

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
jprendes: The unwrap is there due to rust's mutex poisoning.
Mutex poisoning happens when a thread panics while holding a lock on a mutex.
Since (in release) we build with `panic = "abort"` the unwrap should never happen.
In any case, if a thread panics, while holding this lock, I don't think we want to attempt recovering.

Do you think a comment about that in the code would be helpful?

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
jprendes: the operations we do on the container (start, kill, delete...) require mutable access, while we are only receiving an immutable reference. That's why we need the lock.

As to what we were doing before, the container object reads and write the state to disk. We were loading the container from disk. There's nothing enforcing synchronization at the filesystem level, but these operations should indeed be synchronized. With the previous version we could have had two mutable `Container` instances simultaneously trying to write different states to the same files.

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
jsturtevant: >Do you think a comment about that in the code would be helpful?

+1

Mossaka: > Do you think a comment about that in the code would be helpful?

I prefer `.expect("Lock was poisoned")` or something like this

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
jsturtevant: Is there a scenario where that would happen when this is driven by containerd?

jprendes: I don't know. @cpuguy83 , do you know the answer?

In any case, I would expect the lock to be cheaper than loading and parsing the state from disk.

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
andreiltd: ... or use `parking_lot::RwLock` instead, if you feeling fancy :)

andreiltd (approved):

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
jprendes: I think I'll go with the `expect` option to avoid adding a dependency on `parking_lot`.
Also, panic-ing in the case of a mutex poisoning seems like a better idea than just ignoring it.

andreiltd (approved):

Mossaka (approved):
LGTM!

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
Mossaka: It looks like there is a pending question here. I agree with @jprendes that this PR is a net gain for the perf, so I am happy to merge this to the main tree.

jsturtevant: Its probably faster in the startup, maybe a bit more memory consumption 🤷

jsturtevant (commented):
LGTM

PR #765, opened 2024-12-06
Makefile: fixed the install path issue

Mossaka:
If TARGET is empty, then there is a double '/'. This commit fixes this issue


jprendes (approved):
LGTM

Was this causing any issues? Or is it for aesthetics?

jsturtevant (approved):

Mossaka:
Okay, nvm, the issue was that I used `make build` and then used `sudo make install`. After I removed `sudo`, it works now.

PR #768, opened 2024-12-09
Set memory maximum size when checking pool allocation

andreiltd:
What it says on the tin.


jprendes:
Was this not working as intended without this change? Or is it something new in the new qasmtime version?

andreiltd:
I think it was not working as intended without setting memory size. :disappointed:

jprendes:
Ah, sorry, now I see the linked issue. LGTM!

jprendes (approved):

PR #774, opened 2024-12-18
Remove `Engine` from `InstanceConfig`

jprendes:
This PR removes the `Engine` generic from `InstanceConfig`.
As we move away from reusing one engine for all containers, we don't need to pass the engine around anymore.

This is a breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ pub trait Instance {
     type Engine: Send + Sync + Clone;
 
     /// Create a new instance
-    fn new(id: String, cfg: Option<&InstanceConfig<Self::E>>) -> Self;
+    fn new(id: String, cfg: &InstanceConfig) -> Self;
     /// Start the instance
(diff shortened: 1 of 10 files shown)


andreiltd (approved):

jprendes:
Converting to draft until I figure out how to deal with the regression in wasmtime.

jprendes:
There seem to be a regression where wasmtime e2e tests are timing out, so that's a first indication.
But I think it's related to the pool allocator, which is allocating loads of virtual memory.
Before we were doing that operation once in the process, while now we do it twice per container.

@jsturtevant if you know how to get the benchmark reports for PRs, please let me know :-)

jprendes:
> I don't think it will really be something that will be noticed but did we measure the chain against the bench mark at all?

The latest benchmark shows no regression (and no improvement).

Mossaka (approved):

PR #775, opened 2024-12-19
Use Zygote process

### LINKED ISSUE #755: Issue with handling signals in containers

jprendes:
This PR introduces a zygote process (a very small process cloned early on during shim startup), from which new containers are created. This effectively decouples the global state of the shim from the state of the container process.

Fixes #755 and #357

### DIFF
--- a/crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
+++ b/crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
@@ -14,6 +14,7 @@ use nix::errno::Errno;
 use nix::sys::wait::{waitid, Id as WaitID, WaitPidFlag, WaitStatus};
 use nix::unistd::Pid;
 use oci_spec::image::Platform;
+use zygote::{WireError, Zygote};
 
 use crate::container::Engine;
 use crate::sandbox::async_utils::AmbientRuntime as _;
@@ -25,7 +26,7 @@ use crate::sandbox::{
 use crate::sys::container::executor::Executor;
 use crate::sys::stdio::open;
 
-static DEFAULT_CONTAINER_ROOT_DIR: &str = "/run/containerd";
+const DEFAULT_CONTAINER_ROOT_DIR: &str = "/run/containerd";
 
 pub struct Instance<E: Engine> {
     exit_code: WaitableCell<(u32, DateTime<Utc>)>,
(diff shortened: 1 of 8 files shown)


jprendes:
This build on the changes from #774 
For a diff without the changes in #774 see https://github.com/jprendes/runwasi/compare/remove-engine..zygote

The benchmark results show no change in performance w.r.t. the scheduled nightly runs:
* https://github.com/containerd/runwasi/actions/runs/12810771961/job/35718583098?pr=775
* https://github.com/containerd/runwasi/actions/runs/12810771961/job/35718583943?pr=775#step:9:86

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
```diff
@@ -39,36 +40,52 @@ impl<E: Engine + Default> SandboxInstance for Instance<E> {
...
-        }
-        if let Ok(f) = open(cfg.get_stdout()) {
-            builder = builder.with_stdout(f);
-        }
-        if let Ok(f) = open(cfg.get_stderr()) {
-            builder = builder.with_stderr(f);
-        }
-
-        let container = builder.as_init(&bundle).with_systemd(false).build()?;
+        let (root, state) = Zygote::global()
```
Mossaka, line 52: Great, nice interface to use

Mossaka: I would like to see the memory usage of this zygote process vs. the cloned shim process running containers

andreiltd (approved):

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
jprendes: ### Memory usage as reported by `VmPeak` in `/proc/xxx/status`

Without zygote
```
Shim peak memory usage was: 3341772 kB
```

With zygote:
```
Shim peak memory usage was: 3342252 kB
Zygote peak memory usage was: 31056 kB
```

---

I am a bit surprised by the memory usage of the shim, I expected it to consume less memory.

jsturtevant (approved):
This is really great work!  Looking forward to seeing less sporadic CI failures :rocket:

Do I understand correctly that the container creation would look like:

```
->shim 
     ->zygote (cloned from shim)
          -> container 1 (message sent from shim to zygote; cloned from zygote)
          -> container 2 (message sent from shim to zygote; cloned from zygote)
```

jprendes:
Yes, that's correct.

We could still see issues if we ever "pollute" the zygote. We do create an engine inside the zygote before cloning the process, and if that has global state, starts threads, etc, it can be a problem.

Currently the engine struct for all our shims are very simple, so that's not an issue. 

I'm thinking on addressing that as part of the split engine-precompiler.

crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
jprendes: Revisiting the numbers, for wasmtime the shim peaks at ~19000 kB of resident memory with or without the zygote.
The zygote itself peaks at 5500kB of resident memory.

Mossaka (approved):
LGTM!

PR #776, opened 2024-12-19
Make `precompile` and `can_precompile` not take self

jprendes:
This PR makes it so that `precompile` and `can_precompile` do not take self.

The main intention is that ideally we would not need an engine to precompile.
If a runtime can precompile without instantiating the wasm runtime, that would be ideal.

This would allow us to move all engine creation inside of the zygote process.

This makes tests a bit more finicky, as configuring the fake engine is not as simple.

However, the non-test code changes are rather straight forward.

This is marked as draft as I'm still exploring options.

### DIFF
--- a/crates/containerd-shim-wasmtime/src/instance.rs
+++ b/crates/containerd-shim-wasmtime/src/instance.rs
@@ -1,6 +1,7 @@
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
 use std::marker::PhantomData;
+use std::sync::LazyLock;
 
 use anyhow::{bail, Context, Result};
 use containerd_shim_wasm::container::{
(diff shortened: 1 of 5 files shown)


jprendes:
@jsturtevant , what do you think of doing something like this?

My main complain is that it makes the engine precompilation kind of a "global" thing.

Maybe an "EngineBuilder" would be a better idea, and we can move the precompilation to the engine builder.

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -22,6 +23,18 @@ use crate::http_proxy::serve_conn;
 
 pub type WasmtimeInstance = Instance<WasmtimeEngine<DefaultConfig>>;
 
+static PRECOMPILER: LazyLock<wasmtime::Engine> = LazyLock::new(|| {
+    let mut config = wasmtime::Config::new();
+
+    // Disable Wasmtime parallel compilation for the tests
+    // see https://github.com/containerd/runwasi/pull/405#issuecomment-1928468714 for details
+    config.parallel_compilation(!cfg!(test));
```
fibonacci1729, line 31: Does this not run the risk of configurations becoming misaligned with how the runner engine is configured?

jprendes:
Closing in favour of a new API as proposed in #799

PR #782, opened 2024-12-28
Make WasmtimeEngine a marker struct

jprendes:


jprendes:
Wasmtime benchmark results are:
`end-to-end/wasmtime/wasi-demo-app:latest`: 131_303_173 ± 2_909_033 (last scheduled was 141_392_294 ± 3_279_112)
`end-to-end/wasmtime/wasi-demo-oci:latest`: 80_400_136 ± 1_424_339 (last scheduled was 97_305_557 ± 3_055_651)
`aead_chacha20poly13052/Wasmtime`: 121_648_447 ± 1_539_026 (last scheduled was 128_098_875 ± 1_606_609)

so it seems that performance is **not** degraded

Mossaka (commented):
question: why do we want to make the wasmtime engine a marker struct? You've mentioned this work is a preparation for the Zygote work. Could you please elaborate?

jprendes:
When we introduce the zygote, we won't be able to clone the `WasmtimeEngine` from the shim to the child process, as the code will be executed in a different process.
This means that we need to create a new `WasmtimeEngine` rather than cloning.
However, we currently need a `WasmtimeEngine` in the shim to handle precompilation.
Making it a marker struct (i.e., a struct with no fields) means that it's not only cheap to clone, but also cheap to create.
This also allows us to use slightly different `wasmtime::Engine`s for precompilation and for execution (e.g., we don't need pool allocator for precompilation).

Mossaka (approved):

Issue #504, opened 2024-03-06
Create an API for Pre-processing layers


jsturtevant:
We've added the ability to pre-compile layers in  #405 and followed up with some improvements in #492.

During the iterations it was identified that a similar API could be used to do [other pre-initialization work as well](https://github.com/containerd/runwasi/pull/492#issuecomment-1962103124):

- composing multiple layers into a single layer
- pre-compiling (as is already done but could be done in this api instead along with other processing)
- using tools like [Wizer](https://github.com/bytecodealliance/wizer?tab=readme-ov-file#using-wizer-with-a-custom-linker) to do other pre-initialization optimizations

I would like to propose we create an API similar to the precompile but would replace it with something like `process_layers` or  `initialize_layers `.   This should also allow the shim to give us [processed layers that are combined](https://github.com/containerd/runwasi/pull/492#discussion_r1506639405) and the id's of ones the want to remain untouched.

One possible way to do this would look something like the precompile layer (open to suggestions):

```rust
struct ProcessedLayers {
  layers: Vec<Layer>
  unique_cache_id: Option<String> // return None if nothing to do
}

enum Layer {
 Original(digest)  // for files we want untouched
 Processed(Vec<u8>)
}

fn precompile(&self, _layers: &[WasmLayer]) -> Result<ProcessedLayers> {
        // do work
}
```

On the shim side we could take the layers create a new file that stores these digests and link it to the original image.  Upon reloading the image we could look up the file, load those digests. 

Questions: 
- in https://github.com/containerd/runwasi/pull/492#pullrequestreview-1912475229 we add ability to re-use layers that where already processed.  How to accomplish that in this scenario?

PR #792, opened 2025-01-09
Extend precompile to support a DAG

fibonacci1729:
~Still WIP while I address some lingering TODOs but wanted to get this up for 👀 before too long.~

This PR effectively breaks the assumption that layers input to precompile map 1:1 with layers returned. This is necessary to support things like component dependencies in the spin shim via composition where there is not a clear mapping of original to precompiled layers. 

I tried to evolve the precompile API in the most straightforward way without requiring shim developers to have to manage a DAG themselves. 

~NOTE: have to update the other shims to conform to the new API which will be straightforward but i'd like to get feedback on this approach here first.~

cc/ @Mossaka @jsturtevant @kate-goldenring

### DIFF
--- a/crates/containerd-shim-wasm/src/container/engine.rs
+++ b/crates/containerd-shim-wasm/src/container/engine.rs
@@ -1,4 +1,6 @@
+use std::collections::BTreeSet;
 use std::fs::File;
+use std::future::Future;
 use std::io::Read;
 
 use anyhow::{bail, Context, Result};
@@ -60,10 +62,11 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// This is used to precompile the layers before they are run and will be called if `can_precompile` returns `true`.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
     /// The cached, precompiled layers will be reloaded on subsequent runs.
-    /// The runtime is expected to return the same number of layers passed in, if the layer cannot be precompiled it should return `None` for that layer.
-    /// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
-    fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<Vec<u8>>>> {
-        bail!("precompile not supported");
+    fn precompile(
+        &self,
+        _layers: &[WasmLayer],
+    ) -> impl Future<Output = Result<Vec<PrecompiledLayer>>> + Send {
+        async move { bail!("precompile not supported") }
     }
 
     /// Can_precompile lets the shim know if the runtime supports precompilation.
@@ -80,4 +83,15 @@ pub trait Engine: Clone + Send + Sync + 'static {
     fn can_precompile(&self) -> Option<String> {
         None
     }
+}
+
+/// A `PrecompiledLayer` represents the precompiled bytes of a layer and the digests of parent layers (if any) used to process it.
+#[derive(Clone)]
+pub struct PrecompiledLayer {
+    /// The media type this layer represents.
+    pub media_type: String,
+    /// The bytes of the precompiled layer.
+    pub bytes: Vec<u8>,
+    /// Digests of this layers' parents.
+    pub parents: BTreeSet<String>,
 }
--- a/crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashMap;
 use std::path::Path;
+use std::str::FromStr;
 
 use containerd_client;
 use containerd_client::services::v1::containers_client::ContainersClient;
@@ -17,14 +18,14 @@ use containerd_client::tonic::transport::Channel;
 use containerd_client::tonic::Streaming;
 use containerd_client::{tonic, with_namespace};
 use futures::TryStreamExt;
-use oci_spec::image::{Arch, Digest, ImageManifest, MediaType, Platform};
+use oci_spec::image::{Arch, DescriptorBuilder, Digest, ImageManifest, MediaType, Platform};
 use sha256::digest;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::{Code, Request};
 
 use super::lease::LeaseGuard;
-use crate::container::Engine;
+use crate::container::{Engine, PrecompiledLayer};
 use crate::sandbox::error::{Error as ShimError, Result};
 use crate::sandbox::oci::{self, WasmLayer};
 use crate::with_lease;
@@ -406,18 +407,21 @@ impl Client {
             .iter()
             .filter(|x| is_wasm_layer(x.media_type(), T::supported_layers_types()));
 
-        let mut layers = vec![];
+        let mut all_layers = HashMap::new();
+        let media_type_label = precompile_label(T::name(), "media-type");
         for original_config in configs {
-            let layer = self
-                .read_wasm_layer(
-                    original_config,
-                    can_precompile,
-                    &precompile_id,
-                    &mut needs_precompile,
-                )
-                .await?;
-            layers.push(layer);
+            self.read_wasm_layer(
+                original_config,
+                can_precompile,
+                &precompile_id,
+                &mut needs_precompile,
+                &media_type_label,
+                &mut all_layers,
+            )
+            .await?;
         }
+
+        let layers = all_layers.into_values().collect::<Vec<_>>();
 
         if layers.is_empty() {
             log::info!("no WASM layers found in OCI image");
@@ -426,12 +430,11 @@ impl Client {
 
         if needs_precompile {
             log::info!("precompiling layers for image: {}", container.image);
-            let compiled_layers = match engine.precompile(&layers) {
+            let compiled_layers = match engine.precompile(&layers).await {
                 Ok(compiled_layers) => {
-                    if compiled_layers.len() != layers.len() {
-                        return Err(ShimError::FailedPrecondition(
-                            "precompile returned wrong number of layers".to_string(),
-                        ));
+                    if compiled_layers.is_empty() {
+                        log::info!("no precompiled layers returned");
+                        return Ok((layers, platform));
                     }
                     compiled_layers
                 }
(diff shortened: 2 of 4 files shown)


jsturtevant:
/cc @jprendes as well who is doing some larger changes to the engine 

linking a few issues for context: 
https://github.com/containerd/runwasi/issues/504
https://github.com/spinkube/containerd-shim-spin/issues/238

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -62,7 +65,7 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// The cached, precompiled layers will be reloaded on subsequent runs.
     /// The runtime is expected to return the same number of layers passed in, if the layer cannot be precompiled it should return `None` for that layer.
     /// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
-    fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<Vec<u8>>>> {
+    async fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<PrecompiledLayer>> {
```
jsturtevant, line 68: Since this now provides a way to do more than just pre-compiling such as composing I wonder if we would want to rename it

fibonacci1729: Yeah that definitely seems reasonable! Happy to call this `process_layers` or whatever makes sense.

jsturtevant:
a quick look and it is about what I was expecting for changes.  It was helpful to see how it was used in a non trivial example with https://github.com/spinkube/containerd-shim-spin/pull/259

jprendes (commented):
I have some concerns around the GC references, and how we link layers with precompiled artifacts, but I'm no expect on any of that, so feel free to correct me.

I would eventually like to split the precompilation from the `Engine`, which would be a significant breaking change. If you think there's a more drastic change that could help with this work, I'd like to hear it :-)

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -515,44 +537,82 @@ impl Client {
...
+        all_layers: &mut HashMap<Digest, WasmLayer>,
+    ) -> std::prelude::v1::Result<(), ShimError> {
+        let parent_digest = original_config.digest().clone();
+        let digests_to_load = if can_precompile {
+            let info = self.get_info(&parent_digest).await?;
+            let child_digests = info
+                .labels
+                .into_iter()
+                .filter_map(|(key, child_digest)| {
+                    if key.starts_with(&format!("{precompile_id}/child")) {
```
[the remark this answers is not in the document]
fibonacci1729: IIUC only the precompiled artifacts that are reachable via the child-label from the rooted image will be loaded here.

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -442,40 +445,46 @@ impl Client {
...
+                    );
+
+                    let parent_label = format!("{precompile_id}/child.{child_digest}");
+                    parent_layer
+                        .labels
+                        .insert(parent_label, child_digest.clone());
+
+                    let gc_label =
+                        format!("containerd.io/gc.ref.content.precompile.{child_digest}");
+                    parent_layer.labels.insert(gc_label, child_digest.clone());
```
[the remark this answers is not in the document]
jsturtevant: This is a good call out that I hadn't considered and wasn't an issue when we had 1to1 mapping.  

We do have to put the reference on the content that we are compiling since the[ Image reference is mutable](https://github.com/containerd/runwasi/pull/492#issuecomment-1965541517).

I don't know the best way to handle this.  Maybe we could use a subscription to containerd image events to remove labels when an image is removed?  

@cpuguy83 do you have any ideas on how we might be able to remove labels to avoid stale content?

fibonacci1729:
Opening up for review after addressing some of the initial feedback. Pending some tests and resolution to some convos about gc / labeling scheme employed here, i think this is ready to go!

crates/containerd-shim-wasm/src/container/engine.rs
fibonacci1729: Should `can_precompile()` evolve as well? It seems a little strange to rename one and not the other. Additionally there are various labels applied with a `*.precompile` segment; should we rename those to `process` as well to maintain consistency?

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -515,44 +539,82 @@ impl Client {
...
-                log::error!("falling back to original layer and marking for recompile");
-                *needs_precompile = can_precompile; // only mark for recompile if engine is capable
-                self.read_content(original_config.digest())
-                    .await
-                    .map(|module| WasmLayer {
-                        config: original_config.clone(),
-                        layer: module,
-                    })
+        } else {
+            vec![parent_digest.clone()]
```
kate-goldenring, line 570: nit: unnecessary clone
```suggestion
            vec![parent_digest]
```

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -155,31 +157,46 @@ impl Engine for WasmtimeEngine {
...
+                let parent_digest = layer.config.digest().to_string();
+
+                compiled_layers.push(PrecompiledLayer {
+                    media_type: layer.config.media_type().to_string(),
+                    bytes: compiled_layer,
+                    parents: {
+                        let mut parents = BTreeSet::new();
+                        parents.insert(parent_digest);
+                        parents
+                    },
```
kate-goldenring, line 194:
```suggestion
                     parents: BTreeSet::from([parent_digest]),)
```

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
```diff
@@ -406,32 +407,34 @@ impl Client {
             .iter()
             .filter(|x| is_wasm_layer(x.media_type(), T::supported_layers_types()));
 
-        let mut layers = vec![];
+        let mut all_layers = HashMap::new();
+        let media_type_label = precompile_label(T::name(), "media-type");
```
kate-goldenring, line 411: Where is this `"media-type"` string coming from or how was it determined?

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -155,31 +157,46 @@ impl Engine for WasmtimeEngine {
...
+                let compiled_layer = match WasmBinaryType::from_bytes(&layer.layer) {
+                    Some(Module) => PRECOMPILER.precompile_module(&layer.layer)?,
+                    Some(Component) => PRECOMPILER.precompile_component(&layer.layer)?,
+                    None => {
+                        log::warn!("Unknow WASM binary type");
+                        continue;
+                    }
+                };
+
+                let parent_digest = layer.config.digest().to_string();
```
kate-goldenring, line 185: is parents expected to never be empty?

kate-goldenring (commented):
I understood this flow at one point and once again am a little lost in the weeds of the DAG.  @fibonacci1729  Could you add some more comments to describe the functions' objectives and consider breaking the large functions out into smaller ones for better readability/understanding?

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
fibonacci1729: Each `PrecompiledLayer` returned from `precompile` specifies the media-type of the new layer's bytes so that later when reading each child wasm layer we can synthesize a `Descriptor`.

crates/containerd-shim-wasmtime/src/instance.rs
fibonacci1729: At the moment, i think that should be the assumption. Otherwise an empty parents set would imply a new layer that has no relation to the other layers in the image. I'll add a check in the loop above to ensure this precondition for now.

jprendes:
The test in `client.rs` L.1074 needs to be updated to match the new method signature.
The method behavior probably needs to be updated to match the new expected behavior.

crates/containerd-shim-wasmtime/src/instance.rs
kate-goldenring: Can you also specify this in the inline doc comments on the `PrecompiledLayer` structure? Ill leave a comment there too

fibonacci1729: Oh yeah, good catch. Will update the comment. I'd be wary of the `self if none` approach as that would introduce a cycle we'd have to manage when loading.

fibonacci1729:
@jprendes I'm developing using lima on macos and when i run the tests via (which is analogous to what the `make` targets are doing):

```
cargo test --target-dir /tmp/lima/target --package containerd-shim-wasm --lib -- sandbox::containerd::client::tests --show-output
```

I see this error for each of the precompile tests:
```
called `Result::unwrap()` on an `Err` value:  failed to import image
```

Which happens when `oci_helpers::import_image` is called from `generate_test_container`.

FWIW, i'm also running `containerd` along side the tests.

Also, ran the tests as root but same result.

Any insight into what i might be goofing up?

jprendes:
I am not super familiar with lima, but the tests using the OCI image do invoke `ctr`. Maybe you need to add the lima wrapper there?
Are you building and running inside lima, or using lima just for `ctr` / `c8d`?
You also need to load the OCI image in c8d beforehand, IIRC it was with `make load/oci`.

jsturtevant:
lima ships with nerdctl which might need some updates to work with the [oci images](https://github.com/containerd/nerdctl/issues/1472).  Does lima also install ctr, if not you can get it from containerd's releases and should work.

fibonacci1729:
Thanks! I abandoned lima and switch over to setting up an azure Linux vm and got the tests working. Working to update the tests now to reflect the api changes.

fibonacci1729:
Updated the original tests; would appreciate a discerning eye to ensure i'm not horribly breaking any assumptions.

Working to add some new tests.

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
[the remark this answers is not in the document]
cpuguy83: Can you explain what's happening here?

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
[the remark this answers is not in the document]
fibonacci1729: @jprendes Great catch; it turns out that this is the case and is highly undesirable. Unfortunately, i think we'll need to rethink the labeling scheme here..

crates/containerd-shim-wasm/src/container/engine.rs
```diff
@@ -60,10 +62,11 @@ pub trait Engine: Clone + Send + Sync + 'static {
     /// This is used to precompile the layers before they are run and will be called if `can_precompile` returns `true`.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.  
     /// The cached, precompiled layers will be reloaded on subsequent runs.
-    /// The runtime is expected to return the same number of layers passed in, if the layer cannot be precompiled it should return `None` for that layer.
-    /// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
-    fn precompile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<Vec<u8>>>> {
-        bail!("precompile not supported");
+    fn precompile(
```
Mossaka, line 65: question: why don't we define this fucntion to be async?

jprendes: Using async function in public traits is discouraged.
The reason is that usually you want your future to be Send so that you can execute it in a multithreaded runtime.
The async fn syntax doesn't enforce the Send bound, so the suggested approach is to desugar the async fn into an impl Future + Send.

fibonacci1729: Mostly to avoid bringing in the `async_trait` dependency.

crates/containerd-shim-wasm/src/sandbox/containerd/client.rs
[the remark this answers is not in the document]
jsturtevant: After our community meeting was thinking on this a bit, instead of applying labels to try and track the state, could we create new content blob with the correct mappings? The content could then would tied to the Lifecyle of the image digest, and the pre-compiled components tied to the new blob. avoiding having labels propagate through out the rest of the content store.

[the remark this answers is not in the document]
fibonacci1729: @jsturtevant which mappings are you thinking about storing?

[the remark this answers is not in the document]
jsturtevant: essentially the one returned from the preprocess/precompile call,  Instead of distributing that information via labels to the affected layers, we could create a new content with the mappings.  When the digest is called, it could look those up and gather the correct pre-compiled layers.

fibonacci1729:
Closing this since given [spin#2996](https://github.com/spinframework/spin/pull/2996). By default Spin now pushes components composed with their dependencies (specified in the spin.toml) which obviates the need to address composition at the runwasi level.

jsturtevant:
Thanks for the update!  We can revisit in the future if we get different requirements

Issue #787, opened 2025-01-01
Could not jaeger containerd-shim-wasm


giteewif:
I follow the readme.md ,and setup-otel.sh in scripts, and ctr run  wasi-demo-app, but could only get the containerd tracing, there is no containerd-shim-wasm tracing info?

Mossaka:
Have you followed this document about enabling OpenTelmetry in the shim? https://github.com/containerd/runwasi/blob/main/docs/opentelemetry.md

giteewif:
> Have you followed this document about enabling OpenTelmetry in the shim? https://github.com/containerd/runwasi/blob/main/docs/opentelemetry.md

Yes I have followed the document, i could get a part of the tracing info, but not the shim_main. It seems like that the tracing info consists of that after the create_task_service. There is no tracing func like shim_main or start_shim.
![Image](https://github.com/user-attachments/assets/385030b8-f859-4ef9-a8f9-0842a41674a0)

Mossaka:
yea we renamed `shim_main_with_otel` to `shim_main()`. I will update the doc to reflect this change, and thanks for pointing it out!

> Yes I have followed the document, i could get a part of the tracing info, but not the shim_main. 

I will look into this issue but we should be able to see `shim_main` and `start_shim `

giteewif:
> yea we renamed `shim_main_with_otel` to `shim_main()`. I will update the doc to reflect this change, and thanks for pointing it out!
> 
> > Yes I have followed the document, i could get a part of the tracing info, but not the shim_main.
> 
> I will look into this issue but we should be able to see `shim_main` and `start_shim `

there is also the validate or exec tracing info not see.

Mossaka:
There is something broken in the main branch. I was able to trace to a particular commit [f38990d4da42e4de6f3c363998dafec253c33a4f](https://github.com/containerd/runwasi/commit/f38990d4da42e4de6f3c363998dafec253c33a4f)

Before that commit I was able to see the full tracing info. e.x.

<img width="1152" alt="Image" src="https://github.com/user-attachments/assets/b2dc71b3-b036-4dbc-a74e-3522af89f214" />

giteewif:
Yes，I do use the latest in the main branch.

Mossaka:
@giteewif try it out now on the main branch

PR #794, opened 2025-01-09
.github/workflows/action-test-smoke: Add a step to verify the jeager

Mossaka:
Add a GitHub Action step to verify that Jeager actually receives the expected emitted traces, rather than just hoping they made it.

It uses Jeager API to fetch the traces, and note that the limit=0 so that we can get all the operations.

This is to prevent #787 from happening without getting notices.


jprendes:
@Mossaka I think this one needs a rebase

jprendes (approved):

PR #810, opened 2025-01-23
feat(ci): add benchmark results as a comment to the PR for each run

Mossaka:


jprendes:
I wonder why the memory benchmark results are missing

jprendes:
That's odd, now nothing got commented?

Mossaka:
okay it looks like the comment feature is broken :(

Mossaka:
Related: https://github.com/benchmark-action/github-action-benchmark/issues/250

PR #816, opened 2025-01-28
feat(shims): enable opentelemetry for shims

Mossaka:
Currently, only the wasmtime shim enables the otel tracing. This PR enables the otel tracing for all shims.


jprendes (approved):
LGTM

PR #817, opened 2025-01-28
feat(ci): add stress test to CI

Mossaka:
This commit adds a stress test to CI with 100 tasks and either # of cores or 10 parallelism

### DIFF
--- a/.github/workflows/action-build.yml
+++ b/.github/workflows/action-build.yml
@@ -55,6 +55,9 @@ jobs:
         run: echo "OPT_PROFILE=release" >> ${GITHUB_ENV}
       - name: Build
         run: make build-${{ inputs.runtime }}
+      - name: Install binaries
+        if: ${{ inputs.runtime != 'common' }}
+        run: make dist-${{ inputs.runtime }}
       - name: Check if crate is ready to publish
         if: ${{ inputs.runtime == 'common' }}
         shell: bash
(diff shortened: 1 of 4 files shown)


.github/workflows/action-build.yml
```diff
@@ -58,12 +58,24 @@ jobs:
...
         run: |
           make test-${{ inputs.runtime }}
         if: ${{ inputs.arch == 'x86_64' }}
+      - name: Install binaries
+        if: ${{ inputs.runtime != 'common' }}
+        shell: bash
+        run: |  
+          make dist-${{ inputs.runtime }}
+      - name: Run stress tests
+        if: ${{ inputs.runtime != 'common' && inputs.binprefix == 'containerd-shim-' }}
```
jprendes, line 74:
```suggestion
        if: ${{ inputs.runtime != 'common' && inputs.binprefix == 'containerd-shim-' && inputs.arch == 'x86_64' && runner.os == 'Linux' }}
```

Mossaka:
``` /home/runner/work/runwasi/runwasi/dist/bin/containerd-shim-wamr-v1 --count=100 --parallel=4 --timeout=2s`
Error: Scheme "" is not supported
```
I wonder where this error message is from

jprendes:
I think it's from the address provided to the shim, iirc it should be "unix://path/to/socket", and scheme refers to that "unix://", but I don't know why that would be the case

jprendes:
Finally figured it out.
It's because in CI we use cross, and `cross run` executes things in a qemu environment, which has some unexpected behaviours.
Changing to `cargo run` should fix the issue.

PR #820, opened 2025-01-28
feat(*): use published images instead of locally built ones

Mossaka:
- Replaced `make load` with `sudo make pull`
- Updated documents to pull images from `ghcr.io` instead of building them locally

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,16 +220,16 @@ pull: pull-app pull-oci pull-oci-artifact pull-http
 	echo "Pulled all images"
 
 pull-app:
-	ctr image pull ghcr.io/containerd/runwasi/wasi-demo-app:latest
+	sudo ctr image pull ghcr.io/containerd/runwasi/wasi-demo-app:latest
 
 pull-oci:
-	ctr image pull ghcr.io/containerd/runwasi/wasi-demo-oci:latest
+	sudo ctr image pull ghcr.io/containerd/runwasi/wasi-demo-oci:latest
 
 pull-oci-artifact:
-	ctr image pull ghcr.io/containerd/runwasi/wasi-demo-oci-artifact:latest
+	sudo ctr image pull ghcr.io/containerd/runwasi/wasi-demo-oci-artifact:latest
 
 pull-http:
-	ctr image pull ghcr.io/containerd/runwasi/wasi-http:latest
+	sudo ctr image pull ghcr.io/containerd/runwasi/wasi-http:latest
 
 docker/load: dist/img.tar
 	docker load -i $<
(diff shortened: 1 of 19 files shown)


Makefile
```diff
@@ -215,27 +215,30 @@ dist/img-oci-artifact.tar: target/wasm32-wasip1/$(OPT_PROFILE)/img-oci-artifact.
...
+	ctr image pull ghcr.io/containerd/runwasi/wasi-demo-app:latest
+
+pull-oci:
+	ctr image pull ghcr.io/containerd/runwasi/wasi-demo-oci:latest
+
+pull-oci-artifact:
+	ctr image pull ghcr.io/containerd/runwasi/wasi-demo-oci-artifact:latest
+
+pull-http:
+	ctr image pull ghcr.io/containerd/runwasi/wasi-http:latest
```
jprendes, line 232:
```suggestion
pull-%:
	ctr image pull ghcr.io/containerd/runwasi/wasi-demo-$*:latest
```

Makefile
```diff
@@ -215,27 +215,30 @@ dist/img-oci-artifact.tar: target/wasm32-wasip1/$(OPT_PROFILE)/img-oci-artifact.
...
 	cp "$<" "$@"
 
-load: dist/img.tar
-	sudo ctr -n $(CONTAINERD_NAMESPACE) image import --all-platforms $<
+.PHONY: pull pull-app pull-oci pull-oci-artifact pull-http
+pull: pull-app pull-oci pull-oci-artifact pull-http
+	echo "Pulled all images"
+
+pull-app:
+	ctr image pull ghcr.io/containerd/runwasi/wasi-demo-app:latest
```
jprendes, line 223: For load, we do `make load`, and the makefile does `sudo ctr ..`.
This one seems to be the other way around, `sudo make pull`, while the makefile does just `ctr ..`.
Is there any reason for that? Otherwise I think I prefer the way `load` does it.

jprendes (dismissed):
LGTM!
This is great @Mossaka , thanks!

Makefile
jprendes: ah, this would break wasi-http, as it would translate to wasi-demo-http

Mossaka: Perhaps we should just call it `wasi-demo-http` to be consistent with all other images we publish

Makefile
Mossaka: I agree. I will change it to `sudo ctr` in the Makefile

Makefile
Mossaka: This could be a follow up as I need to republish the images.

Mossaka: just created an issue to track: https://github.com/containerd/runwasi/issues/825

jprendes (approved):
LGTM

PR #822, opened 2025-01-29
feat(otel): modify the span naming logic

Mossaka:
- Modify the span name from funcion name to the module path + function name

image

### DIFF
--- a/scripts/verify-jaeger-traces.sh
+++ b/scripts/verify-jaeger-traces.sh
@@ -4,7 +4,17 @@ set -euo pipefail
 TRACE_DATA=$(curl -s "http://localhost:16686/api/traces?service=containerd&limit=0" \
     | jq '[ .data[].spans[].operationName ]')
 
-REQUIRED_OPS=("task_create" "task_wait" "task_start" "task_delete" "task_state" "wait" "shutdown" "shim_main_inner")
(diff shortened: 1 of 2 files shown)


scripts/verify-jaeger-traces.sh
```diff
@@ -4,7 +4,17 @@ set -euo pipefail
 TRACE_DATA=$(curl -s "http://localhost:16686/api/traces?service=containerd&limit=0" \
     | jq '[ .data[].spans[].operationName ]')
 
-REQUIRED_OPS=("task_create" "task_wait" "task_start" "task_delete" "task_state" "wait" "shutdown" "shim_main_inner")
+REQUIRED_OPS=("task_create" "task_wait" "task_start" "task_delete" "task_state")
```
jprendes, line 7:
```suggestion
PREFIX="containerd_shim_wasm::sandbox::shim::local"
REQUIRED_OPS=("$PREFIX::task_create" "$PREFIX::task_wait" "$PREFIX::task_start" "$PREFIX::task_delete" "$PREFIX::task_state")
```
Then you can combine it with wait and shutdown like before

jprendes (dismissed):

jprendes (approved):

PR #826, opened 2025-01-29
fix(ci): fix http latency benchmark on PR

Mossaka:
The HTTP latency benchmark on PR is not working as expected. It gives a warning saying ./cache/latency_results.json is not found. This commit adds a step to download the cache file from previous run as an attempt to fix the issue.


jprendes (approved):
LGTM.
Do you know why the file is sometimes not present?

Mossaka:
image
It seems like the HTTP latency is still missing in the summary page :(

PR #828, opened 2025-01-30
Use a dedicated zygote process per container

jprendes:
This PR creates a dedicated zygote process per container.
This fixes issues with libcontainer changing the shim process global state (like chdir), which leads to race conditions on multithreaded programs.
This also helps with throughput, as the global zygote process was a bottleneck (since it's single threaded).

TODO: Add tests

### DIFF
--- a/crates/containerd-shim-wasm/CHANGELOG.md
+++ b/crates/containerd-shim-wasm/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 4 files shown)


Mossaka:
Great performance increase! I just ran a stress test on 1000 containers and with concurrency=32

**Before**
999 tasks succeeded, 0 tasks failed, 1 tasks didn't finish
elapsed time: 5s 898ms 226us 35ns
 
**After**
1000 tasks succeeded, 0 tasks failed, 0 tasks didn't finish
elapsed time: 2s 48ms 936us 565ns

Mossaka:
Strangely, the labeler does not work for this PR. 

This is the error msg: 
```
An unknown config option was under C-containerd-shim-wasm: any-glob-to-any-file
```

Mossaka (dismissed):

Mossaka (commented):
I think this change is worth a CHANGELOG mentioning how the perf is now different.

PR #850, opened 2025-02-13
fix: Corrected description of containerd_address from GRPC to ttrpc

z63d:
This is probably a mistake.

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/instance.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/instance.rs
@@ -22,7 +22,7 @@ pub struct InstanceConfig {
     bundle: PathBuf,
     /// Namespace for containerd
     namespace: String,
-    /// GRPC address back to main containerd
+    /// ttrpc address back to main containerd
     containerd_address: String,
 }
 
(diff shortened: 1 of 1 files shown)


Mossaka (dismissed):

crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -22,7 +22,7 @@ pub struct InstanceConfig {
     bundle: PathBuf,
     /// Namespace for containerd
     namespace: String,
-    // /// GRPC address back to main containerd
+    /// GRPC address back to main containerd
```
Mossaka, line 25: technically this should be a ttrpc address

Mossaka (approved):

crates/containerd-shim-wasm/src/sandbox/instance.rs
jprendes: Hmm, I think this is indeed the grpc address. The one that the wasm-oci code uses to talk back with c8d.
IIRC, this is /var/containerd/containerd.sock, while the ttrpc socket is /var/containerd/containerd.sock.ttrpc

Mossaka: Ah yes you're right! I merged this PR in but we can revert it back

PR #853, opened 2025-02-14
tracing: add parameters info to the trace instead of skipping them

Mossaka:
this commit also removes 'parent = ...' since it's already in the default

---

image

Notice how `version`, `revision`, and `name` are now included in the tags.


jprendes:
What is, if any, the impact on performance?

Mossaka:
> What is, if any, the impact on performance?

From the benchmarks, I can hardly see much negative impact on performance: https://github.com/containerd/runwasi/actions/runs/13321128097?pr=853

jprendes (approved):
LGTM

Issue #771, opened 2024-12-12
Benchmark the shim performance on wasi-demo-app workloads with mocked Containerd


Mossaka:

Mossaka:
Closing this one as completed by #855

Issue #615, opened 2024-06-07
try running hunderds of containers at once to see how "dense" containers on a node would perform


Mossaka:
- [x] #734 
- [ ] #839

Mossaka:
Closing this one as completed

PR #855, opened 2025-02-14
Add stress-test to the benchmark CI

### LINKED ISSUE #614: Benchmark the start up time for each shim
Currently, we benchmark when the task_start is invoked to the end of the execution of the wasm module.

```rust
fn run_wasmtime_test_with_spec(wasmbytes: &[u8]) -> Result<u32, Error> {
    let (exit_code, _, _) = WasiTest::<WasmtimeTestInstance>::builder()?
        .with_wasm(wasmbytes)?
        .build()?
        .start()?
        .wait(Duration::from_secs(10))?;
    Ok(exit_code)
}
```

We would love to benchmark the start up of the shim, namely from the entrypoint of `shim_main()` to `run_wasi()`.

Mossaka:
- stress-test: add a new json_output flag
- ci: add stress-test to the benchmark CI
- [ci: add stress-test with mock service to the CI](https://github.com/containerd/runwasi/pull/855/commits/f4eb1db0aa3110d72cbf4681ec4f72f56762a3b6)

Closes #614 #615 #771


crates/stress-test/src/main.rs
```diff
@@ -222,11 +241,65 @@ async fn run_stress_test(cli: Cli, c8d: impl Containerd) -> Result<()> {
...
+                    image, count, parallel, duration
+                )),
+            },
+            BenchmarkResult {
+                name: format!(
+                    "Stress Test Tasks Elapsed Time with {} service - {}",
+                    containerd_shim, shim
+                ),
+                unit: "s".to_string(),
+                value: elapsed.as_secs_f64(),
```
jprendes, line 280: This is passed to the benchmark tool in CI as `customBiggerIsBetter`, but this is not the case. 
Personally I wouldn't bother capturing this value as it's implicitly captured in the throughput.

crates/stress-test/src/main.rs
```diff
@@ -222,11 +241,65 @@ async fn run_stress_test(cli: Cli, c8d: impl Containerd) -> Result<()> {
...
+                    image, count, parallel, duration
+                )),
+            },
+            BenchmarkResult {
+                name: format!(
+                    "Stress Test Tasks Setup Time with {} service - {}",
+                    containerd_shim, shim
+                ),
+                unit: "s".to_string(),
+                value: setup_start.elapsed().as_secs_f64(),
```
jprendes, line 292: Same with this. I wouldn't capture this as it's more of an artifact of how the stress-test is implemented.

jprendes: Maybe both setup time and total elapsed time can be added to the notes of the throughput. What do you think?

crates/stress-test/src/main.rs
```diff
@@ -222,11 +241,65 @@ async fn run_stress_test(cli: Cli, c8d: impl Containerd) -> Result<()> {
...
-    let elapsed = format_duration(elapsed);
+    let duration = format_duration(elapsed);
 
     println!("\x1b[32m{success} tasks succeeded\x1b[0m");
-    println!("\x1b[32m  elapsed time: {elapsed}\x1b[0m");
+    println!("\x1b[32m  elapsed time: {duration}\x1b[0m");
     println!("\x1b[32m  throuput: {throuput} tasks/s\x1b[0m");
 
+    let re = Regex::new(r"containerd-shim-(\w+)").unwrap();
+    let shim_path_str = shim_path.to_string_lossy();
```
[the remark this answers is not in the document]
Mossaka: Thanks!

crates/stress-test/src/main.rs
```diff
@@ -222,11 +250,32 @@ async fn run_stress_test(cli: Cli, c8d: impl Containerd) -> Result<()> {
...
+                containerd_shim, shim
+            ),
+            unit: "tasks/s".to_string(),
+            value: throuput,
+            extra: Some(format!(
+                "Image: {}\nTasks: {}\nParallel: {}\nDuration: {}",
+                image, count, parallel, duration
+            )),
+        }];
+        let json = serde_json::to_string_pretty(&results)?;
```
jprendes, line 275:
```suggestion
        let file = File::create(json_output)?;
        serde_json::to_writer_pretty(file, &results)?;
```

jprendes (approved):
LGTM

jprendes (dismissed):

jprendes:
The results are a bit surprising:
* Wamr: 95 tasks/s
* Wasmedge: 85 tasks/s
* Wasmer: 17 tasks/s
* Wasmtime: 16 tasks/s

The two C/C++ runtimes seem to be much faster than the two Rust runtimes.
At first I though we might be building the shim in debug mode, but IIUC the logs indicate that's not the case. I wonder if there's something else going on here.

I think we should merge this PR and investigate these results separately.

PR #856, opened 2025-02-15
Initial work to add a mdbook documentation style

dblnz:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,43 @@
...
+          mdbook-version: "latest"
+      - name: Build mdbook
+        working-directory: ./docs
+        run: mdbook build
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        if: ${{ github.ref == 'refs/heads/main' }}
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./docs/book
```
Mossaka, line 42: Does this build the docs on `https://containerd.github.io/runwasi/docs/book`?

dblnz: No, from what I understand, what it does is:
- it generates the `website` at that path
- takes the files and commits them to `gh-pages` branch which can be selected as source for a Github Page.

The problem is it will use the same URL: https://containerd.github.io/runwasi
So, we need to somehow merge the benchmarks with the book also

Mossaka: @devigned FYI - can we use the `runwasi.dev` website?

Mossaka (approved):
LGTM!

jprendes (approved):
LGTM

PR #861, opened 2025-02-18
ci: Add oci and oci-artifact images to stress test

Mossaka:
In addition, this commit adds a release flag to stress test CLI in Makefile.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -127,25 +127,26 @@ jobs:
       - uses: ./.github/actions/setup-env
       - name: Setup build profile
         shell: bash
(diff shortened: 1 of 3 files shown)


jprendes (approved):
LGTM

jprendes:
Oh, sorry, I approved this and didn't realise this conflicted with a change i made to how we compute the default entrypoint for stress-test.

This means that this is running with the image default arguments, which runs a wasm program in an infinite loop, and the stress-test times out.

Right now the stress test has a special case for wasi-demo-app to use the "echo hello" arguments as the default.

I wonder if we should just make that the default behavior of the demo app and remove the special case from stress-test.

jprendes:
I will revert this PR to unblock CI.

Issue #821, opened 2025-01-29
Pod level metrics are not being surfaced


kate-goldenring:
This is a duplicate of this issue found in the Spin shim: https://github.com/spinkube/containerd-shim-spin/issues/180. I wanted to recreate it here as it is reproducable with the Wasmtime shim and may have to do with how cgroups are being created by Youki.

TLDR: Pod level metrics are not being found by kubelet (even though container level metrics exist).

Repro:

```console
$ make build-wasmtime
$ make test/k3s-wasmtime
# Get running wasmtime shim processes (the test deploys 3 replicas)
$ ps -ax | grep shim-wasmtime
  74103 ?        Sl     0:00 /home/kagold/projects/runwasi/dist/bin/containerd-shim-wasmtime-v1 -namespace k8s.io -id 2e67245de8520087542e916a4d2bb0ab4d8efafdca2c07f69ee5fab37aa62928 -address /run/k3s/containerd/containerd.sock
  74110 ?        Sl     0:00 /home/kagold/projects/runwasi/dist/bin/containerd-shim-wasmtime-v1 -namespace k8s.io -id deee8be6836eace70740e167a402bb7662436df79c165f7d88361791ef4edcde -address /run/k3s/containerd/containerd.sock
  74137 ?        Sl     0:00 /home/kagold/projects/runwasi/dist/bin/containerd-shim-wasmtime-v1 -namespace k8s.io -id 384b86f7322e36eb0aa4d0b312b3a226b268818f29da038f866e69bfb4733eeb -address /run/k3s/containerd/containerd.sock
  74220 ?        Ssl    0:05 /home/kagold/projects/runwasi/dist/bin/containerd-shim-wasmtime-v1 -namespace k8s.io -id 2e67245de8520087542e916a4d2bb0ab4d8efafdca2c07f69ee5fab37aa62928 -address /run/k3s/containerd/containerd.sock
  74221 ?        Ssl    0:05 /home/kagold/projects/runwasi/dist/bin/containerd-shim-wasmtime-v1 -namespace k8s.io -id deee8be6836eace70740e167a402bb7662436df79c165f7d88361791ef4edcde -address /run/k3s/containerd/containerd.sock
[...]

PR #864, opened 2025-02-19
fix: handling of cgroup driver setting

z63d:
fix: https://github.com/containerd/runwasi/issues/821

Fixes an issue where cgroupfs was always used as the cgroup driver.
Receives SystemdCgroup from containerd and allows you to configure the cgroup driver. 
Uses cgroupfs as the default cgroup driver.(Same as runc)


Mossaka:
It looks like one of the test failed with the following error:

```console
Error: failed to stop unit youki-test.scope: dbus error: dbus function call error: Unit youki-test.scope not loaded.

Caused by:
    failed to stop unit youki-test.scope: dbus error: dbus function call error: Unit youki-test.scope not loaded.

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
   2: containerd_shim_wasm::testing::WasiTest<WasiInstance>::wait
   3: containerd_shim_wasmedge::wasmedge_tests::test_seccomp::{{closure}}
   4: core::ops::function::FnOnce::call_once
   5: serial_test::serial_code_lock::local_serial_core_with_return
   6: containerd_shim_wasmedge::wasmedge_tests::test_seccomp
   7: containerd_shim_wasmedge::wasmedge_tests::test_seccomp::{{closure}}
   8: core::ops::function::FnOnce::call_once
   9: test::__rust_begin_short_backtrace
  10: test::run_test::{{closure}}
  11: std::sys::backtrace::__rust_begin_short_backtrace
  12: core::ops::function::FnOnce::call_once{{vtable.shim}}
  13: std::sys::pal::unix::thread::Thread::new::thread_start
test wasmedge_tests::test_seccomp ... FAILED
test wasmedge_tests::test_unreachable ... ok

failures:

failures:
    wasmedge_tests::test_seccomp

```

z63d:
hmm...
This seems to have been a temporary issue. I ran it again and it passed without any issues.

But other problems are occurring...
```
  Warning  FailedCreatePodSandBox  6s (x21 over 4m47s)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: failed to create container: intermediate process error cgroup error: systemd error: dbus error: dbus connection error: ENOENT: No such file or directory. error during cleanup: systemd error: dbus error: dbus connection error: ENOENT: No such file or directory: unknown
```

Makefile
```diff
@@ -363,6 +363,7 @@ test/k3s-%: dist/img.tar bin/k3s dist-%
 	sleep 5s
 	sudo bin/k3s kubectl wait deployment wasi-demo --for condition=Available=True --timeout=5s
 	sudo bin/k3s kubectl get pods -o wide
+	sleep 300s
```
jprendes, line 366: I think we have a 5m timeout in CI, which this would exceed. Does it need to be that long?

test/k8s/Dockerfile
```diff
@@ -17,5 +17,5 @@ ADD dist/bin/* /usr/local/bin/
 ARG RUNTIME
 RUN cat <<EOF >> /etc/containerd/config.toml
 [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.wasm]
-runtime_type = "io.containerd.${RUNTIME}.v1"
+runtime_path = "io.containerd.${RUNTIME}.v1"
```
jprendes, line 20: I didn't fully follow the discussion in Slack about this, but I would expect runtime_path to be a path, while this is not a path.

Makefile
z63d: Something I was testing locally got mixed in 🙏

test/k8s/Dockerfile
z63d: I don't understand why the kind test fails.
This is what I wanted to do. Set runtime_type to path.
Is this difference okay?

https://github.com/containerd/runwasi/blob/472f431542f5a7e37168bc7dd974d011f56ad207/test/k3s/bootstrap.sh#L8
https://github.com/containerd/runwasi/blob/472f431542f5a7e37168bc7dd974d011f56ad207/test/k8s/Dockerfile#L20

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
```diff
@@ -99,6 +100,28 @@ impl<T: Instance + Send + Sync, E: EventSender> Local<T, E> {
...
+            None
+        }
+        let systemd_cgroup = match req.options.as_ref() {
+            Some(any) => match String::from_utf8(any.value.to_vec()) {
+                Ok(opts_str) => parse_systemd_cgroup(opts_str.as_str()).unwrap_or(true),
+                Err(_) => true,
+            },
+            None => true,
+        };
+
```
jprendes, line 124: Would you mind trying this?
It needs 2 new dependencies
```toml
prost = "0.13"
toml = "0.8"
```

```rust
        #[derive(Deserialize, Default, Clone, PartialEq)]
        struct Config {
            #[serde(alias = "SystemdCgroup")]
            systemd_cgroup: bool,
        }
        
        fn get_config(options: Option<&Any>) -> anyhow::Result<Config> {
            let Some(opts) = options else {
                return Ok(Default::default());
            };

            ensure!(
                opts.type_url == "runtimeoptions.v1.Options",
                "Invalid options type {}",
                opts.type_url
            );

            #[derive(Message, Clone, PartialEq)]
            struct Options {
                #[prost(string)]
                type_url: String,
                #[prost(string)]
                config_path: String,
                #[prost(string)]
                config_body: String,
            }

            let opts = Options::decode(opts.value.as_slice())?;

            let config = toml::from_str(opts.config_body.as_str()).map_err(|err| {
                Error::InvalidArgument(format!("invalid shim options: {err}"))
            })?;

            Ok(config)
        }

        let config = get_config(req.options.as_ref()).map_err(|err| {
            Error::InvalidArgument(format!("invalid shim options: {err}"))
        })?;
        let systemd_cgroup = config.systemd_cgroup;
```

z63d: This is great. Works perfectly!
Would it be better to separate the`Options` and `Config` logic into different files?

z63d:
This error occurs randomly.
It doesn't seem to be a specific test.
Although the probability is low, there are cases where it does not occur.
```
Error: failed to stop unit youki-test.scope: dbus error: dbus function call error: Unit youki-test.scope not loaded.

Caused by:
    failed to stop unit youki-test.scope: dbus error: dbus function call error: Unit youki-test.scope not loaded.
```

Probably this happens too.
I followed the Makefile but it doesn't work locally so I haven't tried it yet.
https://github.com/containerd/runwasi/actions/runs/13409133082/job/37455738328
> hmm... This seems to have been a temporary issue. I ran it again and it passed without any issues.
> 
> But other problems are occurring... This did not happen when I was testing with K3s.
> 
> ```
>   Warning  FailedCreatePodSandBox  6s (x21 over 4m47s)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: failed to create container: intermediate process error cgroup error: systemd error: dbus error: dbus connection error: ENOENT: No such file or directory. error during cleanup: systemd error: dbus error: dbus connection error: ENOENT: No such file or directory: unknown
> ```

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
```diff
@@ -34,6 +37,51 @@ use crate::sys::metrics::get_metrics;
...
+#[derive(Deserialize, Clone, PartialEq)]
+struct Config {
+    #[serde(alias = "SystemdCgroup")]
+    systemd_cgroup: bool,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            systemd_cgroup: true,
```
jprendes, line 59: IIUC what runc is doing, I think this should default to `false`.
Also, I think defaulting this to false would fix a few of the CI failures.

jprendes:
I am confused.

From what you observe, it looks like the options have indeed `type_url` = `runtimeoptions.v1.Options`.
https://github.com/containerd/containerd/blob/main/api/types/runtimeoptions/v1/api.proto

But it also looks like the runc shim is not expecting an option like that, but rather a `runc.options.Options`.
https://github.com/containerd/containerd/blob/main/cmd/containerd-shim-runc-v2/runc/container.go#L53-L62
https://github.com/containerd/containerd/blob/main/api/types/runc/options/oci.proto


I don't understand why we receive different type of options, or how to tell to containerd the type of protobuf message that we are expecting (i.e., our own `runwasi.options.Options`)

Also it looks like containerd sends us the `TaskOptions`, and if that's nil, it sends the `RuntimeOptions` instead.
Maybe that explains the difference?
https://github.com/containerd/containerd/blob/main/core/runtime/v2/shim.go#L561-L564

@cpuguy83 are you familiar with any of this?

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
z63d: If you set it to false, most k8s clusters will have metrics bugs unles explicitly specify `SystemdCgroup = true` in containerd/config.toml.
Is this OK?

> Also, I think defaulting this to false would fix a few of the CI failures.

I think the kind problem will be solved.
But when using systemd cgroups, kind pod doesn't seem to work.
As you can see below, kind supports systemd cgroups. It's strange that it doesn't work.
https://github.com/containerd/runwasi/pull/356#issuecomment-1766693160

jprendes: > Is this OK?

IIUC, that would be the same behaviour as runc, so it´s fine with me.

jprendes: Also, I think it's better to err on the side of not using systemd.
If we wrongly DON'T use systemd, the metrincs don't work.
If we wrongly DO use systemd, the container doesn't even start.

cpuguy83:
> I am confused.
> 
> From what you observe, it looks like the options have indeed `type_url` = `runtimeoptions.v1.Options`. https://cloud-native.slack.com/archives/C04LTPB6Z0V/p1739716368576559?thread_ts=1738110849.528919&cid=C04LTPB6Z0V https://github.com/containerd/containerd/blob/main/api/types/runtimeoptions/v1/api.proto
> 
> But it also looks like the runc shim is not expecting an option like that, but rather a `runc.options.Options`. https://github.com/containerd/containerd/blob/main/cmd/containerd-shim-runc-v2/runc/container.go#L53-L62 https://github.com/containerd/containerd/blob/main/api/types/runc/options/oci.proto
> 
> I don't understand why we receive different type of options, or how to tell to containerd the type of protobuf message that we are expecting (i.e., our own `runwasi.options.Options`)
> 
> Also it looks like containerd sends us the `TaskOptions`, and if that's nil, it sends the `RuntimeOptions` instead. Maybe that explains the difference? https://github.com/containerd/containerd/blob/main/core/runtime/v2/shim.go#L561-L564
> 
> @cpuguy83 are you familiar with any of this?

@jprendes This is due to this code: https://github.com/containerd/containerd/blob/6f652853f01ef9ba340a860c2f39edf1701102d1/internal/cri/config/config.go#L658-L667

The runwasi shim is not of type `io.containerd.runc.v2` so it falls back to the raw options.

test/k8s/Dockerfile
Mossaka: Why is runtime_type a path in k3s but not in k8s? I am confused. 

Should `runtime_type` always be in the form of `io.containerd.${RUNTIME}.{version}` for k8s and k3s?

z63d:
@Mossaka @jprendes 
I think CI started failing after the stress test PR was merged.
Do you know why?
https://github.com/containerd/runwasi/actions/runs/13470105500/job/37642539745?pr=864

test/k8s/Dockerfile
z63d: Found it. It's probably mentioned here.
https://github.com/containerd/containerd/blob/v2.0.2/core/runtime/runtime.go#L49-L51
And it is processed by resolveRuntimePath.
https://github.com/containerd/containerd/blob/v2.0.2/core/runtime/v2/shim_manager.go#L308-L384

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
z63d: I don't understand it well enough.
Does this mean that if containerd has `SystemdCgroup = false` and runwasi uses systemd, the container won't start?

Mossaka:
Rebase?

Mossaka (approved):

jprendes:
> Rebase?

I don't think it needs one.

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
jprendes: I'm happy for the cleanup to go in a different PR, and getting this merged.
I would, for example, store the whole Config in the InstanceConfig, not just the systemd bit.
I also think we need at least one test that excercises this code.
But as I said, happy for that to go in other PRs.

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
jprendes: Exactly

jprendes (approved):
LGTM. 
I think we need a bit of cleanup and some tests, but I'm happy for that to be on different PRs.
I would, for example, store the whole Config in the InstanceConfig, not just the systemd bit.
I also think we need at least one test that excercises this code.
@z63d , would you be interested in contributing those changes?

z63d:
@jprendes Sure!

PR #866, opened 2025-02-20
chore: bump rust version to 1.85.0 and edition to 2024

Mossaka:
there are few places that I need to fix:

1. `std::env::set_var` is now unsafe because it's not thread-safe.
2. `WaitableCell::set_guard_with`  function signature needs to use precise capturing.
3. Fixed some cargo fmt and clippy issues.
4. `extern` blocks are `unsafe` now

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-edition = "2021"
+edition = "2024"
 version = "0.4.0"
 license = "Apache-2.0"
 readme = "README.md"
--- a/crates/containerd-shim-wasm/src/sandbox/cli.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/cli.rs
@@ -81,10 +81,10 @@
 
 use std::path::PathBuf;
 
-use containerd_shim::{parse, run, Config};
+use containerd_shim::{Config, parse, run};
 
 #[cfg(feature = "opentelemetry")]
-use crate::sandbox::shim::{otel_traces_enabled, OtlpConfig};
+use crate::sandbox::shim::{OtlpConfig, otel_traces_enabled};
 use crate::sandbox::{Instance, ShimCli};
 
 pub mod r#impl {
@@ -175,8 +175,8 @@ pub fn shim_main<'a, I>(
         use tokio::runtime::Runtime;
         let rt = Runtime::new().unwrap();
         rt.block_on(async {
-            let _guard = OtlpConfig::build_from_env()
-                .expect("Failed to build OtelConfig.")
+            let otlp_config = OtlpConfig::build_from_env().expect("Failed to build OtelConfig.");
+            let _guard = otlp_config
                 .init()
                 .expect("Failed to initialize OpenTelemetry.");
             shim_main_inner::<I>(name, version, revision, shim_version, config);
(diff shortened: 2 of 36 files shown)


crates/containerd-shim-wasm/src/sandbox/sync.rs
```diff
@@ -86,7 +86,7 @@ impl<T> WaitableCell<T> {
     /// The function `f` will always be called, regardless of whether the `WaitableCell` has a value or not.
     /// The `WaitableCell` is going to be set even in the case of an unwind. In this case, ff the function `f`
     /// panics it will cause an abort, so it's recommended to avoid any panics in `f`.
-    pub fn set_guard_with<R: Into<T>>(&self, f: impl FnOnce() -> R) -> impl Drop {
+    pub fn set_guard_with<R: Into<T>, U: FnOnce() -> R>(&self, f: U) -> impl Drop + use<U, T, R> {
```
jprendes, line 89: Oh, interesting! Would just `use<U>` be enough here?
Also, could it be `F` instead of `U`?

crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -212,7 +212,10 @@ fn shim_main_inner<'a, I>(
             OtlpConfig::set_trace_context(&ctx).unwrap();
         } else {
             let ctx = OtlpConfig::get_trace_context().unwrap();
-            std::env::set_var("TRACECONTEXT", ctx);
+            // SAFETY: we are in a single-threaded context
```
jprendes, line 215: This is not really true. When this runs we have already started a miltithreaded tokio runtime.
But I think we can *assume* that all the other threads aren't messing around with env vars at this point.

jprendes (approved):
LGTM.
Just minor comments.

crates/containerd-shim-wasm/src/sandbox/sync.rs
Mossaka: > "currently, all type parameters are required to be mentioned in the precise captures list"

Cargo.toml
```diff
@@ -14,7 +14,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-edition = "2021"
+edition = "2024"
```
jsturtevant, line 17: this is my first time encountering rust editions. Does this mean that our downstream consumers will require this edition too?

Mossaka: no, unless our APIs use features exclusive on edition 2024, then consumers of our crate may have compatibility issues.

crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -212,7 +212,12 @@ fn shim_main_inner<'a, I>(
             OtlpConfig::set_trace_context(&ctx).unwrap();
         } else {
             let ctx = OtlpConfig::get_trace_context().unwrap();
-            std::env::set_var("TRACECONTEXT", ctx);
+            // SAFETY: although it's in a multithreaded context,
+            // it's safe to assume that all the other threads are not
+            // messing with env vars at this point.
+            unsafe {
+                std::env::set_var("TRACECONTEXT", ctx);
```
jsturtevant, line 219: 👀 interesting!

jsturtevant (commented):
LGTM

crates/containerd-shim-wasm/src/sandbox/cli.rs
Mossaka: Indeed!

PR #876, opened 2025-02-25
chore: refactor for handling of cgroup driver

z63d:
Refactor of https://github.com/containerd/runwasi/pull/864 .

### DIFF
--- a/crates/containerd-shim-wasm/src/sandbox/instance.rs
+++ b/crates/containerd-shim-wasm/src/sandbox/instance.rs
@@ -105,8 +105,8 @@ impl InstanceConfig {
     }
 
     /// get the options config for the instance
-    pub fn get_config(&self) -> Config {
-        self.config.clone()
+    pub fn get_config(&self) -> &Config {
+        &self.config
     }
 }
 


crates/containerd-shim-wasm/src/sandbox/shim/mod.rs
```diff
@@ -6,6 +6,7 @@ mod cli;
 mod events;
 mod instance_data;
 mod local;
+pub use local::Config as OptionsConfig;
```
jprendes, line 9: **nit:** Can we rename this to just `Config`?
```suggestion
pub use local::Config;
```

crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -24,8 +25,8 @@ pub struct InstanceConfig {
     namespace: String,
     /// GRPC address back to main containerd
     containerd_address: String,
-    /// Enables systemd cgroup.
-    systemd_cgroup: bool,
+    /// config of containerd runtime options
+    options_config: OptionsConfig,
```
jprendes, line 29: **nit:** Rename to just `config`
```suggestion
    config: Config,
```

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
```diff
@@ -47,8 +48,12 @@ struct Options {
...
-#[derive(Deserialize, Default, Clone, PartialEq)]
-struct Config {
+/// This is generated by decoding the `options` field of a `CreateTaskRequest` to get an `Options` struct,
+/// interpreting the `config_body` field as TOML,
+/// and deserializing it.
+#[derive(Serialize, Deserialize, Default, Clone, PartialEq, Debug)]
+pub struct Config {
+    /// Enables systemd cgroup.
     #[serde(alias = "SystemdCgroup")]
     systemd_cgroup: bool,
```
jprendes, line 58: **nit:** Make this member public and remove the `get_systemd_cgroup` accessor method
```suggestion
    pub systemd_cgroup: bool,
```

crates/containerd-shim-wasm/src/sandbox/instance.rs
```diff
@@ -97,15 +98,15 @@ impl InstanceConfig {
         self.containerd_address.clone()
     }
 
-    /// set the systemd cgroup for the instance
-    pub fn set_systemd_cgroup(&mut self, systemd_cgroup: bool) -> &mut Self {
-        self.systemd_cgroup = systemd_cgroup;
+    /// set the options config for the instance
+    pub fn set_options_config(&mut self, options_config: OptionsConfig) -> &mut Self {
```
jprendes, line 102: Ah, I always disliked that this struct has accessor methods instead or public members. But since that's the pattern here, lets just go with it 🥲

jprendes (commented):
Thanks!
This looks good. Left a few nits.

@z63d, are you planning to add the tests in this PR or in a new one?

crates/containerd-shim-wasm/src/sandbox/instance.rs
z63d: Being unfamiliar with Rust or runwasi, I don't know why you want to remove accessor methods, but I could do this in a separate PR.

z63d:
@jprendes 
I was going to add the test commits to this PR. Of course it can be separated.

I thought it would be better to split the Options struct and Config struct into different files(e.g. runtimeoptions.rs), but what are your thoughts on this?

crates/containerd-shim-wasm/src/sandbox/instance.rs
jprendes: Don't worry with the accessors in `InstanceConfig`, that would need a wider refactor.
Thanks for contributing :-)

jprendes (approved):
Thanks!
LGTM

jprendes:
> Split?

Yeah, that'd be best :-)

PR #877, opened 2025-02-25
Add plugin support to WasmEdge runtime

CaptainVincent:
As title.

I have prepared a verification repo demonstrating how to use the wasi_nn plugin with ggml backend.

https://github.com/CaptainVincent/runwasi-wasmedge-demo

And a CI workflow testing in [here](https://github.com/CaptainVincent/runwasi-wasmedge-demo/actions).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -966,6 +966,7 @@ name = "containerd-shim-wasmedge"
 version = "0.5.0"
 dependencies = [
  "anyhow",
+ "cfg-if 1.0.0",
  "containerd-shim-wasm",
  "libc",
  "log",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ members = [
     "crates/containerd-shim-wasmer",
     "crates/containerd-shim-wamr",
     "crates/stress-test",
-    "benches/containerd-shim-benchmarks", 
+    "benches/containerd-shim-benchmarks",
 ]
 resolver = "2"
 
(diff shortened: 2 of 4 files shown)


jprendes:
Tests are failing with the following message:
```
[error] wasmedge runtime failed: set null value into non-nullable value type, Code: 0x009
[error] Address could not be matched to any shared object. Detailed error information is not available.
hello world
```

@CaptainVincent do you have any clue?

CaptainVincent:
Not yet, but I can replicate this issue locally. I’ll try using gdb to see what’s happening.

CaptainVincent:
https://github.com/bpowers/musl/blob/master/src/ldso/dladdr.c
It seems that musl does not support dynamic library-related functions like dladdr during static linking, so I disabled the plugin for this target. Have any other suggestions?

emirhandurmus:
@CaptainVincent , thank you for the improvement! Personally, I've been waiting for this to use the wasi_nn plugin with the ggml backend. Once I saw your commits I just would like to test it and share my results.
I've build the shim with your commits and replicated your demo cases locally, and they work fine. However, I noticed an issue when running the container without the host network.

I tested it with -p 8080:8080, and the container gets stuck in the "Created" status. 

Then, I also tried running a simple HTTP server from the WasmEdge quick start guide:
docker run -dp 8080:8080 --rm --runtime=io.containerd.wasmedge.v1 --platform=wasi/wasm secondstate/rust-example-server:latest

Unfortunately, the issue persists when running containers without the host network. So before merging, I just wanted to give you a heads-up, so you might want to check this before proceeding.

jprendes (approved):
LGTM.
Just bear in mind that the env-vars you are parsing are those in the container spec, and not those of the shim.
Also the plugin is loaded from the container's rootfs, and not from the host.
I'm happy to merge as is, and discuss if we want to switch to loading from the host in a different issue / PR.

PR #878, opened 2025-02-25
sandbox/cli: add logger to the container process

Mossaka:
this commit initializes the containerd-shim::logger inside of the zygote global process (the container process) so that containers can send logs to containerd. 

This PR is pending this PR to rust-extenions that makes the logger module public: https://github.com/containerd/rust-extensions/pull/378 and a new version released by rust-extensions.

PR #879, opened 2025-02-25
add container ID and pod ID to RuntimeContext and logs

Mossaka:
- **containerd-shim-wasm/src/sandbox/cli: add logger to the container process**
- **feat(all): add container id to runtimeContext so that logs can refer to it**

The purpose of adding `id` to RuntimeContext is that containers can add the id to their log messages to differentiate from other containers in the same pod. 

The disadvantage of this method is that functions that do not import `ctx: RuntimeContext` will not be able to have the container id in their logs.

This builds on top of #878 

---

## Example

The following is containerd log for a pod that has three wasm tasks. 

```console
time="2025-02-25T22:12:39.421124341Z" level=info instance="7e7916133550e01b22c57fbcb85ae0b18282a28dac7a61da18d0ebc52f5cc43c" msg="setting up wasi"
time="2025-02-25T22:12:39.918793845Z" level=info instance="45facf66a6c0fd780272599bf7a1a15604445b9811102c9ebd39dfc24a7690c6" msg="setting up wasi"
time="2025-02-25T22:12:40.460339278Z" level=info instance="60ad7697d284b0ab4a51664ae30da83aefc95284213156ee8d1c5cc1929a3144" msg="setting up wasi"
```

### DIFF
--- a/crates/containerd-shim-wamr/Cargo.toml
+++ b/crates/containerd-shim-wamr/Cargo.toml
@@ -19,3 +19,6 @@ serial_test = { workspace = true }
 [[bin]]
 name = "containerd-shim-wamr-v1"
 path = "src/main.rs"
+
+[package.metadata.cargo-machete]
+ignored = ["log"]
--- a/crates/containerd-shim-wasm/src/container/context.rs
+++ b/crates/containerd-shim-wasm/src/container/context.rs
@@ -39,6 +39,13 @@ pub trait RuntimeContext {
 
     // the container id for the running container
     fn container_id(&self) -> &str;
+
+    // the pod id for the running container (if available)
+    // In Kubernetes environments, containers run within pods, and the pod ID is usually
+    // stored in the OCI spec annotations under "io.kubernetes.cri.sandbox-id"
+    fn pod_id(&self) -> Option<&str> {
+        None
+    }
 }
 
 /// The source for a WASI module / components.
@@ -105,7 +112,7 @@ impl RuntimeContext for WasiContext<'_> {
             .process()
             .as_ref()
             .and_then(|p| p.env().as_ref())
-            .map(|e| e.as_slice())
+            .map(|a| a.as_slice())
             .unwrap_or_default()
     }
 
@@ -141,6 +148,14 @@ impl RuntimeContext for WasiContext<'_> {
 
     fn container_id(&self) -> &str {
         &self.id
+    }
+
+    fn pod_id(&self) -> Option<&str> {
+        self.spec
+            .annotations()
+            .as_ref()
+            .and_then(|a| a.get("io.kubernetes.cri.sandbox-id"))
+            .map(|s| s.as_str())
     }
 }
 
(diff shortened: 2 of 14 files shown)


crates/containerd-shim-wasm/src/container/context.rs
```diff
@@ -36,6 +36,16 @@ pub trait RuntimeContext {
...
+
+    // the container id for the running container
+    fn container_id(&self) -> &str;
+
+    // the pod id for the running container (if available)
+    // In Kubernetes environments, containers run within pods, and the pod ID is usually
+    // stored in the OCI spec annotations under "io.kubernetes.cri.sandbox-id"
+    fn pod_id(&self) -> Option<&str> {
+        None
+    }
```
jsturtevant, line 48: I think this is fine.  We might consider putting this behind a struct/function that hides this behind something like `info()` if we have more things we might want to log.  But as of now I can't think of anything.

crates/containerd-shim-wasmedge/src/instance.rs
```diff
@@ -54,7 +56,7 @@ impl Engine for WasmEdgeEngine {
             .register_module(Some(&mod_name), module)
             .context("registering module")?;
 
-        log::debug!("running with method {func:?}");
+        containerd_shim_wasm::debug!(ctx, "running with method {func:?}");
```
jsturtevant, line 59: is it possible to do something like: 

```suggestion
        containerd_shim_wasm::debug!(ctx, func=func; "running method");
```

so down stream users could add there own structured logs?

jsturtevant (commented):
This is great!  Huge step forward from an operational standpoint with logging.

Maybe add something to the docs?

crates/containerd-shim-wasmedge/src/instance.rs
Mossaka: I don't think this is supported. do you think we need to support it?

jsturtevant: No not necessarily, I though it might just work with the marco as is or small tweak. It is a nice to have but not critical, a downstream consumer could bypass our convenience macro and pass their own structure if they wanted too

Mossaka: okay if there is a demand for it I will add more macros for it.

crates/containerd-shim-wamr/Cargo.toml
```diff
@@ -19,3 +19,6 @@ serial_test = { workspace = true }
 [[bin]]
 name = "containerd-shim-wamr-v1"
 path = "src/main.rs"
+
+[package.metadata.cargo-machete]
+ignored = ["log"]
```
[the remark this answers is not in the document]
Mossaka: yes, we are running `cargo machete` to find unused dependencies. In this case, the log crate is a false positive because we are using it behind the macro.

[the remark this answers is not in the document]
Mossaka: https://github.com/containerd/runwasi/actions/runs/13685357470/job/38267413297?pr=879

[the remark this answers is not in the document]
jsturtevant: oh, I see its in some of the cargo files below.

jsturtevant:
code LGTM, is it possible to add a section describing the fact that we do have structured logging on the operation section of the docs?  Feels like that is something that would be helpful to know.  Maybe it could live under the [troubleshooting ](https://runwasi.dev/resources/troubleshooting.html) section?

crates/containerd-shim-wasm/src/container/log.rs
```diff
@@ -0,0 +1,119 @@
...
+    // Info level log
+    (info, $ctx:expr, $($arg:tt)+) => {
+        {
+            let ctx = $ctx;
+            match ctx.pod_id() {
+                Some(pod_id) => log::info!(instance = ctx.container_id(), pod = pod_id; $($arg)+),
+                None => log::info!(instance = ctx.container_id(); $($arg)+)
+            }
+        }
+    };
```
jprendes, line 34: **question:** would something like this work?
```suggestion
    ($level:ident, $ctx:expr, $($arg:tt)+) => {
        {
            let ctx = $ctx;
            match ctx.pod_id() {
                Some(pod_id) => log::$level!(instance = ctx.container_id(), pod = pod_id; $($arg)+),
                None => log::$level!(instance = ctx.container_id(); $($arg)+)
            }
        }
    };
```

jprendes (approved):
LGTM

crates/containerd-shim-wasm/src/container/log.rs
Mossaka: yup!

jprendes (approved):

jsturtevant (approved):
LGTM. Thanks!

PR #880, opened 2025-02-25
symlink higher level docs ./docs/src etc

devigned:
This PR symlinks in the images and `.md` files from the root into the src directory for the mdbook to ensure the content is rendered during the build. The serve command is alerted to changes in the `.md` files. 

This PR also symlinks the logo image which is currently broken on the `README.md` rendered in the book. I also used a picture tag to make it a little easier to have one 1 image show when rendered in the book. If I rendered the `README.md` with mdbook using the `![]` format, both images will render in the mdbook.


devigned:
@dblnz give this a look when you are available, please.

Mossaka (approved):

dblnz:
LGTM
Nice job, @devigned , thanks for taking care of this.

jprendes (approved):
Tested it locally, works like a charm!
Thanks @devigned

PR #881, opened 2025-02-26
[docs] Add mermaid preprocessor to book

jprendes:
This PR adds a mermaid preprocessor so that the flowchart in `oci-decision-flow.md` renders correctly.


devigned:
I should have mentioned that I had noticed this in #880. Thank you for noticing and following on this PR.

docs/mermaid-init.js
devigned, line 1: Same thing would be useful for the entire site. Right now, the light themes don't actually change the header logo on the root page. I didn't dig into dark vs. light theme much due to time constrains. Maybe something like this could solve that issue.

.github/workflows/docs.yml
```diff
@@ -2,8 +2,10 @@ name: Deploy mdbook documentation
 
 on:
   push:
-    branches:
-      - main
+    branches: [main]
+  pull_request:
+    branches: [main]
```
devigned, line 7: Do you really want this deployed on each PR?

devigned (changes requested):
Looks great. Thank you!

I have one comment that may need to be addressed. It looks like you are deploying when the PR CI build is run. I don't think that should be the case.

docs/mermaid-init.js
jprendes: Ah, sorry, I should have mentioned that this file is autogenerated by `mdbook-mermaid install`.
But yes, it would be interesting to find a way to fix light/dark theme in other parts of the book.

.github/workflows/docs.yml
jprendes: The deploy step is conditional on this running off the main branch, so it would be skipped during PRs.
But I think there's value in *building* the book on each PR.

jprendes: Like here, the actual "Deploy" step is skipped.
https://github.com/containerd/runwasi/actions/runs/13543206506/job/37848751089?pr=881

devigned: Oops! You are right. I missed the conditional and totally agree with your reasoning. Resolved!

PR #882, opened 2025-02-26
[chore] Remove some redundant accessor methods

jprendes:
As in the title.
This PR removes some accessor methods that aren't very rusty, and replaces it with `pub` struct members instead.


Mossaka (commented):
LGTM. Could you please add this change to the changelog since it's a (potential) breaking change?

Mossaka (approved):

PR #887, opened 2025-02-28
[chore] Remove unused engine field

jprendes:
There's an unused engine field in the `Local` struct that somehow was not being flagged as unused.
This PR removes that field, and related now redundant types.


Mossaka (approved):
LGTM, thanks!

PR #890, opened 2025-02-28
Further asyncify the codebase

jprendes:
**TL.DR.: This PR sprinkles _async_ and _await_ throughout the codebase.**

This PR further introduces async to the codebase without the leap of switching to async rust-extensions.

The reason not to switch to async rust-extensions yet is that it doesn't work on windows (I.(i.e., it doesn't build on Windows, so our CI fails).
There are a few options here for the future:
* not build on Windows in CI
* add windows support to async ttrpc-rust and rust-extensions.
* create an adapter that sits in between our codebase and rust-extensions that allows us to build with either sync or async rust-extensions (I have a working prototype of this already)

This PR uses our "ambient" tokio runtime to run our async codebase inside of the sync methos that rust-extensions expects.

This PR doesn't convert our testing framework to async either. This is to reduce the amount of changes.


jprendes:
There's an odd CI failure here
https://github.com/containerd/runwasi/actions/runs/13598352245/job/38019970662#step:10:950
That I haven't been able to reproduce. 
I didn't look too much into it, but I think it's an already existing race condition in our codebase. I suspect it's related to how we wait for the process to exit in different parts of the codebase. One of the things after that happens is change the state from running to stopped, but we also wait and so things in other places.
I will investigate further, but I don't think it needs to be addressed in this PR.

jprendes:
rebased

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
```diff
@@ -418,41 +420,41 @@ impl<T: Instance + Sync + Send, E: EventSender> Task for Local<T, E> {
...
+        };
 
         #[cfg(not(feature = "opentelemetry"))]
-        {
-            Ok(self.task_wait(req)?)
+        let span_exporter = std::future::pending::<()>();
+
+        let res = async {
+            tokio::select! {
+                _ = span_exporter => unreachable!(),
```
Mossaka, line 451: maybe add a comment explaining why `span_exporter` is never expected to end? My understanding is that this async block is solely determined by the completion of `self.task_wait(req)`

Mossaka (approved):
LGTM!

crates/containerd-shim-wasm/Cargo.toml
```diff
@@ -28,14 +28,16 @@ serde_json = { workspace = true }
...
+tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 futures = { version = "0.3.30" }
 wasmparser = { version = "0.226.0" }
 tokio-stream = { version = "0.1" }
 sha256 = { workspace = true }
 serde_bytes = "0.11"
 prost = "0.13"
 toml = "0.8"
+trait-variant = "0.1"
+tokio-async-drop = "0.1"
```
Mossaka, line 40: What is this crate used for?

jprendes: Good catch!
Initially I had added as a dev-dependency for the `sandbox::shim::local::tests`, but then thought we could actually use this for the `sandbox::containerd::client` for when we drop the lease.
I'll update the lease to use tokio_async_drop.

crates/containerd-shim-wasm/src/sandbox/shim/local.rs
```diff
@@ -418,41 +420,47 @@ impl<T: Instance + Sync + Send, E: EventSender> Task for Local<T, E> {
...
 
-            let (tx, rx) = std::sync::mpsc::channel();
-            // Start a thread to export interval span for long wait
-
-            let _ = thread::spawn(move || {
+            // This future never completes as it runs an infinite loop.
+            // It will stop executing when dropped.
+            // We need to keep this future's lifetime tied to this
+            // method's lifetime.
+            // This means we shouldn't tokio::spawn it, but ruther
```
Mossaka, line 440:
```suggestion
            // This means we shouldn't tokio::spawn it, but rather
```

Mossaka (approved):
LGTM. Could you please update the changelog?

Mossaka (approved):
LGTM!

PR #903, opened 2025-03-06
all: Fix broken links and add GitHub Actions workflow for link checking

Mossaka:
This PR does two things:
1. Fixed all the broken links in the docs
2. Add a `linkChecker` job to the CI/CD pipelines. It does so in a two-phase checking. The first phase is to run per PR and does not fail the PR if the links are broken. This is becasue links might not be live prior to deployment. The second phase runs after deployment of `runwasi.dev` and create or update issues if broken links are found.


jprendes:
I think we need to increase the 5min timeout on the build jobs, it's the second time I see is hitting it.

Mossaka:
> I think we need to increase the 5min timeout on the build jobs, it's the second time I see is hitting it.

Sure!

Mossaka:
Just to clarify, do you want me to bump the timeout in this PR? @jprendes

jprendes:
No, doesn't need to be here.

CONTRIBUTING.md
```diff
@@ -9,7 +9,7 @@ which apply to all containerd projects.
 
 ## Setting up your local environment
 
-At a minimum, the Rust toolchain.  When using `rustup` the correct toolchain version is picked up from the [rust-toolchain.toml](./rust-toolchain.toml) so you don't need to worry about the version.
+At a minimum, the Rust toolchain.  When using `rustup` the correct toolchain version is picked up from the "rust-toolchain.toml" file so you don't need to worry about the version.
```
devigned, line 12: Why use the string rather than a link to the repo?

CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ scripts/setup-cross.sh
 
 ## Project structure and architecture
 
-For a detailed overview of the Runwasi structure and architecture, please refer to the [Architecture Documentation](./docs/src/user-guide/architecture.md).
+For a detailed overview of the Runwasi structure and architecture, please refer to the [Architecture Documentation](https://runwasi.dev/developer/architecture.html).
```
devigned, line 42: Does this check happen before or after deploy? If it happens before deploy, then new links added with new files will not pass the link check b/c the file would not be published to runwasi.dev yet.

CONTRIBUTING.md
```diff
@@ -61,7 +61,7 @@ To build a shim for specific runtime (wasmtime, wasmer, wasmedge, wamr, etc):
 make build-<runtime>
 ```
 
-By default the runtimes will build for your current OS and architecture.  If you want to build for a specific OS and architecture you can specify `TARGET`, where it matches a target in [Cross.toml](./Cross.toml). If your target doesn't match your host OS and architecture [Cross](https://github.com/cross-rs/cross) will be used. As an example will build a static binary:
+By default the runtimes will build for your current OS and architecture.  If you want to build for a specific OS and architecture you can specify `TARGET`, where it matches a target in "Cross.toml". If your target doesn't match your host OS and architecture [Cross](https://github.com/cross-rs/cross) will be used. As an example will build a static binary:
```
devigned, line 64: Link to repo rather than string?

CONTRIBUTING.md
```diff
@@ -90,9 +90,9 @@ RUST_LOG=DEBUG cargo test --package containerd-shim-wasmtime --lib -- wasmtime_t
 
 ### End to End tests
 
-The e2e test run on [k3s](https://k3s.io/) and [kind](https://kind.sigs.k8s.io/).  A test image is built using [oci-tar-builder](./crates/oci-tar-builder/) and is loaded onto the clusters.  This test image is not pushed to an external registry so be sure to use the Makefile targets to build the image and load it on the cluster.
+The e2e test run on [k3s](https://k3s.io/) and [kind](https://kind.sigs.k8s.io/).  A test image is built using [oci-tar-builder](https://github.com/containerd/runwasi/tree/main/crates/oci-tar-builder) and is loaded onto the clusters.  This test image is not pushed to an external registry so be sure to use the Makefile targets to build the image and load it on the cluster.
```
devigned, line 93: I think floating links like this would be most desirable in the docs since a PR moving or removing the file should cause a failure of the link check.

README.md
```diff
@@ -54,4 +54,4 @@ Check out these projects that build on top of runwasi:
 
 ## Contributing
 
-To begin contributing, please read our [Contributing Guide](https://runwasi.dev/developer/contributing.html).
+To begin contributing, please read our [Contributing Guide](https://runwasi.dev/CONTRIBUTING.html).
```
devigned, line 57: I think anything based on checking runwasi.dev will fail only after runwasi.dev is deployed.

devigned (changes requested):
I added some feedback about different types of links. Please consider the link check timing.

CONTRIBUTING.md
Mossaka: The linkChecker runs every midnight and it doesn't run for a deployment. So it might be before or after. Is your concern that there might be a period where we checked in broken links and not got caught by the CI in a timely manner? I think as long as we keep an eye on the CI and fix the broken links timely I am fine with that.

devigned: Interesting. I would imagine a nightly checker would be best for checking links that have broken leading to sites that are not controlled by the project. For links that are controlled by the project, they can be verified at PR / deployment time.

Mossaka: Moved the linkChecker to the ci

CONTRIBUTING.md
Mossaka: what are floating links? Do you mean links like `./crates/oci-tar-builder/`? The issue with this is that in `runwasi.dev` it can't read links outside of the `/docs` folder so the linkChecker would fail.

README.md
Mossaka: How about a two phase linking strategy? 

1. Run link checkers for every PR, skip checking links to `runwasi.dev` or do not fail the PRs if there are errors related to links not found on `runwasi.dev`
2. Run another link checker for every deployment of the `runwasi.dev` and make sure that the links are working in the livesite.

Mossaka: To elaborate more on the second point, I've added a action to create a GitHub issue for broken links it finds at post-deployment of runwasi.dev.

devigned (approved):
lgtm!

PR #911, opened 2025-03-06
ci(all): add read-all permission to all workflows

Mossaka:
By explicitly adding the read-all permission to all workflows, we can
reduce the number of false positives in the scorecard check.

E.g. https://github.com/containerd/runwasi/security/code-scanning/21


jprendes (approved):
LGTM

PR #914, opened 2025-03-06
use async rust-extensions

jprendes:


crates/containerd-shim-wasm/src/sandbox/cli.rs
```diff
@@ -199,69 +227,24 @@ pub fn shim_main<'a, I>(
...
+        run::<ShimCli<I>>(&shim_id, Some(config)).await;
     }
+    .block_on();
 
     #[cfg(target_os = "linux")]
     log_mem();
 }
-
-#[cfg_attr(feature = "tracing", tracing::instrument(level = "Info"))]
-fn shim_main_inner<'a, I>(
```
Mossaka, line 237: I tried a similar refactoring before, but I realized that if there is only one `shim_main` function, then it's own trace will not be exported to the trace collector becuase this function initializes the opentelemetry infras. This is the main reason I splitted it to `shim_main` and `shim_main_inner` so that one can initialize a otel runtime and another one can export it's function tracing to the collector. 

I am assuming with your refactoring we will not be able to see `shim_main` trace anymore.

jprendes: oh, good catch, thanks!

PR #927, opened 2025-03-12
breaking: `shim::cli` is now crate public

Mossaka:
in addition, added detailed documentation for key components and example usage in the `container` module.


jprendes (approved):
LGTM

PR #930, opened 2025-03-13
Move `Instance` API to a new crate `shimkit`

jprendes:
This PR moves the `Instance` API to a new crate.

The `Instance` API doesn't provide much in the way of WASI support, but rather a higher level wrapper for `rust-extensions`.
This PR moves that API to a new crate, allowing containerd-shim-wasm to focus on adding WebAssembly linked value to the shim.

This PR removes a lot of previously public API that none of our downstream dependencies were using, thus moving us closer to a runwasi 1.0 release. It also allows us to continue iterating on the new crate without having to commit to a stable API for it.


crates/shimkit/Cargo.toml
```diff
@@ -0,0 +1,109 @@
+[package]
+name = "shimkit"
+description = "Opinionated library for building containerd shims"
+version = "0.3.1"
```
Mossaka, line 4: question: why does shimkit starts with version 0.3.1?

crates/shimkit/Cargo.toml
```diff
@@ -0,0 +1,109 @@
+[package]
+name = "shimkit"
```
Mossaka, line 2:
```suggestion
name = "containerd-shimkit"
```

crates/shimkit/README.md
```diff
@@ -0,0 +1,3 @@
+# shimkit
```
Mossaka, line 1:
```suggestion
# containerd-shimkit
```

crates/containerd-shim-wasm/src/container/log.rs
```diff
@@ -19,56 +19,26 @@
...
         }
     };
-}
-
-/// Convenience macro for info level logs
-#[macro_export]
-macro_rules! info {
-    ($ctx:expr, $($arg:tt)+) => {
-        $crate::log_with_ctx!(info, $ctx, $($arg)+)
+    ($d:tt, $level:ident, $($rest:ident),+) => {
```
Mossaka, line 38: Nice!

crates/shimkit/Cargo.toml
```diff
@@ -0,0 +1,109 @@
+[package]
```
Mossaka, line 1: Would like to have a CHANGELOG.md file in this crate and a would like to discuss how this will change our release process defined in the release.md.

I'd propose we still do monthly release with the two crates releasing at the same time. What do you think?

crates/shimkit/Cargo.toml
jprendes: I had published shimkit as an experiment a long time ago, and was on 0.2. If we use containerd-shimkit we can start from 0.1.
The only caveat is that we need to publish it for CI to pass!

Mossaka: > The only caveat is that we need to publish it for CI to pass!

why is that?

jprendes: Otherwise the step that check if containerd-shim-wasm can be published fails, because it uses an unpublished dependency

Mossaka (approved):
LGTM, I suggest to get another lgtm from @containerd/runwasi-reviewers before merging it.

crates/containerd-shim-wasm/README.md
```diff
@@ -6,20 +6,12 @@ A library to help build containerd shims for Wasm workloads.
...
-There are two ways to implement a shim:
-1. Using the `Instance` trait
-2. Using the `Engine` trait
-
-What trait to use depends on how much control you need over the container lifecycle and the level of sandboxing you want to provide. The main difference is that the `Engine` trait uses [Youki](https://github.com/youki-dev/youki)'s `libcontainer` crate to manage the container lifecycle, such as creating the container, starting it, and deleting it, and youki handles container sandbox for you. The `Instance` trait gives you more control over the container lifecycle.
-
-### Using the Engine trait
-
-Implement the `Engine` trait for a simpler integration:
+To implement a shim simple implement the `Engine` trait:
```
andreiltd, line 9:
```suggestion
To implement a shim, simply implement the `Engine` trait:
```

crates/containerd-shimkit/src/sandbox/instance_utils.rs
```diff
@@ -0,0 +1,112 @@
...
+use super::{Error, InstanceConfig};
+use crate::sys::DEFAULT_CONTAINER_ROOT_DIR;
+use crate::sys::stdio::open;
+
+#[derive(Serialize, Deserialize)]
+struct Options {
+    root: Option<PathBuf>,
+}
+
+impl InstanceConfig {
```
andreiltd, line 18: Why doesn't this appear where `InstanceConfig` is defined?

andreiltd (approved):

crates/containerd-shimkit/src/sandbox/instance_utils.rs
jprendes: Originally these were free functions defined in this file, and not members of `InstanceConfig`.
I agree that it would be better where `InstanceConfig` is defined, but I would prefer doing that in a different PR.
I think there's a lot of cleanup we can do in this crate, including shuffling things around, and renaming stuff (like the `Cli` struct, the `Local` struct, etc).

cpuguy83 (approved):

PR #931, opened 2025-03-21
make the engine trait async

jprendes:
This PR makes most of the methods in the `Engine` trait `async`.
The `name` and `supported_layers_types` methods are left as non-async methods, as they are intended more as configuration than for doing any actual computation.

### DIFF
--- a/crates/containerd-shim-wasm/src/test/signals.rs
+++ b/crates/containerd-shim-wasm/src/test/signals.rs
@@ -28,6 +28,7 @@ use std::time::Duration;
 
 use anyhow::{Result, bail};
 use containerd_shim_wasm_test_modules::HELLO_WORLD;
+use tokio::time::sleep as async_sleep;
 
 use crate::container::{Engine, RuntimeContext};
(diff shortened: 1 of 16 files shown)


crates/containerd-shim-wasm/src/test/signals.rs
```diff
@@ -40,31 +40,27 @@ impl Engine for SomeEngine {
...
-                    pending().await
-                };
-                tokio::select! {
-                    _ = signal => {}
-                    _ = task => {}
-                };
-                Ok(0)
-            })
+
+        use tokio::time::sleep;
```
Mossaka, line 46: nit: move `use` to the top of the file

Mossaka (dismissed):
Great work. I am curious to see the performance impact of this change. 

nit: remember to update the CHANGELOG for our next release.

jprendes:
> I am curious to see the performance impact of this change.

we would see a test failure if perf degrades, right?

Mossaka (approved):
lgtm

PR #943, opened 2025-03-25
Simplify the signature of `shim_main`

jprendes:
This PR simplifies the signature of `shim_main`.
Previously `shim_main` required:
* the shim's name, which we can obtain from the `Engine` trait to avoid duplication.
* the shim's version, used to compute a `shim_id`, this `shim_id` is only used to print an error message to stderr if starting the shim fails, so we can directly use the shim's name for this purpose.


Mossaka (approved):
LGTM

Mossaka (commented):
Oh this needs a changelog update before we release

PR #944, opened 2025-03-25
Move `shim_main` from free function to method in a trait

jprendes:
This PR removes the `shim_main` free function and replaces it with a `ShimCli::run` trait method.
This trait is implemented for all Shim traits.
It also adds a new `fn version() -> Version` method to engine, and the corresponding `Version` struct.


Mossaka (approved):
LGTM! Need a rebase?

crates/containerd-shim-wamr/src/instance.rs
```diff
@@ -28,6 +29,10 @@ impl Shim for WamrShim {
     fn name() -> &'static str {
         "wamr"
     }
+
+    fn version() -> Version {
+        version!()
```
Mossaka, line 34: a macro calling into a macro! I love this!

Mossaka (approved):
LGTM!

PR #957, opened 2025-03-27
Move modules and items around

jprendes:
Organize items in modules.
the main modules are:
* `contianerd_shim_wasm::shim::{Config, Version, version, Cli, Compiler, Shim}`
  Things related to the shim process behaviour
* `contianerd_shim_wasm::sandbox::Sandbox`
  Things related to the sandbox (container) process
* `contianerd_shim_wasm::sandbox::context::{Entrypoint, WasmLayer, Source, WasmBinaryType, RuntimeContext}`
  Things related to the context passed to the sandbox process


crates/containerd-shim-wasm/src/shim/mod.rs
```diff
@@ -68,3 +70,26 @@ mod tests;
...
+
+pub use cli::Cli;
+pub use containerd_shimkit::shim_version as version;
+
+/// Config of shim binary options provided by shim implementations
+#[derive(Debug)]
+pub struct Config {
+    /// Disables automatic configuration to use the shim FIFO
+    pub no_setup_logger: bool,
+    // Sets the the default log level. Default is info
```
Mossaka, line 84:
```suggestion
    /// Sets the the default log level. Default is info
```

crates/containerd-shim-wasm/src/sandbox/path.rs
```diff
@@ -1,7 +1,7 @@
 use std::path::{Path, PathBuf};
 
 /// PathResolve allows to resolve a file path in a set of directories.
-pub trait PathResolve {
+pub(crate) trait PathResolve {
```
Mossaka, line 4: this is a breaking change I think? Can you document it in the CHANGELOG?

crates/containerd-shim-wasm/src/sandbox/path.rs
```diff
@@ -19,7 +19,7 @@ pub trait PathResolve {
 }
 
 /// Gets the content of the `PATH` environment variable as an iterator over its components
-pub fn paths() -> impl Iterator<Item = PathBuf> {
+pub(crate) fn paths() -> impl Iterator<Item = PathBuf> {
```
Mossaka, line 22: this is a breaking change I think? Can you document it in the CHANGELOG?

crates/containerd-shim-wasm/src/sandbox/path.rs
jprendes: It is not. This trait was not accessible because the module was not public and it was not reexported anywhere.

crates/containerd-shim-wasm/src/sandbox/path.rs
jprendes: Same as above.

Mossaka (approved):
lgtm!

PR #959, opened 2025-03-28
.github/workflows/release: Update crates.io ownership check to use GH actor instead of runwasi committers

Mossaka:
the issue with adding runwasi committers as an owner is that the release bot does not have the visibility to see it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Add crates.io ownership
         if: ${{ steps.runtime_sub.outputs.is_shim != 'true' && github.repository == 'containerd/runwasi' && !inputs.initial_release }}
         run: |
-          cargo owner --list ${{ inputs.crate }} | grep ${{ github.actor }} || \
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -86,8 +86,8 @@ jobs:
       - name: Add crates.io ownership
         if: ${{ steps.runtime_sub.outputs.is_shim != 'true' && github.repository == 'containerd/runwasi' && !inputs.initial_release }}
         run: |
-          cargo owner --list ${{ inputs.crate }} | grep github:containerd:runwasi-committers || \
-          cargo owner --add github:containerd:runwasi-committers ${{ inputs.crate }}
+          cargo owner --list ${{ inputs.crate }} | grep ${{ github.actor }} || \
```
jprendes, line 89: I think that if runwasi-committers is an owner we don't need to add the GH actor
```suggestion
          cargo owner --list ${{ inputs.crate }} | grep -E 'github:containerd:runwasi-committers|${{ github.actor }}' || \
```

jprendes (approved):
Thanks @Mossaka !

cpuguy83 (approved):

PR #966, opened 2025-04-06
fix: CHIPSET in setup containerd script

z63d:
There is a discrepancy between the comment and the default CHIPSET.

It would be nice to be able to choose interactively when starting the script.
I'll make changes if necessary.


jprendes:
Thanks for this PR, good job spotting this error!
Did you run into any issue that required this fix?
As far as I can tell `setup-containerd.sh` is not used anywhere in the repo, so I would prefer deleting it rather than fixing it.
But I might be missing something :-)
Are you using `setup-containerd.sh` as part of any workflow?

z63d:
I used it for the setup when I was trying out other Issue.
It uses `ctr`.
Easy to set up.

I don't think it's used in CI/CD workflows.

We might not need this script if we had the containerd documentation.
https://github.com/containerd/containerd/blob/main/docs/getting-started.md#installing-containerd

Is it not necessary to specify which version of containerd to use?

jprendes:
> We might not need this script if we had the containerd documentation.

We do have a link to that page in [our doc](https://runwasi.dev/getting-started/installation.html#prerequisites) (from [docs/src/getting-started/installation.md](https://github.com/containerd/runwasi/blob/main/docs/src/getting-started/installation.md?plain=1#L10))

z63d:
Ah, so we don't need this script!

jprendes (approved):
Thanks @z63d !

PR #971, opened 2025-04-11
Remove unused platform, pod_id, and container_id from WasiContext

jprendes:


jsturtevant:
I don't remember the exact reasoning.  At the time, it might have been to pass additional context to the shims so the could make log/make decisions but looks like it didn't end up being used. I don't see it being used in the Spin shim.  

While we are cleaning this up, we could probably remove it from the `containerd_client` api. 
 We use it to check if we can find the correct layers but then just pass it through to the context. I am not sure its actually needed outside the client. 

bhttps://github.com/containerd/runwasi/blob/d24e9e66bcec13b41fe57f97c7531bd3be1e0f7c/crates/containerd-shim-wasm/src/containerd/client.rs#L391-L398

Mossaka (approved):

Mossaka (dismissed):
lgtm!

need to update the changelog

Mossaka:
Runner ubuntu-20.04 are retired.

Mossaka (approved):

Issue #995, opened 2025-05-20
Update the CI runners from Ubuntu 20.04 to the latest one


Mossaka:
We kept Ubuntu 20.04 LTS runners because we wanted to test cgroups v1 as it is the default on Ubuntu 20.04. However, Ubuntu 20.04 is EOL and was retired from GitHub Actions. We will need to either remove the cgroups v1 test or move to a supported Ubuntu version.

PhantomInTheWire:
Hi @Mossaka, I'm a first-time contributor would really appreciate it if you could take a look at [#996](https://github.com/containerd/runwasi/pull/996) when you have time. Let me know if anything needs to be changed. Thanks!

Mossaka:
Thanks, @PhantomInTheWire !

PR #996, opened 2025-05-21
Remove cgroup v1 and Ubuntu 20.04 tests from CI workflow

PhantomInTheWire:
**Solves:** [https://github.com/containerd/runwasi/issues/995](https://github.com/containerd/runwasi/issues/995)

Ubuntu 20.04 LTS has reached EOL and has been removed from GitHub-hosted runners. runwasi ci initially used it to test cgroups v1, but after reviewing all currently supported GitHub-hosted Linux runners, none appear to support cgroups v1 by default.

Given this, I've removed the Ubuntu 20.04 runner and the corresponding cgroups v1 test.
Let me know if you'd like any adjustments. Happy to help!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,8 +106,7 @@ jobs:
     needs: [build-ubuntu, test-image]
     strategy:
       matrix:
-        # 20.04 uses cgroupv1, 22.04 uses cgroupv2
-        os: ["ubuntu-20.04", "ubuntu-22.04"]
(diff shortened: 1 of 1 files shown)


Mossaka (approved):
Thanks, this LGTM! @jprendes do you have any concerns over removing cgroup v1 from the CI? I'd prefer to merge this PR in as is, and if we do want to support cgroup v1 tests, we can run LIMA to support it (see https://github.com/containerd/runwasi/issues/995#issuecomment-2897592206)

Issue #915, opened 2025-03-06
Migrate the wasmtime shim README to /docs


Mossaka:
the README from wasmtime shim contains a WASI HTTP demo application which I feel it's useful to put into our docs.

Owais-04:
Is this still open? I'm new to open source contribution and would like to start with documentation. Thank you!

Mossaka:
Hey @Owais-04 , yes this is still open! Thank you

PR #1040, opened 2025-08-27
Migrate the wasmtime shim README to /docs

Owais-04:
Added the WASI HTTP demo application from wasmtime shim to docs as mentioned in #915


Mossaka (approved):
lgtm, thanks!

Issue #847, opened 2025-02-10
wasmtime: Implement the Config file from v30 release


Mossaka:
The upcoming wasmtime v30 release supports a TOML configuration file https://github.com/bytecodealliance/wasmtime/pull/9811.

It would be great to incorperate this feature into the wasmtime shim and allow it to be configurable by an external configuration file.

kate-goldenring:
Would to toml be added as part of the OCI artifact or as a Pod volume?

PR #1105, opened 2026-04-14
wasmtime: read config from env variable TOML

### LINKED ISSUE #1004: Allow configuration of wasmtime features
Currently, `containerd-shim-wasmtime-v1` doesn't seem to be able to configure the underlying wasmtime runtime.

https://github.com/containerd/runwasi/blob/7a9c892274a9b4353f7bbd5a0e4132ab51f0624d/crates/containerd-shim-wasmtime/src/instance.rs#L67-L82 

This limitation prevent us from running our Wasm OCI artifact, built from component binary that requires `-W function-references,gc` unable to run on the shim.

https://github.com/users/tanishiking/packages/container/package/scala-wasm-echo-server

https://github.com/scala-wasm/scala-wasm/tree/scala-wasm/examples/echo-server

Is there a way to configure wasmtime from `runwasi`?

tanishiking:
fix https://github.com/containerd/runwasi/issues/1004 and
https://github.com/containerd/runwasi/issues/847 ?

Allow workloads to configure Wasmtime per deployment by setting `RUNWASI_WASMTIME_CONFIG_TOML` environment varialbe toml string. Not sure this is the best way to passing configuration, but it'd be a good starting point for discussing the design :)

Example:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: wasmtime-config
data:
  wasmtime.toml: |
    [wasm]
    gc = true
    function-references = true
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: hello
spec:
  template:
    spec:
      runtimeClassName: wasmtime
      containers:
        - name: app
          image: ghcr.io/example/hello-wasm:latest
          command: ["/"]
          env:
            - name: RUNWASI_WASMTIME_CONFIG_TOML
              valueFrom:
                configMapKeyRef:
                  name: wasmtime-config
                  key: wasmtime.toml
```

### DIFF
--- a/crates/containerd-shim-wasm/src/containerd/client.rs
+++ b/crates/containerd-shim-wasm/src/containerd/client.rs
@@ -394,6 +394,7 @@ impl Client {
         engine_name: impl AsRef<str> + Debug,
         supported_layer_types: &[&str],
         compiler: Option<&impl Compiler>,
+        envs: &[String],
     ) -> Result<Vec<WasmLayer>> {
         let container = self.get_container(containerd_id).await?;
         let (manifest, image_digest) = self.get_image_manifest_and_digest(&container.image).await?;
@@ -435,7 +436,7 @@ impl Client {
             return Ok(layers);
         };
 
-        let precompile_id = precompile_label(engine_name.as_ref(), compiler.cache_key());
+        let precompile_id = precompile_label(engine_name.as_ref(), compiler.cache_key(envs));
 
         let image_info = self.get_info(&image_digest).await?;
         let mut needs_precompile = !image_info.labels.contains_key(&precompile_id);
@@ -459,7 +460,7 @@ impl Client {
 
         if needs_precompile {
             log::info!("precompiling layers for image: {}", container.image);
-            let compiled_layers = match compiler.compile(&layers).await {
+            let compiled_layers = match compiler.compile(&layers, envs).await {
                 Ok(compiled_layers) => {
                     if compiled_layers.len() != layers.len() {
                         return Err(ShimError::FailedPrecondition(
@@ -695,6 +696,7 @@ mod tests {
                 "fake",
                 &[WASM_LAYER_MEDIA_TYPE],
                 NO_COMPILER.as_ref(),
+                &[],
             )
             .await
             .unwrap();
@@ -723,6 +725,7 @@ mod tests {
                 "fake",
                 &[WASM_LAYER_MEDIA_TYPE],
                 Some(&engine),
+                &[],
             )
             .await
             .unwrap();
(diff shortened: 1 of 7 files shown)


crates/containerd-shim-wasm/src/containerd/client.rs
```diff
@@ -459,7 +460,7 @@ impl Client {
 
         if needs_precompile {
             log::info!("precompiling layers for image: {}", container.image);
-            let compiled_layers = match compiler.compile(&layers).await {
+            let compiled_layers = match compiler.compile(&layers, envs).await {
```
kate-goldenring, line 463: I am not sure we want to pass all container environment variables to compile.

crates/containerd-shim-wasm/src/shim/shim.rs
```diff
@@ -55,15 +55,19 @@ pub trait Compiler: Sync {
...
     /// This is used to precompile the layers before they are run.
     /// It is called only the first time a module is run and the resulting bytes will be cached in the containerd content store.
     /// The cached, precompiled layers will be reloaded on subsequent runs.
     /// The runtime is expected to return the same number of layers passed in, if the layer cannot be precompiled it should return `None` for that layer.
     /// In some edge cases it is possible that the layers may already be precompiled and None should be returned in this case.
-    async fn compile(&self, _layers: &[WasmLayer]) -> Result<Vec<Option<Vec<u8>>>>;
+    async fn compile(
+        &self,
+        _layers: &[WasmLayer],
+        _envs: &[String],
```
kate-goldenring, line 69: What was your decision process around adding the configurability to `Compiler::compile`. Instead of `Shim::compiler` where many engines (wasmtime and spin) configure the runtime? We could look specify a known configuration file environment variable that is passed to `shim::compiler` to inform the construction of the compiler

kate-goldenring (commented):
Some initial thoughts. Thank you for looking into this. A current workaround should be to set this as a containerd environment variable: https://github.com/spinframework/skips/tree/main/proposals/003-shim-runtime-options#spin-shim-executor-environment-variables

crates/containerd-shim-wasm/src/shim/shim.rs
tanishiking: I kinda misunderstood the role of `Shim::compiler`...
Passing the configuration through `Shim::compiler` makes sense :+1: with `Shim::compiler` configuration change will be reflected  in `cache_key` as well.

tanishiking:
@kate-goldenring Thanks a lot for taking a look! and sorry for the delayed reply.


---

I sent a PR here, but my actual need has become running Wasm binary with GC/EH on Spin. So I think I’ll move this discussion over to `containerd-shim-spin` and open a new issue or PR there instead :)

PR #1106, opened 2026-04-14
deps: bump wasmtime to 43.0.1

tanishiking:


jprendes (dismissed):
LGTM

crates/containerd-shim-wasmer/Cargo.toml
```diff
@@ -9,8 +9,8 @@ containerd-shim-wasm = { workspace = true, features = ["opentelemetry"] }
 log = { workspace = true }
 tokio = { workspace = true }
 
-wasmer = "6.0.1"
-wasmer-wasix = "0.600.1"
+wasmer = "6.1.0"
+wasmer-wasix = "0.601.0"
```
tanishiking, line 13: wasmtime-wasi 43.0.1 requires `url >= 2.5.7`, and `wasmer-package` 0.600.1 pins url `=2.5.2`.
Updated them so it relax to `url = ^2.5`.

crates/containerd-shim-wasmtime/src/instance.rs
```diff
@@ -76,6 +77,7 @@ impl Default for WasmtimeSandbox {
 
         Self {
             engine: wasmtime::Engine::new(&config)
+                .map_err(anyhow::Error::from)
```
tanishiking, line 79: In wasmtime 43 `wasmtime::Error` becomes its own type separate from `anyhow::Error`.
We might want to adapt to the `wasmtime::Error` from `anyhow::Error` in future.

jprendes:
@tanishiking , could you rebase on the latest main? it should include the musl fix.

jprendes:
@tanishiking , it looks like there are some CI failures due to some methods being renamed in wasmtime. Do you mind looking into it?

tanishiking:
Ah yeah, https://github.com/bytecodealliance/wasmtime/pull/12371 makes `Config::async_support` no op. And async APIs can be used without that. (not sure about lint and dependency-review)

jprendes:
you should be able to fix the lints with `make fix` (you might need to run it more than once, and commit changes in between)

tanishiking:
dependency-review failed because of https://github.com/advisories/GHSA-cq8v-f236-94qc
but it seems not very relevant to runwasi, because in the custom loggers under `crates/containerd-shimkit` there's no usage on `rand` I guess. We might wanna `allow-ghsas: GHSA-cq8v-f236-94qc` in GHA ?

Regarding wasmer update  https://github.com/containerd/runwasi/pull/1106/commits/60c8123d2ec78a6348c29544236b962b12617dbb TBH I vibe coded and not fully understand how this change fix the problem 😓 (tried using `WasiRunner` but it didn't work)

therayquaza:
hey 👋  @tanishiking @jprendes any updates on this ?

tanishiking:
TBH, my current focus has moved to running WasmGC binaries using `containerd-shim-spin` (which has a relatively newer version of `wasmtime`), and now I'm not very interested in updating wasmtime of `containerd-shim-wasmtime`...

If you're interested, please feel free to take over :) I believe the remaining issue is that wasmer's transitive dependencies are too outdated, which is blocking the wasmtime update (can we isolate each crates dependencies?)

PR #1107, opened 2026-04-14
Update registry for apk-anywhere

jprendes:
The jorgeprendes420 user from dockerhub seems to be disabled, switching to GHCR instead.


tanishiking:
Ah, nice, I had no idea where the image went, and was giving up musl build locally 😄

devigned (approved):
lgtm!

PR #1115, opened 2026-05-04
Cross-build target for macos compilation

TheRayquaza:
Hey runwasi folks 👋 

I have this tiny PR for enabling compilation on macos using cross

## Usage

```bash
# Build all runtimes for x86_64-unknown-linux-musl
make cross-build

# Build a single shim with release mode
OPT_PROFILE="release" make cross-build-wasmtime

# Override target arch
make cross-build-wasmtime CROSS_BUILD_TARGET=aarch64-unknown-linux-musl
```

## Test plan

- [x] `make cross-build` builds all runtimes in `RUNTIMES`
- [x] `make build` unaffected (no regression)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ override CARGO = cross
 endif
 
 ifeq ($(CARGO),cross)
+export DOCKER_DEFAULT_PLATFORM ?= linux/amd64
 override TARGET_DIR := $(or $(TARGET_DIR),./target/build/$(TARGET)/)
 # When using `cross` we need to run the tests outside the `cross` container.
 # We stop `cargo test` from running the tests with the `--no-run` flag.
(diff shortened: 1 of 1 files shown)


jprendes:
Hi!

I dont see how this change enables aarch64 support. Also aarch64 should already be supported.

Is there a missing commit in the PR maybe?

therayquaza:
Hey @jprendes, you're right that this isn't exactly about adding aarch64 target support. Let me rephrase:

**Problem**: make build fails entirely on macOS because procfs v0.17.0 only supports Linux/Android. There's no workaround, it uses linux syscalls.

**Solution**: cross wraps cargo in docker, so the actual compilation happens on linux. The `DOCKER_DEFAULT_PLATFORM` solves the macos default pulling arm64 images. We need amd64 image because the cross-rs toolchain images are amd64 (without the override, the arm64 image is selected and the build fails).

The new cross-build / cross-build-* targets are just a replacement for running cross instead of cargo, which work for macos (should also work for other platforms I think 🤔)

PR #1117, opened 2026-05-06
ci: pin github actions and fix missing action

TheRayquaza:
Hey runwasi folks :wave:

This is a proposition for
- pinning all your github actions to avoid any supply chain attacks.
- replacing the `cli/setup-gh` with the `wusatosi/setup-gh` action ([details](https://github.com/marketplace/actions/setup-gh))

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,8 +14,8 @@ jobs:
     outputs:
       dirs: ${{ steps.filter.outputs.changes }}
     steps:
-      - uses: actions/checkout@v5
-      - uses: dorny/paths-filter@v4
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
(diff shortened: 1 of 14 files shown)


.github/workflows/docs.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       
       - name: Setup GitHub CLI
         if: ${{ failure() }}
-        uses: cli/setup-gh@v1
+        uses: wusatosi/setup-gh@7dae6c8246092454793fe35dd4d5a9e8bf8ed34d # v1
```
jsturtevant, line 67: I guess we never hit this action so never ran into it.  I don't think we need this at all since the clinshould already be installed.

jsturtevant:
Hey thanks for doing this.  Through some clean up i pinned a few but there are some still left.  If you want to refresh this great other wise i can do it.  

Thanks again for going through this!

PR #1128, opened 2026-05-18
Add OCI spec aware compiler hook (for precompile)

tanishiking:
spinframework/containerd-shim-spin#444 adds Wasmtime TOML feature configuration. Runtime configuration can be provided through a Pod mounted `/runtime/wasmtime.toml` for example, but precompilation currently creates the compiler before the shim has access to the configuration.

This commits adds `Shim::compiler_with_spec(&Spec)`. Instance creation now loads `config.json` before loading OCI Wasm layers and passes the spec to the compiler through that new API.
Downstream shims precompilation should be able to read the spec to configure the wasmtime engine.

Concerns, this might add a small per-container-start cost: 🤔 
- we read `config.json` before precopmile
- construct a precompiler per container rather than once per shim process.

### DIFF
--- a/crates/containerd-shim-wasm/src/shim/shim.rs
+++ b/crates/containerd-shim-wasm/src/shim/shim.rs
@@ -3,6 +3,7 @@ use std::hash::Hash;
 use anyhow::Result;
 #[doc(inline)]
 pub use containerd_shimkit::sandbox::cli::Version;
+use oci_spec::runtime::Spec;
 
 use crate::sandbox::Sandbox;
 use crate::sandbox::context::WasmLayer;
@@ -29,6 +30,11 @@ pub trait Shim: Sync + 'static {
     /// to precompile layers.
     async fn compiler() -> Option<impl Compiler> {
         async move { NO_COMPILER }
+    }
+
+    /// Returns the compiler to be used by this engine with access to the OCI spec.
+    async fn compiler_with_spec(_spec: &Spec) -> Option<impl Compiler> {
+        Self::compiler()
     }
 
     /// Return the supported OCI layer types
--- a/crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
+++ b/crates/containerd-shim-wasm/src/sys/unix/container/instance.rs
@@ -29,31 +29,33 @@ pub struct Instance<S: Shim> {
 
 #[async_trait]
 trait OciClient {
-    async fn load_modules(&self, id: &str) -> Result<Vec<WasmLayer>, SandboxError>;
+    async fn load_modules<P: Compiler>(
+        &self,
+        id: &str,
+        precompiler: Option<&P>,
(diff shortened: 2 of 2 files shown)


jsturtevant:
If I understand correctly from https://github.com/spinframework/containerd-shim-spin/pull/444 you using the spec to find the spin toml which contains the wasmtime configuration options?  

I think you have that information when we call  `async fn compiler` on the trait?  Could you read that file then and hand back the properly configured compiler?

tanishiking:
> If I understand correctly from https://github.com/spinframework/containerd-shim-spin/pull/444 you using the spec to find the spin toml which contains the wasmtime configuration options?

Yes, that is what I’m trying to do. I’d like to read the OCI spec to find the path of the Wasmtime configuration file on the node.
Aat precompile time, the container mounts are not available via the container-visible path (`/runtime/wasmtime.toml` in https://github.com/spinframework/containerd-shim-spin/pull/444), so the shim needs to use the OCI spec to resolve where the configuration file is located on the node.


> I think you have that information when we call async fn compiler on the trait? Could you read that file then and hand back the properly configured compiler?

Yeah... but IIUC it's not possible with the current API (?). To make it possible, I think we need a new API that passes OCI spec information to the shim.

- `runwasi` have that information (`cfg.bundle`) at the call site.
- However, the current `Shim::compiler()` API does not pass any spec  information  to the shim.
- Therefore, `containerd-shim-spin` cannot create a "properly configured compiler": it does not know where the Wasmtime configuration file is on the node.

That's why I’d like to add a new API that passes the OCI spec information to the shim, so that shim can create a properly configured compiler.

---

Please let me know if I’m misunderstanding something.
Is there a way for the shim (containerd-shim-spin) to access the OCI spec during precompilation, or to locate the file created from the `ConfigMap` on the node, without making any changes to `runwasi` ?

jsturtevant:
Thanks, its been awhile but what you've described sounds right.  

>Is there a way for the shim (containerd-shim-spin) to access the OCI spec during precompilation, or to locate the file created from the ConfigMap on the node, without making any changes to runwasi ?

I think using the higher level trait in `containerd-shim-spin` means you don't have access to this.  In a prior version the shims had a bit more information.  

There might be other way's we might be able to make this more generic but would need some concentrated effort.  So I think I am ok with this as an option now.

jprendes (approved):
LGTM

tanishiking:
Thanks for getting this merged! :tada:
I was wondering is there a plan to release a next vesrion of `containerd-shim-wasm` (1.0.2?) with this? This change would unblock WasmGC (or any wasm features) binaries on downstream shims https://github.com/spinframework/containerd-shim-spin/pull/444

tanishiking:
Hi @jprendes, thank you so much for reviewing and merging this PR :)
Would you be able to cut a new release of `containerd-shim-wasm` sometime soon? We'd love to pick up this change downstream.

PR #1138, opened 2026-05-20
Bump rest of Deps to be upto date with the latest

jsturtevant:


jsturtevant:
Notes from checking the protobuf upgrade path:

- The latest `protobuf` crate is currently `4.35.0-release`, but this repo's containerd-facing generated types are still on the rust-protobuf 3 API surface.
- `containerd-shim-protos` releases through `0.11.0` still depend on protobuf 3.x. `0.11.0`, `0.10.0`, and `0.9.0` require protobuf `3.7.x`; `0.8.0` is the newest line that allows the protobuf `3.2.x` range.
- Trying protobuf 4 caused compile errors because APIs/types used by the generated/containerd messages are no longer available in the same form, e.g. `protobuf::well_known_types`, `MessageDyn`, and `SpecialFields`.
- Official `libcontainer 0.6.0` pins `protobuf = "=3.2.0"`, so it cannot resolve with `containerd-shim-protos >=0.9`, which requires protobuf `3.7.x`.

So for the official `libcontainer 0.6.0` release, the compatible stack is `containerd-shim`/`containerd-shim-protos` `0.8.x` with protobuf `3.2.x`. Moving to latest protobuf would need containerd's generated shim protos to be updated/regenerated for protobuf 4 and released first.

jprendes (approved):

